PR #1, opened 2020-01-19
Initial sketch of solution layout

tomcrane:
This application works, and I can run it locally as long as I have AWS creds in a sensible location, and supply a connection string (which might be tunnelled over SSH) in the appSettings.

It serves thumbnails assuming the new layout in S3 introduced with https://github.com/digirati-co-uk/deliverator/pull/8, and also includes a utility to _migrate_ the old layout to the new one (although it does not delete the old layout, yet).

The solution contains bits and pieces of what could grow into a solution containing all the current .NET functionality in Deliverator. This is very much a WIP and we need to decide whether this approach makes sense. I've tried to keep it as light as possible, and as conventional as possible. The only third party addition is really _Dapper_.


tomcrane:
Next step - containerise, deploy to a production environment servicing a test path (e.g., /newthumbs/) and observe.
Write an RFC for this approach to thumbs.

Thumbs/InfoJsonBuilder.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                if (width > 0) sizeStr.Append(", ");
+                width = wh[0];
+                height = wh[1];
+                sizeStr.Append("{ \"width\": ");
+                sizeStr.Append(width);
+                sizeStr.Append(", \"height\": ");
+                sizeStr.Append(height);
+                sizeStr.Append(" }");
+            }
```
stephenwf: It would be nice to have `sizes`  serializable like in `ThumbReorganiser`, although I'm not familiar with how it works.

Thumbs/ThumbsMiddleware.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            else
+            {
+                // mode for debugging etc
+                switch (context.Request.Query["mode"])
+                {
+                    case "dump":
+                        await WriteRequestDump(context, thumbnailRequest);
+                        break;
+                    default:
```
stephenwf: Should we disable this in production? Or is it going to be useful? Anything in the context that would be suitable for responding with?

DLCS.Repository/Assets/ThumbRepository.cs
```diff
@@ -0,0 +1,155 @@
...
+                return (List<int[]>) serializer.Deserialize(jsonTextReader, typeof(List<int[]>));
+            }
+        }
+
+        private List<ThumbnailPolicy> GetThumbnailPolicies()
+        {
+            const string key = "ThumbRepository_ThumbnailPolicies";
+            return memoryCache.GetOrCreate(key, entry =>
+            {
+                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(10);
```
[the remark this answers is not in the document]
tomcrane: It's VERY long-lived, months, years... but if you added a new one you'd want it available straight away. Need better strategies for this kind of lookup.

IIIF/ImageApi/ImageRequest.cs
```diff
@@ -0,0 +1,63 @@
...
+                request.IsInformationRequest = true;
+                return request;
+            }
+
+            request.Region = Region.Parse(parts[1]);
+            request.Size = Size.Parse(parts[2]);
+            request.Rotation = Rotation.Parse(parts[3]);
+            var filenameParts = parts[4].Split('.');
+            request.Quality = filenameParts[0];
+            request.Format = filenameParts[1];
```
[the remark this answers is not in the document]
tomcrane: This needs lots of exception handling, it's very much a happy path at the moment. Just want to make sure the path is the right one.

Thumbs/ThumbsMiddleware.cs
tomcrane: probably useful in prod. Shows how the image request was parsed.

Thumbs/InfoJsonBuilder.cs
tomcrane: Actually this could be a bit cleverer and use https://github.com/dlcs/protagonist/blob/thumbs/IIIF/ImageApi/Size.cs#L107

tomcrane: The info.json generation is not exactly elegant atm, it's stringtastic. But, fast.

Thumbs/ThumbsMiddleware.cs
```diff
@@ -0,0 +1,122 @@
...
+            ILogger<ThumbsMiddleware> logger,
+            IThumbRepository thumbRepository)
+        {
+            this.configuration = configuration;
+            this.cacheSeconds = configuration.GetValue<int>("responseCacheSeconds", 0);
+            this.logger = logger;
+            this.thumbRepository = thumbRepository;
+        }
+
+        public async Task Invoke(HttpContext context,
```
donaldgray: Is it worth having this `public Task Invoke(...` and then omitting the `await` statements inside and instead return the `Task<T>`. E.g. `return RedirectToInfoJson(context);` vs `await RedirectToInfoJson(context);`. 

If performance is key, this would save framework creating a statemachine under the covers - would need tested to verify that it actually makes a difference.

Other than perf the only real difference is that exceptions would be caught by caller, which is framework in this instance so wouldn't make a difference here. However, `await` may make sense if we're wanting to handle exceptions inside this function.

Thumbs/ThumbsMiddleware.cs
```diff
@@ -0,0 +1,122 @@
...
+            var thumbInBucket = thumbRepository.GetThumbLocation(
+                request.Customer.Id, request.Space, request.IIIFImageRequest);
+            context.Response.ContentType = "image/jpeg";
+            SetCacheControl(context);
+            await bucketReader.WriteObjectFromBucket(await thumbInBucket, context.Response.Body);
+        }
+
+        private static async Task WriteRequestDump(HttpContext context, ThumbnailRequest request)
+        {
+            await context.Response.WriteAsync(JsonConvert.SerializeObject(request));
```
donaldgray: As above, for perf (if it does indeed help) this await can be omitted. 
```csharp
private static Task WriteRequestDump(HttpContext context, ThumbnailRequest request)
  => context.Response.WriteAsync(JsonConvert.SerializeObject(request));
```

DLCS.Repository/Assets/AssetRepository.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+
+    public class AssetRepository : IAssetRepository
+    {
+        private readonly IConfiguration configuration;
+        private readonly ILogger<AssetRepository> logger;
+
+        public AssetRepository(IConfiguration configuration, ILogger<AssetRepository> logger)
+        {
+            this.configuration = configuration;
```
donaldgray: Use [options pattern](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1) rather than pass around `IConfiguration`?

Thumbs/ThumbsMiddleware.cs
tomcrane: This is something we should dissect next week.

DLCS.Repository/Assets/AssetRepository.cs
tomcrane: Again - this slice through a potential arrangement of the stack using a very simple service allows us to decide our patterns and practices on something manageable.

PR #6, opened 2020-02-23
handle different thumb paths

tomcrane:
This feels a bit dumb - best way to have same service handle alternative paths for testing?


tomcrane:
Orchestrator:
https://dlcs-demo.com/thumbs/2/1/thumbsizetest/full/!400,400/0/default.jpg

Thumbs:
https://dlcs-demo.com/thumbschk/2/1/thumbsizetest/full/!400,400/0/default.jpg
https://dlcs-demo.com/thumbsnochk/2/1/thumbsizetest/full/!400,400/0/default.jpg

PR #7, opened 2020-03-17
Update to handle authenticated/open thumbs paths

donaldgray:
Quite a few changes, the main ones being related to handling of new `/open` and `/auth` paths that thumbs are generated in. Only thumbs that are in `/open` will be returned. A 404 will be returned for `/auth` thumbs. Updated `ThumbReorganiser` to generate the new format and clean up previous bounded square thumbs.

In summary
* Added `GetAvailableThumbSizes`, migrated from previous app.
* Rework login in `ThumbReorganiser` to use `/open` and `/auth` paths and tidy up previous bounded thumbs (e.g. `/100.jpg`, `/200.jpg` etc). Added tests.
* Refactored `ThumbReorganiser` to implement an interface and be injected. Introduced `IThumbnailPolicyRepository` and took this out of the `ThumbRepository` to avoid a circular reference between `ThumbRepository` and `ThumbReorganiser`.
* Refactored `IBucketReader.WriteObjectFromBucket` to `IBucketReader.GetObjectFromBucket` to return a stream rather than write to provided stream as it would allow us to determine if an object has been found or not without resorting to throwing an exception.
* Helper methods for dealing with collections and guard checks.
* Added new Middleware to handle `HttpExceptions` and render correct status code.

Lastly refactored some of the models in IIIF project, main changes:
  * Renamed `Region` -> `RegionParameter`
  * Renamed `Size` -> `SizeParameter`
  * Renamed `Rotation` -> `RotationParameter`
  * Introduced `Size` class which is only `Width`, `Height` and some related (e.g. `IsConfinedWithin`, `Confine`) etc


donaldgray:
Merged #8. Will look at how to handle ParameterStore when running locally. Will also add new commit with JenkinsFile

PR #15, opened 2020-06-10
first versions of c4 diagrams

tomcrane:
Slight tweak to diagram to fix an error; time for this to be in the protagonist repo I think.


tomcrane:
Updated

PR #20, opened 2020-07-28
Add cors to thumb responses

donaldgray:


Thumbs/Startup.cs
```diff
@@ -31,7 +34,15 @@ public void ConfigureServices(IServiceCollection services)
...
                 .AddNpgSql(Configuration.GetPostgresSqlConnection());
-            services.AddCors();
+            services.AddCors(options =>
+            {
+                options.AddPolicy(corsPolicy,
+                    builder => builder
+                        .AllowAnyMethod()
+                        .AllowAnyHeader()
+                        .SetIsOriginAllowed(host => true)
+                        .AllowCredentials());
```
tomcrane, line 44: What ACAO header does this set? Does it set * or does it set <your-host>?
I think we can drop the .AllowCredentials() - If you set the allow credentials header, and a client sends credentials, it puts additional constraints on the ACAO header, it can't be *.
As thumbs never needs credentials, I think we can ignore this?

I wish the MS docs were useful and discussed what the call actually does at the HTTP level.

donaldgray: When testing locally this adds: 
```
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: {origin-domain}
```

It will only add the header if the origin differs from domain (didn't initially realise this when testing locally!). There's logic in the middleware to work this out (added by `UseCors()`). If the origins differ the `.SetIsOriginAllowed(host => true)` delegate is called, to allow specific logic to be run.

Agreed that we can remove the credentials. Will push that in another change.

donaldgray:
Closing this, using: https://github.com/dlcs/protagonist/pull/21 instead as it is more cacheable and doing less

PR #55, opened 2020-11-10
Add API project and allow image to be ingest by bytes

donaldgray:
Add new API project, this is very minimal just now and hands off auth + actual ingest request to existing API. 

This adds a single `POST /customers/{customer}/spaces/{space}/images/{image-name}` endpoint that takes base64 encoded file in new `file` property, alongside any other properties that can be sent to existing PUT image endpoint. The bytes are saved to S3, and then the JSON body is forwarded to the existing API with Origin set to point to new S3 location.

### DIFF
--- a/docs/adr/0000-api-project-design.md
+++ b/docs/adr/0000-api-project-design.md
@@ -17,7 +17,7 @@ How best can we structure the API project to make it easy for developers to navi
 
 ## Decision Outcome
 
-"Organise by Feature and use MediatR", because we can use MediatR to encapsulate the various use-cases of the API. Together with feature folders this should allow developers to more easily identify the capabilities of the API and target the section they need to,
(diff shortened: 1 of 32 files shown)


docs/adr/0000-api-project-design.md
```diff
@@ -0,0 +1,93 @@
...
+How best can we structure the API project to make it easy for developers to navigate and comprehend.
+
+## Considered Options
+
+* Organise by Feature and use MediatR
+* Traditional ASP.NET/MVC layout
+
+## Decision Outcome
+
+"Organise by Feature and use MediatR", because we can use MediatR to encapsulate the various use-cases of the API. Together with feature folders this should allow developers to more easily identify the capabilities of the API and target the section they need to,
```
tomcrane, line 20: full stop?

tomcrane (approved):

docs/adr/0000-api-project-design.md
donaldgray: Fixed.

PR #87, opened 2021-06-02
List of images in Space, for portal, with stub image page

tomcrane:


API/JsonLd/HydraImageCollection.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace API.JsonLd
+{
+    public class HydraImageCollection : HydraCollectionBase
```
donaldgray, line 5: Could this be `: SimpleCollection<Image>`

Portal/Pages/Spaces/Details.cshtml
```diff
@@ -1,11 +1,43 @@
...
+    </tr>
+    </thead>
+    <tbody>
+    @foreach (var image in Model.SpacePageModel.Images.Members)
+    {
+        <tr>
+            <td>
+                @if (image.Family == 'I')
+                {
+                    <iiif-thumb
```
donaldgray, line 25: `<iiif-thumb />` is nice! 👍

donaldgray (approved):

API/JsonLd/HydraImageCollection.cs
tomcrane: It gets the PartialView from the hydraCollectionbase

PR #89, opened 2021-06-02
Dropzone...

tomcrane:
Does not do anything with DLCS yet
Uploads images you drag in from your desktop to /tmp
Does not work with dragged in images from web pages, need to investigate why... the drop event might be different since iiif.ly

DropxoneController accepts POSTed files, and also XHR POSTed external.json which are placeholders for dragged-in web images (the bit that doesn't work yet).


Portal/Pages/Spaces/Details.cshtml
```diff
@@ -6,6 +6,21 @@
     ViewData["Title"] = $"Space {Model.SpacePageModel.Space.GetLastPathElementAsInt()}: {Model.SpacePageModel.Space.Name}";
 }
 
+@section header
+{
+    <link href="~/lib/dropzone-5.7.0/dist/min/dropzone.min.css" rel="stylesheet" />
+}
```
stephenwf, line 12: The minified assets are available on CDN if you want to avoid source controlling the modules
https://www.jsdelivr.com/package/npm/dropzone?path=dist%2Fmin

Portal/Controllers/DropzoneController.cs
```diff
@@ -0,0 +1,101 @@
...
+            var targetDir = Path.Combine(Path.GetTempPath(), customer.ToString(), space.ToString());
+            Directory.CreateDirectory(targetDir);
+            var target = Path.Combine(targetDir, fileNameForSaving);
+            return target;
+        }
+
+
+        [HttpPost]
+        [Route("[controller]/{customer}/{space}/[action]")]
+        public async Task<IActionResult> External(int customer, int space, [FromBody] ExternalImage image)
```
donaldgray, line 71: Is it worth omitting `{customer}` slug and always just take from cookie?

Portal/Controllers/DropzoneController.cs
```diff
@@ -0,0 +1,101 @@
...
+        {        
+            if (currentUser.GetCustomerId() != customer)
+            {
+                throw new InvalidOperationException("Customer ID mismatch");
+            }  
+            if (string.IsNullOrWhiteSpace(image.ExternalUrl))
+            {
+                return Json(new { error = "No external image Url in request" });
```
donaldgray, line 79: Return 400 rather than 200?

donaldgray (approved):

Portal/Pages/Spaces/Details.cshtml
tomcrane: Will do

Portal/Controllers/DropzoneController.cs
tomcrane: I wasn't sure about that... it's there for security, obvs - no uploading to other people's spaces - but also to inform when the request isn't constructed properly "I meant cust 2 space 6 and I thought I'd supplied the right cookie..."

Portal/Controllers/DropzoneController.cs
tomcrane: The UI on the client will handle this (if I can work out how to trigger it)

PR #93, opened 2021-06-04
Asset ordering

tomcrane:
Add jquery-ui for sortable(), but a custom build that only includes what we need.
Added buttons that also move the table rows about.

Both mechanisms re-populate a form field in each row - currently visible for debugging but will be `hidden` on tidy up later.

These form elements get processed on the server: we do a batch patch operation by getting the current space images and updating their Number1 fields from the form elements, to match the new order.


Portal/Pages/Spaces/Details.cshtml
```diff
@@ -42,51 +45,83 @@
...
             </td>
             <td>@image.Number1</td>
-            <td><a asp-page="/Images/Index" 
-                   asp-route-space="@Model.Space" 
+            <td>
+                <a asp-page="/Images/Index"
+                   asp-route-space="@Model.Space"
                    asp-route-image="@image.GetLastPathElement()">@image.GetLastPathElement()</a></td>
             <td>@image.Created</td>
+        @if (Model.SpacePageModel.Space.IsManifestSpace())
```
donaldgray, line 91: `IsManifestSpace` is on the viewmodel now (`Model.SpacePageModel.IsManifestSpace`) but it does the exact same thing.

donaldgray (approved):

PR #113, opened 2021-09-09
Single asset generation from /iiif-manifest/ route

donaldgray:
Add new route for `/iiif-manifest/{cust}/{space}/{asset}` which will generate a single asset manifest. This only works for Image assets, as per current version. Refactored `ImageController` and introduced `IIIFAssetControllerBase` base class as the error handling/response handling for info.json and manifests are very similar.

Took this opportunity to update info.json generation to use models from `iiif-net` nuget package, rather than string manipulation. This is still a WiP, only those methods called have been updated.

Updated `Asset` entity to store `AssetFamily` enum, rather than just a char.

> Note - this is merging into branch for #112 as that's required to use iiif-net nuget package.

### DIFF
--- a/DLCS.Model/Assets/InfoJsonBuilder.cs
+++ b/DLCS.Model/Assets/InfoJsonBuilder.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using System.Text;
 using IIIF;
 using IIIF.ImageApi.Service;
-using IIIF.Presentation.V2;
 
 namespace DLCS.Model.Assets
 {
@@ -19,7 +18,7 @@ namespace DLCS.Model.Assets
         /// <param name="serviceEndpoint">URI for image</param>
         /// <param name="sizes">List of sizes image is available in.</param>
         /// <returns>info.json object</returns>
-        public static JsonLdBase GetImageApi2_1Level0(string serviceEndpoint, List<int[]> sizes)
+        public static ImageService2 GetImageApi2_1Level0(string serviceEndpoint, List<int[]> sizes)
         {
             var imageService = new ImageService2
             {
@@ -47,7 +46,7 @@ namespace DLCS.Model.Assets
         /// <param name="width">Width of image</param>
         /// <param name="height">Height of image</param>
         /// <returns>info.json string</returns>
-        public static JsonLdBase GetImageApi2_1Level1(string serviceEndpoint, int width, int height,
+        public static ImageService2 GetImageApi2_1Level1(string serviceEndpoint, int width, int height,
             List<int[]> sizes)
         {
             var imageService = new ImageService2
--- a/DLCS.Model/Assets/InfoJsonBuilder.cs
+++ b/DLCS.Model/Assets/InfoJsonBuilder.cs
@@ -2,6 +2,9 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using IIIF;
+using IIIF.ImageApi.Service;
+using IIIF.Presentation.V2;
 
 namespace DLCS.Model.Assets
 {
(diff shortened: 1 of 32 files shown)


DLCS.Model/Assets/InfoJsonBuilder.cs
```diff
@@ -48,31 +46,26 @@ public static string GetImageApi2_1Level0(string serviceEndpoint, List<int[]> si
...
+                    Qualities = new[] { "native", "color", "gray" },
+                    Supports = new[] { "regionByPct","sizeByForcedWh","sizeByWh","sizeAboveFull","rotationBy90s","mirroring","gray" }
+                }
+            };
+            SetSizes(imageService, sizes, width, height);
+            SetScaleFactors(imageService);
+            return imageService;
         }
         
         public static string GetImageApi2_1Level1Auth(string serviceEndpoint, int width, int height, List<int[]> sizes, string services)
```
tomcrane, line 71: Eventually here I think we should ask Cantaloupe what info.json it would produce for this image, and then store the relevant variable bits of info. We'd still use the IIIF lib because we'd want to decorate the info.json with extra Orchestrator-level stuff like Auth. The reason for this is the tile size and scaleFactors - we assume always 256-size tiles and I don't want to assume that. But for now this is nice.

DLCS.Repository/Assets/DapperAssetRepository.cs
```diff
@@ -52,7 +49,54 @@ public async Task<ImageLocation> GetImageLocation(AssetId assetId)
...
+                await using var connection = await DatabaseConnectionManager.GetOpenNpgSqlConnection(configuration);
+                dynamic? rawAsset = await connection.QuerySingleOrDefaultAsync(AssetSql, new { Id = id });
+                if (rawAsset == null)
+                {
+                    entry.AbsoluteExpirationRelativeToNow =
+                        TimeSpan.FromSeconds(cacheSettings.GetTtl(CacheDuration.Short));
+                    return NullAsset;
+                }
 
+                return new Asset
```
tomcrane, line 68: We'll need a bit of a tidy up of the various representations of this DB row that are present in different bits of the code!

Orchestrator/Features/Manifests/GetManifestForAsset.cs
```diff
@@ -0,0 +1,163 @@
...
+using IIIF.Presentation.V2.Strings;
+using IIIF.Serialisation;
+using MediatR;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Orchestrator.Infrastructure.Mediatr;
+using Orchestrator.Models;
+using Orchestrator.Settings;
+
+namespace Orchestrator.Features.Manifests
```
tomcrane, line 22: Looking forward a bit - one of the St Louis tasks is IIIF v3 for named queries (and single asset manifest). That's when the Handle method below will need to make a canvas appropriate to the asset type (it only works for images atm).

In future, if the asset is not an image and you ask for the v2 version, the DLCS should just redirect to the v3 version I think. It certainly won't do the IxIF fake stuff like Wellcome.

Test.Helpers/Integration/DatabaseTestDataPopulation.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static class DatabaseTestDataPopulation
+    {
+        public static ValueTask<EntityEntry<Asset>> AddTestAsset(this DbSet<Asset> assets, 
+            string id, 
+            AssetFamily family = AssetFamily.Image,
+            int customer = 99,
+            int space = 1,
+            string origin = "",
+            string roles = "",
+            string mediaType = "imag/jpeg",
```
tomcrane, line 18: typo! image/jpeg

tomcrane (changes requested):
Mostly some general comments, there is one typo though - (see comment)

Orchestrator/Features/Manifests/GetManifestForAsset.cs
donaldgray: Yeah - I'm not sure whether the request to this handler should have a `Version` param, or whether it should be 2 different request/handlers, one for each version, and the Controller can decide which to dispatch.

DLCS.Model/Assets/InfoJsonBuilder.cs
donaldgray: At some point it might be better to make make this class non-static to ease the transition to Cantaloupe.

Test.Helpers/Integration/DatabaseTestDataPopulation.cs
donaldgray: Fixed

tomcrane (approved):

PR #116, opened 2021-09-14
Signup link

tomcrane:
Includes DB migration


API.Client/AdminDlcsClient.cs
```diff
@@ -0,0 +1,68 @@
...
+            jsonSerializerSettings = new JsonSerializerSettings
+            {
+                NullValueHandling = NullValueHandling.Ignore,
+                ContractResolver = new CamelCasePropertyNamesContractResolver()
+            };
+        }
+
+        public void SetBasicAuth(string basicAuth)
+        {
+            httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", basicAuth);
```
donaldgray, line 34: I think you could use `DLCS.Web.Requests.HttpRequestHeadersX.AddBasicAuth(this HttpRequestHeaders headers, string base64Encoded)` here, although i'm not sure if it's worth it as it's a single line.

donaldgray (approved):

Issue #156, opened 2021-10-04
Handle BearerTokens for DescriptionResources


donaldgray:
Validate provided Bearer token when request made for description resource.

The bearer token can serve to lookup available roles in DB. Validate that auth roles match asset roles.

See [IIIF Auth Spec](https://iiif.io/api/auth/1.0) for more details.

donaldgray:
As per call with @tomcrane there is no need to do validation check on `/iiif-manifest/`, only info.json. The [Authentication for description resources](https://iiif.io/api/auth/1.0/#authentication-for-description-resources) in auth spec mentions manifests but these are mentioned as an example, rather than what should require auth.

PR #176, opened 2021-10-07
Handle Bearer token on GET request for Description Resources

donaldgray:
Update GET `/iiif-img/{cust}/{space}/{image}/info.json` and GET `/iiif-manifest/{cust}/{space}/{image}` to validate Bearer token header if asset is restricted.

Check is done in `GetImageInfoJson` and `GetManifestForAssetHandler` Mediatr handlers, respectively.

Introduced 2 new classes to help with checks:
* `AccessChecker` - checks wither specified `SessionUser` can access at least 1 of the roles (from `"Images"."Roles"`)
* `AssetAccessValidator` - validates whether current `HttpContext` contains bearer-token, and whether that can access image (using above class). If item is restricted and user has access then `Set-Cookie` header is added to Response, using existing `AuthCookieManager`. This class just has `TryValidateBearerToken` for now but will gain `TryValidateCookie` in future.

Modified the cached `OrchestrationImage` to store list of Roles to make this easier (`RequiresAuth` is derived from that property now).

Renamed `RequestHandlerBase` -> `AssetRequestProcessor` and switched this to composition, rather than inheritence, for Image + Timebased handlers. Originally I was using this for auth checks but opted for another method but left in as it's cleaner without inheritance.

Renamed `IIIFJsonResponse` -> `DescriptionResourceResponse`.

Fix for #156


Orchestrator/Infrastructure/Auth/AssetAccessValidator.cs
```diff
@@ -0,0 +1,106 @@
...
+            
+            // Validate current user has access for roles for requested asset
+            var canAccess = await accessChecker.CanSessionUserAccessRoles(authToken.SessionUser, customer, assetRoles);
+
+            if (!canAccess)
+            {
+                return AssetAccessResult.Unauthorized;
+            }
+            
+            cookieManager.SetCookieInResponse(authToken);
```
tomcrane, line 76: DLCS validates access for 2 types of request:
 - info.json requests made by script that won't include any cookies the user might have but might include a bearer token in the authorisation header - where the client acquired this token via postMessage when it opened the token service in an iframe. Direct HEAD requests are also made this way by script clients because there's no info.json for AV
 - image/av requests made by anything else - e.g., img tag, openseadragon - which will include any cookies the user has acquired.

Image requests won't have a bearer token in them (how would an img tag set that?). At least in auth 1.0 they won't.
So I think this line can just be removed. In fact it's less secure; a client that only has a bearer token (maybe acquired with bad intent) could cause the DLCS to set a cookie for the image.

PR #182, opened 2021-10-10
Draft version of basic auth handler

tomcrane:
This replaces the delegated version.
Extends the Dapper customer repository to look up customers by key. Heavy API use shouldn't hit the DB.
Still uses the Deliverator approach to key and secrets.

Makes some assumptions about how different paths are authed that needs validating.

The ingestFromFile hasn't been touched yet, just the auth part.


API/Auth/DlcsBasicAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            this.customerRepository = customerRepository;
+            this.deliveratorApiAuth = deliveratorApiAuth;
+        }
+
+        protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
+        {
+            // skip authentication if endpoint has [AllowAnonymous] attribute
+            // Not all API paths require auth...
+            var endpoint = Context.GetEndpoint();
+            if (endpoint?.Metadata?.GetMetadata<IAllowAnonymous>() != null)
```
donaldgray, line 56: Is this necessary? I thought that `[AllowAnonymous]` would mean that this scheme wasn't called at all?

API/Auth/DlcsBasicAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                if (int.TryParse(customerIdRouteVal.ToString(), out int result))
+                {
+                    resourceCustomerId = result;
+                }
+            }
+            var authHeader = Encoding.UTF8.GetString(Convert.FromBase64String(headerValue.Parameter));
+            string[] keyAndSecret = authHeader.Split(':');
+            var apiCaller = new {Key = keyAndSecret[0], Secret = keyAndSecret[1]};
+            var customerForKey = await customerRepository.GetCustomerForKey(apiCaller.Key, resourceCustomerId);
```
donaldgray, line 82: Replace lines 79-81 with a single call? `var apiCaller = await GetApiCaller(headerValue.Parameter)`?

_Really_ slim chance of it happening but this method could have error handling to deal with a invalid auth header (e.g. something without ":" as that would throw an exception as-is).

API/Auth/DlcsBasicAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            // Some of these could be classed as authorisation rather than authentication, though they are not user-specific.
+            if (resourceCustomerId.HasValue)
+            {
+                // the request is for a particular customer's resource (e.g., an asset)
+                if (customerForKey.Id != resourceCustomerId.Value)
+                {
+                    // ... but the requester is not this customer. Only proceed if they are admin.
+                    if (!customerForKey.Administrator)
+                    {
+                        return AuthenticateResult.Fail("Only admin user may access this customer's resource.");
```
donaldgray, line 103: Is it possible to end up here? If I'm customer 10 (not admin) and try to access customer 11's resource, wouldn't `customerForKey` be null?

donaldgray, line 103: Is it possible to end up here? If I'm customer 10 (not admin) and try to access customer 11's resource, wouldn't `customerForKey` be null?

API/Auth/DlcsBasicAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            
+            // At this point our *authentication* has passed.
+            // Downstream handlers may still refuse to *authorise* the request for other reasons
+            // (it can still end up a 401)
+            Logger.LogInformation("Authentication passed");
+
+            var claims = new List<Claim>
+            {
+                new (ClaimTypes.Name, customerForKey.Name), // TODO - should these be the other way round?
+                new (ClaimTypes.NameIdentifier, apiCaller.Key), // ^^^
```
donaldgray, line 127: I think `new (ClaimTypes.Name, customerForKey.Name)` is correct, iirc `ClaimTypes.Name` is for display. `ClaimTypes.NameIdentifier` is for immutable identifier. I wonder if apikey is neither of those..

Issue #153, opened 2021-10-04
Portal pause and evaluate


tomcrane:
- Should Portal be allowed to use the database at all? Or should everything be done through API, extending API where needed?
 - If Portal can use DB, it should share Mediatr commands with API project
 - Review and rationalise conventions for Mediatr
 - Review and Rationalise testing strategy for Portal
 - Rationalise API client usage #152

tomcrane:
Summary - might need to make issues for these

Portal should only use the DB for signup links and other things that are above the level of the DLCS and could just as well be done in an external DB. Anything that we find useful in Portal driving DLCS, other people are likely to find useful building their own apps, dashboards etc. So we should dogfood everything and Portal should not talk to the DB.

This means the API rewrite can't wait until after the portal rewrite - needs to be simultaneous. There are issues with API that need to be fixed, and everything the Portal needs, the API needs.

* @tomcrane to reproduce enough API required to drive portal, then review approach and revise if necessary. When developing locally running portal calls a locally running API, rather than a deployed API.
* Mediatr in general - don't separate into requests/commands folders, group all into /../Requests/*
* In API, Mediatr requests talk to DB and return (usually) JSON-LD to controllers, so the controller is a very light web-layer wrapper.
* In Portal, Mediatr requests call API and either pass through the JSON-LD response, or (more usual perhaps) convert to more suitable model classes.
* These decisions make the page/controller very light, so we test the Mediatr handling and not the final view part.
* Move JSON-LD API classes into a separate class library project API.JsonLD, where they can be used by API and API.Client
* Use Entity Counters properly; space and customer are identity seeds, customer-images and space-images are counts that need to be re-counted after changes.

tomcrane:
This issue is out of date (and mostly done)

PR #199, opened 2021-10-22
Portal uses new api

tomcrane:
This is a partial implementation of the new API, for experimenting with different approaches.
It's likely that this PR will never be merged as-is but be split into smaller PRs that can be added gradually.

The new API needs to reproduce the ability to POST a file directly. This feature started out as the ONLY feature of the new API, and it just called the Deliverator API after placing the Asset in a bucket. But obviously nothing here can make calls to the Deliverator API, so we need at least that working natively before we can start deploying the new API to production environments.

The process of producing this PR so far involved adapting the equivalent Deliverator implementation, sometimes borrowing literal SQL directly, sometimes adapting it for EF.

From https://github.com/dlcs/protagonist/issues/153#issuecomment-933640506 - with edits

- API no longer has a reference to API.Client
- reproduce enough API required to drive portal, then review approach and revise if necessary. When developing locally running portal calls a locally running API, rather than a deployed API.
- Mediatr in general - don't separate into requests/commands folders, group all into /../Requests/*
- In API, Mediatr requests talk to DB, return mapped DB objects to controllers, which then convert them to Hydra JSON-LD. The controller is a very light web-layer wrapper.
- In Portal, Mediatr requests call API and either pass through the Hydra JSON-LD response, or (more usual perhaps) convert to more suitable model classes.
- These decisions make the page/controller very light, so we test the Mediatr handling and not the final view part.
- Move JSON-LD API classes into a separate class library project `DLCS.HydraModel`, where they can be used by API and API.Client
- Use Entity Counters properly; space and customer are identity seeds, customer-images and space-images are counts that need to be re-counted after changes.
- This PR adapts DLCS.Mock and DLCS.HydraModel from the 2016 DLCS prototype. DLCS.Mock provides a fake back-end for testing, and can also be used to generate documentation from attributes on the Hydra classes, in HTML and Markdown.

This PR uses a mixture of EF and Dapper, sometimes in the same Mediator command. This is not ideal.

Questions

Why do we have context objects living in two different namespaces?
 `DLCS.Model.*.*` and `DLCS.Repositories.Entities.*`


tomcrane:
Closing this PR as its changes are now in the develop branch, via many smaller PRs!

Issue #30, opened 2020-10-25
SpecialServer


tomcrane:
Handles requests for /full/ images that Orchestrator hasn't matched to a thumb size.
May still forward on to image server cluster. Or might do some other image processing directly, possibly even from S3.
Avoid mixing /full/ load with tile load.

.NET Core service.

tomcrane:
Have a good look at https://github.com/imazen/imageflow-dotnet-server

Consider contributing an IIIF handler to it.

tomcrane:
Could Engine use SpecialServer to make thumbs? Instead of or as well as Appetiser?
And could it do them on demand, after the fact - e.g., if you added some extra sizes to the thumbnail policy for an image. The Thumbs service, just before it returns a 404 for a non existent thumb size, could check the policy and see if it is actually a supported size and create a new thumb on demand - either by reducing the next thumb size up using the above imageflow, or if the new size is bigger than any existing thumbs, by invoking imageflow on the source image?

NB imageflow does NOT support JP2 and has no plans to. 
But SpecialServer might have its own personal Cantaloupe / IIP to call on, as well, maybe...

tomcrane:
Further notes on this:

/full handling: https://github.com/dlcs/protagonist/blob/master/docs/architectural-considerations.md#dealing-with-full-image-requests
/full thumbnails: https://github.com/dlcs/protagonist/issues/39

donaldgray:
> Have a good look at https://github.com/imazen/imageflow-dotnet-server
> 
> Consider contributing an IIIF handler to it.

As imageflow doesn't support JP2 this doesn't seem like an option, instead use a custom Cantalopue with an s3 source

donaldgray:
Implemented this via the [Cantaloupe delegate system](https://cantaloupe-project.github.io/manual/5.0/delegate-system.html). 

This will allow us to use an S3Source in Cantaloupe, with the image source being `ImageLocation.S3`. The delegate will parse this to get bucket + key.

Added this to the same Cantaloupe image that we are already using as the change is only adding a `delegates.rb` file to the image - the usage of it is then controlled via properties.

PR #212, opened 2021-11-03
Architecture and design for implementation of a handler for consuming "composite" formats

danielgrant:


tomcrane (commented):
In the flow chart, the incoming payload might be a Collection (POST to a queue) or an individual asset (PUT to location). The decision to invoke the PDF handler would be based on whether the member(s) of the Collection, or the single PUTted asset, are of type vocab:Composite.
This is slightly different to the flow in the diagram I think.
We can have an extra condition for this work - you can't mix composites with other types in a single API call. You can't post a collection to a queue that has an image and a composite in it. Do that in two separate calls please!

danielgrant:
> In the flow chart, the incoming payload might be a Collection (POST to a queue) or an individual asset (PUT to location). The decision to invoke the PDF handler would be based on whether the member(s) of the Collection, or the single PUTted asset, are of type vocab:Composite. This is slightly different to the flow in the diagram I think. We can have an extra condition for this work - you can't mix composites with other types in a single API call. You can't post a collection to a queue that has an image and a composite in it. Do that in two separate calls please!

I've updated this diagram to make it a bit simpler now by requiring the external system which invokes the API to call the Composite Handler directly, instead of having to go via the API. I believe this is what we discussed, and it removes the need to implement any modifications in that API component.

docs/adr/0001-composite-handler-design.md
```diff
@@ -0,0 +1,73 @@
...
+- Artifact processing / rasterization code already highly optimised.
+
+#### Negative Consequences
+
+- Requires modification of multiple existing components, i.e.
+  Orchestrator, thumbnail generation, etc.
+- Potentially a heavyweight solution, Cantaloupe image server can be
+  resource intensive.
+- Results in two image servers deployed within the DLCS - IIPImage and
+  Cantaloupe - with signficant effort to remove IIPImage entirely.
```
stephenwf, line 69: I guess vendor lock-in would also be a negative too

danielgrant: Good point - updated.

docs/adr/0001-composite-handler-design.md
```diff
@@ -0,0 +1,73 @@
...
+which can process "composite" artifacts. Such artifacts typically take
+the form of a single object containing multiple embedded resources, each
+of which can be rasterized and treated as an individual resource within
+the DLCS.
+
+## Considered Options
+
+- Additional standalone component which consumes a DLCS ingest payload,
+  retrieves the object, rasterizes accordingly, and then invokes the
+  existing DLCS ingestion workflow.
```
stephenwf, line 25: Could the standalone tool use cantaloupe to rasterize? That could make it easier if we wanted to switch or had to switch down the line. Or does that defeat the purpose of having it standalone?

danielgrant: When we discussed this, that was tabled as an option, but I think in doing so we end up doing more work than we have to - deployment, configuration and maintenance of Cantaloupe, in addition to the development of this additional standalone component which simply proxies through to that image server.

Contrast this with implementing rasterization within the handler - which _should_ be a relatively simple task - and to me I don't think it makes sense to do both.

tomcrane: Yeah - we _could_ do this, but I think we'll come back to it later when we have Cantaloupe as a part of the DLCS.

There is a related issue here: https://github.com/dlcs/protagonist/issues/30#issuecomment-721274027

...where Cantaloupe might be used at ingest time as well as at end-user-facing runtime, to generate thumbs from sources without having to move them from S3.

donaldgray:
LGTM. Agree this reflects what we discussed on the call. We could revisit and go via the API once we've ported it over to .net 5 (likely 6 by the time it's done) but what's in the flowchart will definitely work.

One observation on the flowchart - doesn't necessarily affect the viability of the suggested solution but could have an affect on longer term maintenance - is the _Extract the origin URI and retrieve the artifact..._ step. Currently the Engine component does this by using a set of know "Origin Strategies" (http, s3, ftp) and a DB lookup to determine which strategy to use. Adding logic to handle each of these to the CompositeHandler isn't particularly complex but does involve having that logic in multiple places and we'd need to keep them in sync. Until now the ingest process has been for the engine to move an artifact from Origin->LocalStorage then handing off to another component for processing. 

Given that this component is driven by a client requirement we could just implement that single strategy (likely S3) that they need and can expand later.

danielgrant:
> One observation on the flowchart - doesn't necessarily affect the viability of the suggested solution but could have an affect on longer term maintenance - is the _Extract the origin URI and retrieve the artifact..._ step. Currently the Engine component does this by using a set of know "Origin Strategies" (http, s3, ftp) and a DB lookup to determine which strategy to use. Adding logic to handle each of these to the CompositeHandler isn't particularly complex but does involve having that logic in multiple places and we'd need to keep them in sync. Until now the ingest process has been for the engine to move an artifact from Origin->LocalStorage then handing off to another component for processing.
> 
> Given that this component is driven by a client requirement we could just implement that single strategy (likely S3) that they need and can expand later.

Good point. As you note, I think we can do one of two things here:

- Duplicate the origin strategies logic in this component - which requires DB interaction - to accommodate all potential strategies; or
- Implement the only strategy that we know we need to meet the specific client requirement that is driving this piece of work.

I'm including to do the latter, and kick the former into the long grass until the API is updated to accept composite requests. When we do this work, we could have the API handle retrieval of the artifact from the origin and place it into an accessible location within the DLCS, and then delegate the request to the CompositeHandler. This way, the API continues to do the "heavy lifting" of artifact retrieval, and there's no duplication of that logic.

Something to think about.

tomcrane:
I'm OK to approve the RFC part of this PR once it's separate from the Composite handler implementation part.

On the format strings - it's easier to insist that the format string syntax is that of the code it's going to hit in the DLCS, which is now Python - so it's easier for us, the onus is on the caller to get it right rather than for us to transform it into a different language format.

And, it's probably more likely that clients will be scripting ingests in Python than in C#, so it's good from that POV.

Later, if/when we make the DLCS API accept and process the composite directly, we will need to transform the syntax to C# format strings. Maybe there's a nice library to do that (couldn't see one from a quick Google).

PR #213, opened 2021-11-03
Add docker image build + push to github actions

donaldgray:
Modified existing github actions. Previously this just ran `dotnet test` but it now runs `dotnet test` then calls other `build_*.yml` files to build + push docker images. The docker files will run in parallel after tests have run [example](https://github.com/dlcs/protagonist/actions/runs/1420561777). 

The reference to other workflow files is `dlcs/protagonist/.github/workflows/build_thumbs.yml@deploy_container` - I'll need to push a fix once this is merged to use a different github ref (`@develop` then `@main` once it's merged). I couldn't get relative reference working.

Decided on `build_*.yml` file per project for now - these have `env` vars at the top. Adding a new image should just be a case of copying an existing one and modifying these vars. I left these as separate files, rather than reusing a parameterised file, for now to see if they need to deviate from each other. If not we can easily switch to using a single parameterised file.

Tags created are (Docker file is build but not pushed for PRs):
* branch name when push to branch (e.g. `develop`, `main`)
* long sha1
* `{version}`, `{major}.{minor}` and `{major}` (e.g. 1.2.3, 1.2 and 1) when version tag pushed.

Move jenkins file to `.build` folder and modified it to bounce ECS service, rather than build images too. The previous way of working is in a new Jenkinsfile.

Added changelog - for now I've added 1.0.0 as the release of Thumbs.


.github/workflows/build_orchestrator.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  build-push:
+    env:
+      IMAGE_NAME: "orchestrator"
+      DOCKERFILE: "Dockerfile.Orchestrator"
+
```
stephenwf, line 12: I think an improvement in the future would be to have a single workflow for this, and pass in these environment variables. However that could be more general (at a github org level).

stephenwf (approved):

Issue #232, opened 2021-12-09
Composite Handler handles batching requests


donaldgray:
The composite-handler should handle requests containing a large number of items and split these into multiple downstream dlcs batches rather than having 1:1 composite-request:dlcs-batch. (e.g. a 1200 page PDF becomes 3 x 400 dlcs batches).

This change will likely break the current 301 + Location header response for completed jobs as there may be multiple dlcs-batches per single composite-request.

#230 has details of enforcing a max batch size to the API but this isn't likely to be implemented in the near future.

Issue #230, opened 2021-12-09
Set maximum batch size


donaldgray:
The API should have a configurable maximum size for batches. This can be set per dlcs deployment and may be altered as improvements are made for how batches are ingested.

DLCS is generally deployed to AWS behind an ELB, ELB has a default idle timeout of 60s, after which time it will return a 504. A payload of 1064 images to the API running on an m5.large took 70s. Increasing timeout allowed this request to go through successfully but we should benchmark to see a suitable limit and return a status code (413 - Payload Too Large or 400 - Bad Request) if the number of images is larger than maximuim.

donaldgray:
Closing - this has been implemented via the `MaxBatchSize` API config setting.

PR #237, opened 2022-01-13
Implements batching within the Composite Handler component

danielgrant:
This PR introduces "batching" within the Composite Handler component, which allows larger PDF's to be split into multiple DLCS ingestion requests. The purpose of this is to prevent excessively large ingestion requests (containing potentially thousands of images, or more) from exceeding request timeouts and / or clogging up the DLCS in general.

The batch size is a default of `100`, and can be tuned using the `DLCS_BATCH_SIZE` environment variable.

This PR addresses #230 and #232.


donaldgray (approved):
LGTM

PR #279, opened 2022-04-20
RFC for using oauth

donaldgray:
Initial write up of RFC for using DLCS as a client of an authorization server. 

The sample app written in conjunction with this was auth0 based but this should be usable for alternative providers.


docs/rfcs/008-more-access-control-oidc-oauth.md
```diff
@@ -0,0 +1,95 @@
+# Open Id Connect / OAuth
```
danielgrant, line 1: OpenID Connect / OAuth 2.0*

FWIW, this is an OIDC specific implementation since it depends explicitly on ID tokens (which exist as part of the OIDC specification only). Any references to OAuth 2.0 can probably be replaced with references to OIDC.

docs/rfcs/008-more-access-control-oidc-oauth.md
```diff
@@ -0,0 +1,95 @@
...
+#### Role Provider - OAuth2
+
+An alternative configuration block could be (this example uses values from the Wellcome [auth-test](https://github.com/wellcomecollection/iiif-builder/tree/master/src/AuthTest) application):
+
+```json
+{
+    "default": {
+        "config": "oauth2",
+        "provider": "auth0",
+        "domain": "<domain>.eu.auth0.com",
```
danielgrant, line 53: Being pedantic here, but I would suggest that documentation such as this shouldn't imply usage of a specific authorisation server such as Auth0.

docs/rfcs/008-more-access-control-oidc-oauth.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Role Provider - OAuth2
+
+An alternative configuration block could be (this example uses values from the Wellcome [auth-test](https://github.com/wellcomecollection/iiif-builder/tree/master/src/AuthTest) application):
+
+```json
+{
+    "default": {
+        "config": "oauth2",
+        "provider": "auth0",
```
danielgrant, line 52: What's the purpose of this `"provider"` key here? How does it affect the underlying implementation? I would have thought that providing standard configuration like domain, scopes, etc would be sufficient. Is this an implementation detail of the .NET framework that was used to prototype it?

danielgrant (changes requested):
I wonder if we can expand the design a little bit to incorporate the concept of API into it.

In Auth0 terms, you can read up on this here: https://auth0.com/docs/api. Essentially, you create a resource in Auth0 which represents an API. You can add scopes to this API, e.g. `weco:something`. Then, you can create roles in Auth0 and assign those scopes to a role, and assign the roles to users.

Now when the authorization code flow runs, you provide an "?audience=` parameter which targets this API. This gives you a couple of things:

1. The Auth0 user profile (returned as part of a JWT access token, ID token, or in response to `/userinfo`) will only contains the scopes that the user actually has assigned to them (via their Auth0 role assignments) and will remove any requested scopes which the user doesn't have access to.
2. By virtue of the `?audience=` parameter, you'll now receive an actual JWT access token from Auth0, instead of an opaque access token. This JWT contains the user profile embedded into it, but is also signed and thus allows your client to validate its authenticity by checking the signature against the public key used to sign it.

Let me know if you want a demonstration.

docs/rfcs/008-more-access-control-oidc-oauth.md
donaldgray: Agreed - removed in favour of `"<domain>"` only.

docs/rfcs/008-more-access-control-oidc-oauth.md
donaldgray: Updated usages.

docs/rfcs/008-more-access-control-oidc-oauth.md
donaldgray: domain, scopes etc should be sufficient - if we need to alter implementation based on `"provider"` then we have done something wrong. I thought the inclusion could be useful for logging purposes, e.g. if we are seeing slower responses with 1 type of provider over another

danielgrant (approved):

PR #304, opened 2022-05-23
Hydra Model updated

tomcrane:
This updates the develop branch with changes to the Hydra model that will be used by the updated API and Portal.

I have started a DLCS.Hydra.Tests project but it only contains some simple tests for now.

The DLCS.Mock project (adapted from the 2015 original) is a kind of test of this model, but it's not a runnable test like the others. It's not unit testing!

So we need to think how the Hydra base library and our DLCS.HydraModel are tested (I don't think we need separate test projects for these).

While there needs to be more Hydra testing I felt like I was making fairly trivial tests at the moment. 
I have removed some, but they can come back when testing API responses.


API.Client/HydraHttpResponseMessageX.cs
```diff
@@ -0,0 +1,51 @@
...
+using DLCS.Web.Response;
+using Hydra;
+using Hydra.Model;
+using Newtonsoft.Json;
+
+namespace API.Client
+{
+    public static class HydraHttpResponseMessageX
+    {
+        public static async Task<T?> ReadAsHydraResponseAsync<T>(this HttpResponseMessage response,
```
donaldgray, line 14: xml-comment here?

DLCS.Hydra.Tests/Model/ImageTests.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        var expected = BaseUrl + ImageApiPath;
+        var image = new Image(BaseUrl, 1, 1, "my-image");
+        var jsonString = JsonConvert.SerializeObject(image, serializerSettings);
+        var jObject = JObject.Parse(jsonString);
+        jObject["@id"].Value<string>().Should().Be(expected);
+    }
+    
+    [Fact]
+    public void ImageClass_SerialisesTo_shortStringForImage_Id()
```
donaldgray, line 44: Small thing `ImageClass_SerialisesTo_shortStringForImage_Id` -> `ImageClass_SerialisesTo_ShortStringForImage_Id`

DLCS.HydraModel/Customer.cs
```diff
@@ -135,9 +168,18 @@ public override void DefineOperations()
...
                 .GetStandardCollectionOperations(operationId + "space_", "Space", "vocab:Space");
 
+            
+            var images = GetHydraLinkProperty("allImages");
+            images.SupportedOperations = CommonOperations
+                .GetStandardCollectionOperations("_:customer_image_", "Image", "vocab:Image");
+            images.SupportedOperations.WithMethod("GET").Description =
+                "Can take query parameters";
+            images.SupportedOperations.WithMethod("POST").Description =
+                "Push an image for immediate processing, asynchronously. Might fail or timeout. This operation is rate-limited.";
```
donaldgray, line 181: `"This operation is rate-limited."` is it? Or should it be?

DLCS.HydraModel/DlcsResource.cs
```diff
@@ -68,8 +68,8 @@ public virtual JObject GetCollectionForm()
 
         public override string Context =>
             // do this better later - prefer not have FullName but makes reflection easier!
-            BaseUrl + "/contexts/" + GetType().FullName + ".jsonld";
+            BaseUrl + "/contexts/" + GetType().Name + ".jsonld";
```
donaldgray, line 71: Use string interpolation here for consistency?

Hydra/HydraStringX.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public static int? GetLastPathElementAsInt(this string? path)
+        {
+            var last = path.GetLastPathElement();
+            if (string.IsNullOrWhiteSpace(last))
+            {
+                return null;
+            }
+
+            // We want this to throw if not an int
```
donaldgray, line 35: Add a test for this, to confirm behaviour? Something like

```cs
var path = "foo/bar";
Action action = () => path.GetLastPathElementAsInt();
action.Should().Throw<InvalidCastException>()
```

donaldgray (approved):
Couple of comments but all minor

DLCS.HydraModel/DlcsResource.cs
tomcrane: Yeah a lot of Hydra code here from 2016, will fix up

Issue #245, opened 2022-02-28
Update YARP


donaldgray:
Reverse proxy logic is within the dotnet Orchestrator code thanks to [YARP](https://microsoft.github.io/reverse-proxy/).

Upgrade the libraries with latest version and investigate if there are improvements or new functionality to take advantage of.

donaldgray:
Look at `ReassignProxyRequest`, see https://github.com/microsoft/reverse-proxy/issues/1159#issuecomment-1119006548

PR #307, opened 2022-05-25
Update Yarp to 1.0.1

donaldgray:
#245 - update Yarp dependencies. 

This allowed us to take advantage of the new `IProxyStateLookup` class which gives access to the underlying Yarp managed cluster config. This is wrapped in a helper class `DownstreamDestinationSelector`.

`DownstreamDestinationSelector` handles finding target destination by leveraging the build-in `ILoadBalancingPolicy` handling, while still using the `ProxyDestination` enum. Kept `ProxyDestination` as it has destinations that we don't directly use Yarp to forward to (e.g. S3). Portions of `DownstreamDestinationSelector` have been copied from Yarp internals, these can be moved later if the APIs are exposed.

`DownstreamDestinationSelector` allowed us to use Yarp internals for health-checks, and meant we can get rid of `ReverseProxySettings` and managing these separately.

Slight change in routing logic to return a 502 if no available (healthy) destinations are available.

Refactored `RegionalisedObjectInBucket` to share extension methods with `ObjectInBucket` for generating http + s3 uris. This now returns `System.Uri` rather than a string. These now return S3 [Virtual-hosted-style](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#virtual-host-style-url-ex) uris rather than older path-style.

Added `AWS:S3:StorageBucket` appsetting key and method to `StorageKeyGenerator`, routing logic doesn't need to create bath and know about buckets etc now.


tomcrane:
> destinations that we don't directly use Yarp to forward to (e.g. S3)

Could we use YARP? or would there be no benefit?

donaldgray:
> > destinations that we don't directly use Yarp to forward to (e.g. S3)
> 
> Could we use YARP? or would there be no benefit?

What I've said isn't clear. We _do_ use Yarp DirectForwarding to stream AV from S3 if it requires Auth but the full S3 path will have been constructed separately https://github.com/dlcs/protagonist/blob/develop/Orchestrator/Features/TimeBased/TimeBasedRequestHandler.cs#L61-L62

What we don't do is have have any `ReverseProxy` Routes or Cluster config for S3 , so we'd never call `downstreamDesitnationSelector.TryGetCluster(ProxyDestination.S3, var out cluster);` as Yarp doesn't directly know about S3, even though we DirectForward to it via the `PathRewriteTransformer` which alters the entire path. 

`/iiif-av/2/1/my-video/full/full/max/max/0/default.mp4 -> YarpForwarded -> https://s3-eu-west-1.amazonaws.com/my-bucket/2/1/my-video/full/full/max/max/0/default.mp4`

tomcrane (approved):

PR #309, opened 2022-05-27
Start moving new API into develop

tomcrane:
tbc

### DIFF
--- a/API.Client/DlcsClient.cs
+++ b/API.Client/DlcsClient.cs
@@ -64,8 +64,12 @@ namespace API.Client
             var space = await response.ReadAsHydraResponseAsync<Space>(jsonSerializerSettings);
             return space;
         }
-        
-        
+
+        public async Task<HydraCollection<Image>> GetSpaceImages(int spaceId)
+        {
+            return await GetSpaceImages(1, 100, spaceId);
+        }
+
         public async Task<HydraCollection<Image>> GetSpaceImages(int page, int pageSize, int spaceId, string? orderBy = null)
         {
             var url = $"/customers/{currentUser.GetCustomerId()}/spaces/{spaceId}/images?page={page}&pageSize={pageSize}";
--- a/API.Client/DlcsClient.cs
+++ b/API.Client/DlcsClient.cs
@@ -47,7 +47,16 @@ namespace API.Client
                 ContractResolver = new CamelCasePropertyNamesContractResolver()
             };
         }
-        
+
+        public async Task<HydraCollection<Space>?> GetSpaces(int page, int pageSize, int? customerId = null)
+        {
+            customerId ??= currentUser.GetCustomerId();
+            var url = $"/customers/{customerId}/spaces?page={page}&pageSize={pageSize}";
+            var response = await httpClient.GetAsync(url);
+            var space = await response.ReadAsHydraResponseAsync<HydraCollection<Space>>(jsonSerializerSettings);
+            return space;
+        }
+
         public async Task<Space?> GetSpaceDetails(int spaceId)
         {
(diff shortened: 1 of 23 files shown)


tomcrane:
(NB once this is in, it can set the pattern for how the rest of the big PR comes across)

API.Tests/appsettings.Testing.json
```diff
@@ -0,0 +1,9 @@
+﻿{
+  "AWS": {
+    "Profile": "test-profile",
+    "Region": "eu-west-1"
+  },
+  "DLCS": {
```
donaldgray, line 6: Should this be "AWS" ?

Test.Helpers/Integration/ProtagonistAppFactoryX.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Test.Helpers.Integration.Infrastructure;
+
+namespace Test.Helpers.Integration;
+
+public static class ProtagonistAppFactoryX
+{
+    public static HttpClient ConfigureIntegrationTestClient<T>(
```
donaldgray, line 11: I don't know what it is but I think this should have an alternative name as it won't be usable by all services, e.g. Orchestrator won't use as that doesn't have Authentication.

Is the expectation that this'll be the default setup only and if any customisations are required then manually call `.WithConnectionString().WithTestServices()` etc? Thinking of cases where specific tests may need to call `.WithLocalStack()` or add other test services.

donaldgray (approved):
LGTM - added a few comments but not sure if some will be addressed in future PRs.

For Transactions I created #310.

> I verified the Deliverator behaviour by creating a new customer there and collecting all the changes that happened in the DB
I don't know how but we should try and keep track of mentions/links to Deliverator and at some point run through and remove them all as it's all in private repo.

API.Client/DlcsClient.cs
```diff
@@ -47,19 +47,36 @@ public class DlcsClient : IDlcsClient
...
             var space = await response.ReadAsHydraResponseAsync<Space>(jsonSerializerSettings);
             return space;
         }
-        
-        
+
         public async Task<HydraCollection<Image>> GetSpaceImages(int spaceId)
         {
-            var url = $"/customers/{currentUser.GetCustomerId()}/spaces/{spaceId}/images";
+            return await GetSpaceImages(1, 100, spaceId);
```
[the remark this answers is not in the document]
tomcrane: have made `pageSize` a settings with a default of 100, but `page` should still be explicitly `1` here - the first page, if no page specified.

API.Tests/Customers/CustomerValidationTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        var customer = GetValidNewCustomer();
+        customer.Name = "Admin";
+        
+        var errors = HydraCustomerValidator.GetNewHydraCustomerErrors(customer);
+
+        errors.Length.Should().BeGreaterThan(0);
+    }
+
+    [Fact]
```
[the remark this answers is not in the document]
tomcrane: I'll do that but remove some of these as they have been rolled into the big integration test.

API.Tests/appsettings.Testing.json
tomcrane: yes moved into AWS.S3

Test.Helpers/Integration/ProtagonistAppFactoryX.cs
tomcrane: I'll go with the clunky `ConfigureBasicAuthedIntegrationTestHttpClient` for now

PR #321, opened 2022-06-21
Add API for new Spaces

tomcrane:
Building on the previous customer PR, this uses the same patterns to API operations on Spaces.

In API, Controllers are v lightweight, just marshal Hydra <-> Entity model

Remove all use of DB and replace by API calls in Portal
API supports Deliverator operations on Space, and paging
GET spaces
GET space
POST to /spaces
PATCH space
(no DELETE though)

Return standardised HydraProblem on error

Tests include paging tests
The behaviour of Deliverator entity counters is reproduced with one exception - when creating a new space, it gets an entity counter. Deliverator sets this entity counter's `Next` to 0, which feels wrong, I've set it to 1.

DropZoneController in portal is partially refactored to eliminate a dependency on SpaceRepository (should only call HTTP API) but this will be completed in the next PR with deals with assets.

This PR does include GetSpaceImages, which can be tested manually against the spinup DB, but needs other Asset code in  the next PR before it gets an automated test.

Space.Tags and Space.Roles are now string[] and converted by the DBContext.
Portal "PageBehinds" no longer make Mediatr requests - they should JUST use DLCS API Client. If they can't - add more to API!


API/Features/Space/Requests/CreateSpace.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Threading.Tasks;
+using DLCS.Model.Customers;
+using DLCS.Model.Spaces;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace API.Features.Space.Requests
+{
+    /// <summary>
+    /// See Deliverator: API/Architecture/Request/API/Entities/CustomerSpaces.cs
```
donaldgray, line 11: This is more suited to `<remark>` than summary

API.Tests/Integration/SpaceTests.cs
```diff
@@ -0,0 +1,331 @@
...
+        
+        string spaceTestCustomerJson = $@"{{
+  ""@type"": ""Customer"",
+  ""name"": ""{customerName}"",
+  ""displayName"": ""Display - {customerName}""
+}}";
+        
+        var content = new StringContent(spaceTestCustomerJson, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsAdmin().PostAsync("/customers", content);
+        var apiCustomer = await response.ReadAsHydraResponseAsync<Customer>();
```
donaldgray, line 301: Is it necessary to create customer via API rather than DB? Feels like we're making these tests more brittle. A change in `POST /customers` behaviour (e.g. a new, required field) could end up breaking some of these `Space` tests; when in reality all of the `Space` endpoints are fine.

The benefit is that `POST /customers` could do something more than create a customer that is necessary for creating spaces but that 'more' behaviour should be asserted in customer tests.

API/Features/Space/Requests/GetPageOfSpaces.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace API.Features.Space.Requests
+{
+    /// <summary>
+    /// Request to get details of all spaces available for current user.
+    /// </summary>
+    public class GetPageOfSpaces : IRequest<PageOfSpaces>
+    {
+        public GetPageOfSpaces(int page, int pageSize, int? customerId = null, string? orderBy = null)
```
donaldgray, line 16: Should there be a test for `orderBy`?

API/Features/Space/Requests/GetPageOfSpaces.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+        
+        public async Task<PageOfSpaces> Handle(GetPageOfSpaces request, CancellationToken cancellationToken)
+        {
+            int? customerId = request.CustomerId ?? principal.GetCustomerId();
+            if (customerId == null)
+            {
+                throw new BadRequestException("No customer Id supplied");
+            }
+            var result = await spaceRepository.GetPageOfSpaces(customerId.Value, request.Page, request.PageSize, cancellationToken);
```
donaldgray, line 51: I'm only this far down the PR but it's so much easier to read than previous API!

API/Features/Space/SpaceController.cs
```diff
@@ -0,0 +1,130 @@
...
+                return HydraProblem("A space must have a name.", null, 400, "Invalid Space", null);
+            }
+            if (customerId <= 0)
+            {
+                return HydraProblem("Space must be created for an existing Customer.", null, 400, "Invalid Space", null);
+            }
+            
+            var command = new CreateSpace(customerId, space.Name)
+            {
+                // ImageBucket = space.ImageBucket, // not there
```
donaldgray, line 78: DeletE?

API/Features/Space/SpaceController.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+        
+        [HttpGet]
+        [Route("{spaceId}")]
+        public async Task<DLCS.HydraModel.Space> Index(int customerId, int spaceId)
+        {
+            var baseUrl = Request.GetBaseUrl();
+            var dbSpace = await mediator.Send(new GetSpace(customerId, spaceId));
+            return dbSpace.ToHydra(baseUrl);
```
donaldgray, line 104: Return a 404 if not found

API/Startup.cs
```diff
@@ -67,10 +71,12 @@ public void ConfigureServices(IServiceCollection services)
...
                 })
                 .AddLazyCache()
+                .AddScoped<ISpaceRepository, SpaceRepository>()
                 .AddSingleton<ICustomerRepository, DapperCustomerRepository>()
                 .AddScoped<IEntityCounterRepository, EntityCounterRepository>()
                 .AddSingleton<IAuthServicesRepository, DapperAuthServicesRepository>()
                 .AddScoped<ICustomerQueueRepository, CustomerQueueRepository>()
+                .AddScoped<IAssetRepository, DapperAssetRepository>()
                 .ConfigureMediatR()
                 .ConfigureSwagger()
```
donaldgray, line 81: Not specifically related to this PR but we should take note to update XML comments etc for better swagger docs.

DLCS.Model/Spaces/Space.cs
```diff
@@ -1,22 +1,25 @@
...
+    public class Space
     {
         public int Id { get; set; }
         public string Name { get; set; }
         public int Customer { get; set; }
         public DateTime Created { get; set; }
         public string ImageBucket { get; set; } = string.Empty;
-        public string Tags { get; set; } = string.Empty;
-        public string Roles { get; set; } = string.Empty;
+        public string[] Tags { get; set; } = {};
```
donaldgray, line 15: This is a lot cleaner than dealing with strings everywhere

DLCS.Repository/DlcsContext.cs
```diff
@@ -451,7 +451,12 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
                 entity.Property(e => e.Id).HasMaxLength(500);
 
-                entity.Property(e => e.Aliases).HasMaxLength(1000);
+                entity.Property(e => e.Aliases)
+                    .HasMaxLength(1000);
+                    // We have to roll this back because we have a Dapper mapping for this
```
donaldgray, line 456: Is it worth fixing dapper, or removing this comment? Trying to avoid 18months from now this comment still being there!

Looking at deliverator I'm not sure this field is ever used anywhere other than mapping back and forth to/from db/hydra

DLCS.Repository/Spaces/SpaceRepository.cs
```diff
@@ -30,31 +37,144 @@ public SpaceRepository(DlcsContext dlcsContext)
...
-            CancellationToken cancellationToken)
+            return newModelId;
+        }
+        
+
+        private async Task<Space?> GetSpaceInternal(int customerId, int spaceId, 
+            CancellationToken cancellationToken, string? name = null)
         {
-            throw new System.NotImplementedException();
+            Space space;
```
donaldgray, line 95: `Space? space`

DLCS.Repository/Spaces/SpaceRepository.cs
```diff
@@ -30,31 +37,144 @@ public SpaceRepository(DlcsContext dlcsContext)
...
+            if (maxUnauthorised != null)
+            {
+                dbSpace.MaxUnauthorised = (int)maxUnauthorised;
+            }
+
+            // ImageBucket?
+            
+            await dlcsContext.SaveChangesAsync(cancellationToken);
+
+            var retrievedSpace = await GetSpace(customerId, spaceId, cancellationToken);
```
donaldgray, line 176: Could that be a call to `GetSpaceInternal()`?

DLCS.Repository/Spaces/SpaceRepository.cs
```diff
@@ -30,31 +37,144 @@ public SpaceRepository(DlcsContext dlcsContext)
...
+                space = await dlcsContext.Spaces
+                    .Where(s => s.Customer == customerId)
+                    .SingleOrDefaultAsync(s => s.Name == name, cancellationToken: cancellationToken);
+            }
+            else
+            {
+                space = await dlcsContext.Spaces.AsNoTracking().SingleOrDefaultAsync(s =>
+                    s.Customer == customerId && s.Id == spaceId, cancellationToken: cancellationToken);
+            }
+            var counter = await dlcsContext.EntityCounters.AsNoTracking().SingleOrDefaultAsync(ec =>
```
donaldgray, line 107: Given it's an additional DB hit and not always required, is it worth having `bool withImageCount` param on this method?

Then, before getting `counter` you could have `if (space == null || !withImageCount) return space` to avoid extra DB hit.

donaldgray (approved):

PR #328, opened 2022-06-29
Basic Asset operations

tomcrane:
Add `orderBy` and `orderByDescending` as query string params to API and also to 2 portal pages (listing spaces, and images within a space).

Internally the code translates these to a single ordering string and a boolean `descending` - I did have a pass at them using the same mechanism but the former felt nicer for API params and the latter read easier in code.

Space and Asset queries take an additional LINQ step to add the necessary orderBy/orderByDescending clause.
We might need to think about the performance of this over very large counts of Images.

Added Integration Tests for assets, including

GET, GET collection, PUT, PATCH, PATCH multiple, POST ImageWithFile
Tests ordering and paging of assets for the collection GET

Some of these tests involve an immediate ingest (synchronous) call to Engine, which is mocked.
This engine call is made by MessageBus, which accepts the form of message used in the new Engine rewrite but translates it into the Deliverator engine format.

This does work in the new Portal sending to the old Engine - but we can strip out most of the MessageBus code once it runs against new Engine.

Modified ControllableMessageHttpHandler to work for multiple callbacks registered with unique keys - injected as a class fixture.

`Asset` now has nullable fields rather than default values. This is a big change, and appears to be OK. The benefit is not having a separate DTO with nullable fields to indicate UNSET fields.

This means it must be prepared for insertion into the DB, because it will fail if it makes it to AssetRepository with nullable fields still present. Business logic in the conversion from Hydra to Asset, and then in the Mediatr handlers for assets, ensure that 

Several separate Deliverator behaviours are represented by a single PutOrPatchImage Mediatr handler. This is also used multiple times for batch-patching. This has slightly broadened what's allowed to be updated in various calls - I think it could be broadened still further (e.g., PUT a change to ImageOptimisationPolicy to re-trigger an ingest, currently not allowed).
I'm disallowing batch-patch changes that would cause an Engine call - as this is synchronous, that's too much work for API to be doing in one call. (use the queue instead!)

`PutOrPatchImage` ONLY handles images atm. It won't handle AV or docs. I need to do that next.

`AssetConverter` class converts to and from Hydra form.

`ToDlcsModel` takes an API Hydra Image and does some initial data validation.

`(this Image hydraImage, int customerId, int? spaceId = null, string? modelId = null)`

Those optional params can be used as assertions (e.g., in an ActionResult with path params, I expect this image to be cust 5 space 23) or left for the converter to determine from whatever fields have been supplied. Also if the incoming Hydra Asset has an `@id` then it must match any cust/space/modelId information provided as fields, or as assertions.
https://github.com/dlcs/protagonist/pull/328/files#diff-1372b363df075fa111bc3c0d734c7542a03d202289b4c4a758f92fa7bab9bff6R89

Asset fields are only set if the Hydra fields are not null, so an Asset will come out of the converter with null fields.
The Mediatr handlers (but not the controller) will then use the services of `AssetPreparer` to ready the Asset for insert/update.
There is a lot of business logic on https://github.com/dlcs/protagonist/pull/328/files#diff-c8c41c55e7317f14f7b1fe5a07f8dcf092e255b3ddb34daeba131a39f0834addR48 
(see comments throughout)

Moved a few things to the base API HydraController to neaten up the actual controllers.

ImageController follows the now standard pattern of just marshalling Hydra to DLCS.Model and setting paging info required by views.

Added more `ProducesResponseType` attributes for Swagger etc but I think more are needed. 
I have been more specific in several places where Deliverator would just return HTTP 500.

The IngestBytes action method has been updated quite a bit - 
It calls a new Mediatr handler `HostAssetAtOrigin` - "stage this for me" - and then just passes to the standard PUT action method - because the asset is now a normal Hydra blob that states an Origin.

Implemented CustomerStorageRepository but very unlike the Deliverator code - re-did previous work as EF. The Dapper version is still there for reference - https://github.com/dlcs/protagonist/pull/328/files#diff-28d69d6f64155f27cb54152213a51c4754a642803a9b5ddef7de48f622089d59R13 

MessageBus has some overlap with the Engine work - e.g., https://github.com/dlcs/protagonist/blob/0fb19b431a3bf7b2e9db80b9616974e2ad08eaab/DLCS.Model/Messaging/IngestAssetRequest.cs#L6

### DIFF
--- a/DLCS.Model.Tests/Assets/ConverterTests.cs
+++ b/DLCS.Model.Tests/Assets/ConverterTests.cs
@@ -132,7 +132,7 @@ public class ConverterTests
         asset.Family.Should().Be(DLCS.Model.Assets.AssetFamily.Image);
         asset.Ingesting.Should().Be(true);
         asset.Origin.Should().Be(origin);
-        asset.InitialOrigin.Should().BeNull(); // not patchable
+        asset.InitialOrigin.Should().Be(initialOrigin); // not patchable but still carried on the Asset class.
         asset.NumberReference1.Should().Be(1);
         asset.NumberReference2.Should().Be(2);
         asset.NumberReference3.Should().Be(3);


API/Converters/AssetConverter.cs
```diff
@@ -0,0 +1,286 @@
...
+                asset.ThumbnailPolicy = thumbnailPolicy;
+            }
+            
+            var imageOptimisationPolicy = hydraImage.ImageOptimisationPolicy.GetLastPathElement("imageOptimisationPolicies/");
+            if (imageOptimisationPolicy != null)
+            {
+                asset.ImageOptimisationPolicy = imageOptimisationPolicy;
+            }
+            
+            // Not patchable? Does Engine set this?
```
donaldgray, line 273: IIRC in Deliverator `InitialOrigin` can only be set on the initial creation - it's passed to Engine, encoded in the SQS message. Engine will use it for ingestion if present. That's all it's used for, it's not persisted anywhere.

API/Features/Image/ImageController.cs
```diff
@@ -1,63 +1,342 @@
...
+        [FromBody] DLCS.HydraModel.Image hydraAsset)
+    {
+        // DELIVERATOR: https://github.com/digirati-co-uk/deliverator/blob/master/API/Architecture/Request/API/Entities/CustomerSpaceImage.cs#L74
+            
+        var assetId = new AssetId(customerId, spaceId, imageId);
+        var asset = hydraAsset.ToDlcsModel(customerId, spaceId, imageId);
+        asset.Id = assetId.ToString();
+    
+        // In the special case where we were passed ImageWithFile from the IngestBytes action, 
+        // it was a POST - but we should revisit that as the direct image ingest should be a PUT as well I think
```
donaldgray, line 154: Is it worth creating tickets for stuff like this to come back to? Easier to track when done/not done than a comment.

API/Features/Image/ImageController.cs
```diff
@@ -1,63 +1,342 @@
...
+                    
+                return HydraProblem(
+                    "All assets must have a ModelId", 
+                    null, 400, "Missing identifier", null);
+            }
+            foreach (var hydraImage in images.Members)
+            {
+                try
+                {
+                    // Here a Hydra object is being passed into the MediatR layer.
```
donaldgray, line 236: This comment doesn't look to be relevant anymore

API/Features/Image/ImageController.cs
```diff
@@ -1,63 +1,342 @@
...
+        if (images.Members is { Length: > 0 })
+        {
+            if (BulkPatchMayCauseReprocessing(images))
+            {
+                return HydraProblem(
+                    "Bulk patching operations may not contain origin or image policy information.", 
+                    null, 400, "Not Supported", null);
+            }
+            
+            // Is it OK if a patched image has a different space?
```
donaldgray, line 220: Is this a check for something? What does it refer to?

API/Features/Image/ImageController.cs
```diff
@@ -1,63 +1,342 @@
...
+            Id = Request.GetDisplayUrl() + "?patch_" + Guid.NewGuid()
+        };
+        return Ok(output);
+    }
+
+    private bool BulkPatchMayCauseReprocessing(HydraCollection<DLCS.HydraModel.Image> images)
+    {
+        if (images.Members == null) return false;
+        
+        // this should check the same things as AssetPreparer::PrepareAssetForUpsert
```
donaldgray, line 280: Could `PrepareAssetForUpsert` be used and the result checked to see if it requires reingest?

API/Features/Image/Requests/HostAssetAtOrigin.cs
```diff
@@ -0,0 +1,68 @@
...
+                var message = $"Failed to upload file to S3, aborting ingest. Key: '{objectInBucket}'";
+                logger.LogError(message);
+                return new HostAssetAtOriginResult { Error = message };
+            }
+
+            return new HostAssetAtOriginResult { Origin = objectInBucket.GetHttpUri().ToString() };
+        }
+        
+        private RegionalisedObjectInBucket GetObjectInBucket(AssetId assetId)
+            => new RegionalisedObjectInBucket(settings.AWS.S3.OriginBucket,
```
donaldgray, line 65: Pattern I've been using elsewhere for this is to use a method on `IStorageKeyGenerator`/`S3StorageKeyGenerator`; that way calling code doesn't worry about buckets + regions.

API/Features/Image/ImageController.cs
```diff
@@ -1,63 +1,342 @@
...
+                    var asset = hydraImage.ToDlcsModel(customerId, spaceId);
+                    var request = new PutOrPatchImage { Asset = asset, Method = "PATCH" };
+                    var result = await mediator.Send(request);
+                    if (result.Asset != null)
+                    {
+                        patchedAssets.Add(result.Asset);
+                    }
+                    else
+                    {
+                        logger.LogError($"We did not get an asset back for {asset.Id}");
```
donaldgray, line 247: `logger.LogError("We did not get an asset back for {AssetId}", asset.Id);`

API/Features/Image/Requests/PutOrPatchImage.cs
```diff
@@ -0,0 +1,378 @@
...
+using DLCS.Model.Messaging;
+using DLCS.Model.Spaces;
+using DLCS.Model.Storage;
+using MediatR;
+using Microsoft.Extensions.Options;
+
+namespace API.Features.Image.Requests
+{
+    /// <summary>
+    /// Unlike Deliverator we will handle PUTs and PATCHes with a single command, with
```
donaldgray, line 17: If you're unfamiliar with Deliverator this doesn't make sense - feels more like a remark, or comment than summary.

DLCS.Repository.Tests/Assets/AssetRepositoryTests.cs
```diff
@@ -0,0 +1,129 @@
...
+    public class AssetRepositoryTests
+    {
+        private readonly DlcsContext dbContext;
+        private readonly IAssetRepository sut;
+
+        public AssetRepositoryTests(DlcsDatabaseFixture dbFixture)
+        {
+            dbContext = dbFixture.DbContext;
+            // We want this turned on to match live behaviour
+            dbContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.TrackAll;
```
donaldgray, line 31: Might be best to new up a DbContext here and pass that to the sut rather than changing the one on the `DlcsDatabaseFixture`. 
`DlcsDatabaseFixture` is a collection fixture so it'll be shared across all classes in this same assembly. Settings `TrackAll` will set it for all other test classes too, which might lead to unexpected behaviour.

DLCS.Repository.Tests/DLCS.Repository.Tests.csproj
```diff
@@ -28,6 +28,7 @@
     </ItemGroup>
 
     <ItemGroup>
+      <ProjectReference Include="..\API\API.csproj" />
```
donaldgray, line 31: Doesn't seem right?

DLCS.Repository/Assets/AssetQueryX.cs
```diff
@@ -0,0 +1,67 @@
...
+                "String3" => "Reference3",
+                _ => pascalCase
+            };
+        }
+
+
+        // Create an Expression from the PropertyName. 
+        // I think Split(".") handles nested properties maybe - seems unnecessary but from an SO post
+        // "x" means nothing when creating the Parameter, it's just used for debug messages
+        private static LambdaExpression CreateExpression(Type type, string propertyName)
```
donaldgray, line 54: 👍

DLCS.Repository/Messaging/MessageBus.cs
```diff
@@ -0,0 +1,277 @@
...
+using DLCS.Core.Settings;
+using DLCS.Model.Assets;
+using DLCS.Model.Messaging;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace DLCS.Repository.Messaging
+{
+    public class MessageBus : IMessageBus
```
donaldgray, line 17: Reading the name of this and seeing impl feels at odds. I'd expect `MessageBus` to be putting stuff onto a queue rather than calling direct. 

Would this be better in the API project? Would anywhere else be likely to use it?

And/or would `EngineMessage<Something>` be a better name?

Orchestrator/Assets/MemoryAssetTracker.cs
```diff
@@ -159,9 +159,9 @@ T SetDefaults<T>(T orchestrationAsset)
                     return SetDefaults(new OrchestrationImage
                     {
                         S3Location = getImageLocation.Result?.S3, // TODO - error handling
-                        Width = asset.Width,
-                        Height = asset.Height,
-                        MaxUnauthorised = asset.MaxUnauthorised,
+                        Width = asset.Width ?? 0,
+                        Height = asset.Height ?? 0,
+                        MaxUnauthorised = asset.MaxUnauthorised ?? 0,
```
donaldgray, line 164: `-1`?

donaldgray (approved):
Once this section of the API has been done we should run through all of the comments and clean them up to avoid the _has this been done?_ situation when we come back to it later

API.Tests/Integration/AssetTests.cs
```diff
@@ -0,0 +1,657 @@
...
+        var id = "99/2999/asset1";
+        await dbContext.Images.AddTestAsset(id, space:2999);
+        await dbContext.SaveChangesAsync();
+        var getUrl = "/customers/99/spaces/2999/images";
+        var response = await httpClient.AsCustomer(99).GetAsync(getUrl);
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+    }
+    
+    [Fact]
+    public async Task Get_Assets_In_Space_Returns_Page_of_Assets()
```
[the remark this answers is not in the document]
tomcrane: Removed

API.Tests/Integration/AssetTests.cs
```diff
@@ -0,0 +1,657 @@
...
+            else
+            {
+                nextColl.Members.Should().HaveCount(5);
+                view.Next.Should().BeNull();
+            }
+        }
+    }
+    
+    [Fact]
+    public async Task Paged_Assets_Support_Ordering()
```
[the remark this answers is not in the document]
tomcrane: I've added a variation of this, but I couldn't do it without creating new assets for each run.

API/Converters/AssetConverter.cs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                throw new APIException($"Asset {dbAsset.Id} does not start with expected prefix {prefix}");
+            }
+
+            var modelId = dbAsset.Id.Substring(prefix.Length);
+            
+            var image = new Image(urlRoots.BaseUrl, dbAsset.Customer, dbAsset.Space, modelId)
+            {
+                InfoJson = $"{urlRoots.ResourceRoot}iiif-img/{dbAsset.Id}",
+                ThumbnailInfoJson = $"{urlRoots.ResourceRoot}thumbs/{dbAsset.Id}",
```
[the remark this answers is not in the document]
tomcrane: They come from the config but could come in a more complex way

API/Converters/AssetConverter.cs
tomcrane: Now also sent to engine (with test)

API/Features/Image/ImageController.cs
tomcrane: It could but we'd have to have acquired the existing Asset (rather than Hydra version).
I was conscious that there's a little bit of duplication here but it's specific to the batch patch, a quick check on the hydra objects before the next layer is invoked.

DLCS.Repository.Tests/Assets/AssetRepositoryTests.cs
tomcrane: All works nicely... but my test `AssetRepository_Throws_if_Same_Asset_Tracked` now fails, because it no longer throws an exception! It's testing that you're not using the same dbcontext to make changes to tracked assets - but now the tests are using a different dbcontext to the API.

Orchestrator/Assets/MemoryAssetTracker.cs
tomcrane: Yeah I'm not sure what the default should be - the asset from the DB will have a value so something's gone wrong if it doesn't which maybe suggests it should be restricted in some way? Not sure.

donaldgray: Yeah true, if in doubt lock it down. `0` makes sense in that case.

API.Tests/Integration/AssetTests.cs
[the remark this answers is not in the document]
donaldgray: One option is configure known items in ctor, these will then be in the DB by the time the test is run (`NamedQueryRepositoryTests` uses this approach)

DLCS.Repository/Messaging/MessageBus.cs
tomcrane: At the moment the only implemented thing on this class is the direct engine call, which is not the standard sort of message - but won't it also later be enqueueing  incoming assets from API calls to the queue, broadcasting asset change events etc?

PR #345, opened 2022-07-22
Feature/api portalusers

tomcrane:
Add CRUD operations for portal users.

### DIFF
--- a/API/Features/Customer/Requests/DeletePortalUser.cs
+++ b/API/Features/Customer/Requests/DeletePortalUser.cs
@@ -1,0 +1,63 @@
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using DLCS.Core.Collections;
+using DLCS.Repository;
+using MediatR;
+
+namespace API.Features.Customer.Requests;
+
+public class DeletePortalUser : IRequest<DeletePortalUserResult>
+{
+    public DeletePortalUser(int customerId, string userId)
+    {
+        CustomerId = customerId;
+        UserId = userId;
+    }
+
+    public int CustomerId { get; }
+    public string UserId { get; }
+    
+}
+
+
+public class DeletePortalUserResult
+{
+    public string Error { get; set; }
+}
+
+public class DeletePortalUserHandler : IRequestHandler<DeletePortalUser, DeletePortalUserResult>
+{
+    private readonly DlcsContext dbContext;
+
+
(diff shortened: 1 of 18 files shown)


API/Features/Customer/Requests/DeletePortalUser.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        var dbUser = await dbContext.Users.FindAsync(new object?[]{request.UserId}, cancellationToken);
+        if (dbUser == null)
+        {
+            return new DeletePortalUserResult { Error = "No such user" };
+        }
+
+        if (dbUser.Customer != request.CustomerId)
+        {
+            return new DeletePortalUserResult { Error = "User doesn't belong to customer" };
```
donaldgray, line 51: Is it risky to return this information as it highlights which emails are in the system? Maybe log a message but return something less detailed?

API/Features/Customer/Requests/GetPortalUsers.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public GetPortalUsersHandler(DlcsContext dbContext)
+    {
+        this.dbContext = dbContext;
+    }
+
+    public async Task<IList<User>> Handle(GetPortalUsers request, CancellationToken cancellationToken)
+    {
+        var users = await dbContext.Users.Where(u => u.Customer == request.CustomerId)
+            .ToListAsync(cancellationToken);
```
donaldgray, line 29: Use `AsNoTracking()`.

Another thought, for later - handlers like this would be a good candidate for the new `IStreamRequest<T>` handler in [Mediatr 10](https://github.com/jbogard/MediatR/releases/tag/v10.0.0), which supports returning `IAsyncEnumerable<T>`. Would mean using `.AsAsyncEnumerable()` rather `.ToListAsync()`.

DLCS.Core/Strings/StringX.cs
```diff
@@ -70,5 +72,50 @@ public static string ToConcatenated(this string str, char separator, params stri
 
             return sb.ToString();
         }
+        
+        
+        public static bool IsValidEmail(this string email)
```
donaldgray, line 77: Worth linking back to the docs.microsoft.com page? (e.g. in a `<remark>`)

donaldgray (approved):
LGTM - few comments but no show stoppers.

Issue #250, opened 2022-02-28
Restructure repository layout


donaldgray:
With the inclusion of the composite-handler (#222 ) we have a mixture of dotnet and python in the main repo.

Currently all dotnet projects are in the root of the repository, move this down to `src/` folder that includes subfolder for composite-handler and the dotnet code.

This will involve changes to gh-actions etc to accomodate changes.

donaldgray:
Take this opportunity to apply solution wide refactorings, e.g. using file-scoped namespaces from dotnet 6

PR #349, opened 2022-07-26
Restructure Repository

donaldgray:
Move the following and update github action paths to match:
* `composite_handler` -> `src/composite-handler`
* All dotnet projects from root to `src/protagonist`

Use `dotnet format` to use file-scoped namespaces

See #250


garyttierney (approved):
LGTM

PR #358, opened 2022-08-09
Do not add Services to info.json if no roles or auth-services

donaldgray:
Fix issue where a restricted image with no roles would result in a `null` service being added to the info.json. This would be served fine on initial request but would fail subsequent requests as it can't be deserialized from S3.

Example abberviated info.json as written to S3:

```
{
  "@context": "http://iiif.io/api/image/3/context.json",
  "id": "1/1/image",
  "type": "ImageService3",
  "profile": "level2",
  "protocol": "http://iiif.io/api/image",
  "width": 1724,
  "height": 1719,
  "sizes": [ ... ],
  "tiles": [ ... ],
  "extraQualities": [ ... ],
  "service": [ null ],
  "extraFormats": [ ... ],
  "extraFeatures": [ .. .]
}
```

Fix is to not add a `"service"` if there are no roles or if no authService found.

Bug spotted when testing #357, introduced as part of change in auth logic (#334) as not all images that require auth will have a role.

### DIFF
--- a/src/protagonist/Orchestrator.Tests/Integration/ImageHandlingTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Integration/ImageHandlingTests.cs
@@ -15,6 +15,7 @@ using FluentAssertions;
 using IIIF;
 using IIIF.ImageApi.V2;
 using IIIF.ImageApi.V3;
+using IIIF.Serialisation;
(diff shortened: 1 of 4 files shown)


donaldgray:
Switched base-branch and not picked up by GH action. will close and recreate

PR #360, opened 2022-08-09
standardise naming in API controllers

tomcrane:
Generally:

Standardising comments on Action methods, so they include the HTTP verb+path in the summary, e.g.,
GET /customers/{customerId}

Standardising action method names to _VerbEntity_ or _VerbEntities_
GetApiKey, PatchImages etc
One consequence of this is that the Action method name is then usually the same as the Mediatr Request class name. is that a problem?

Removing some null check warnings and unnecessary passing of nulls to the HydraProblem response.

### DIFF
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -50,12 +50,12 @@ public class ImageController : HydraController
     /// <summary>
     /// GET /customers/{customerId}/spaces/{spaceId}/images/{imageId}
     /// 
-    /// A single Hydra GetImage.
+    /// A single Hydra Image.
     /// </summary>
     /// <param name="customerId">(from resource path)</param>
     /// <param name="spaceId">(from resource path)</param>
     /// <param name="imageId">(from resource path)</param>
-    /// <returns>A Hydra JSON-LD GetImage object representing the Asset.</returns>
+    /// <returns>A Hydra JSON-LD Image object representing the Asset.</returns>
     [HttpGet]
     [ProducesResponseType(200, Type = typeof(DLCS.HydraModel.Image))]
     [ProducesResponseType(404, Type = typeof(Error))]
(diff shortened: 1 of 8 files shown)


src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -115,19 +117,19 @@ public async Task<IActionResult> Image(int customerId, int spaceId, string image
...
+    /// (Unlike Deliverator, the same method handles a PATCH.)
     /// </summary>
     /// <param name="customerId">(from resource path)</param>
     /// <param name="spaceId">(from resource path)</param>
     /// <param name="imageId">(from resource path)</param>
     /// <param name="hydraAsset">The body of the request contains the Asset in Hydra JSON-LD form (Image class)</param>
     /// <returns>The created or updated Hydra Image object for the Asset</returns>
-    [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))] // for Patch
+    [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))] // for PATCH
     [ProducesResponseType((int)HttpStatusCode.Created, Type = typeof(DLCS.HydraModel.Image))] // for PUT when created
```
donaldgray, line 133:
```cs
[ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))] // for PATCH
[ProducesResponseType((int)HttpStatusCode.Created, Type = typeof(DLCS.HydraModel.Image))] // for PUT when created
```

Given that these are for specific verb types, is it worth having 2 separate action methods that internally use the same method? Keeps the fact that `PUT` + `PATCH` being handled by same logic as an implementation detail. Would also make the swagger/openapi generated docs clearer.

```cs
/// <summary>
/// PUT /customers/{customerId}/spaces/{spaceId}/images/{imageId}
///
/// PUT an asset to its ID location
/// </summary>
[ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))]
[ProducesResponseType((int)HttpStatusCode.Created, Type = typeof(DLCS.HydraModel.Image))]
[HttpPut]
public Task<IActionResult> PutImage(..) => PutOrPatchImage(..);

/// <summary>
/// PATCH /customers/{customerId}/spaces/{spaceId}/images/{imageId}
///
/// PATCH asset at that location.
/// </summary>
[ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))]
[HttpPatch]
public Task<IActionResult> PatchImage(..) => PutOrPatchImage(..);

private Task<IActionResult> PutOrPatchImage(..) { ... }
```

donaldgray (approved):
LGTM 

> One consequence of this is that the Action method name is then usually the same as the Mediatr Request class name. is that a problem?

Looking at existing mediatr handlers it looks fine. `Get<Entity>` + `Delete<Entity>` make sense, and we have have `Create<Entity>` rather than `Post<Entity>`.

PR #362, opened 2022-08-10
Proposed shared storage use tracking solution

garyttierney:


docs/adr/0003-storage-use-tracking.md
```diff
@@ -0,0 +1,101 @@
...
+flags 
+t=target_FID 
+ef=extended_flags
+u=uid:gid
+nid=client_NID
+p=parent_FID 
+target_name
+```
+
+Caveat: `target_name` refers to the _basename_ of the file being accessed, retreiving the full path requires a lookup on `parent_FID`.
```
donaldgray, line 76: What is the basename? Is this, e.g. if I had `/path/to/image.ext` would basename be `image.ext`?

donaldgray (commented):
LGTM, something that occurred while reading:

Would file system location/directory depth have any bearing on performance at all? Is it better to have fewer, larger dirs, or more smaller/deeper dirs - or does it not matter? 

The reason for mentioning is current orchestrator process (for deliverator) uses the asset identifier to derive save location on disk. It chunks first X characters into 2-character strings. This was to allow an even spread of files across different folders and made sense when all image id's were guids so distribution would be even but that's not always the case. This has been migrated to this orchestrator 

e.g. (for illustrative purposes - the number of 2-char dirs may differ, I can't quite recall)
* an asset with this Id `10/20/this-is-an-example` (asset Ids are in the form `/{customer}/{space}/{identifier}` )
* would be saved to `/mnt/lustre/10/20/th/is/-i/s-/this-is-an-example`

garyttierney:
> Would file system location/directory depth have any bearing on performance at all? Is it better to have fewer, larger dirs, or more smaller/deeper dirs - or does it not matter?

If going with the recursive file walk that could have a fairly significant impact on how long it takes to walk a file tree. The overhead of context switching to traverse single directory files can quickly add up. 

This task aside, the Lustre MDS will have to do additional bookkeeping and create additional `inode`s for every directory and each `inode` occupies [at least 2.56KiB on each OST](https://wiki.lustre.org/Lustre_Tuning#Planning_for_Inodes). Here are some example figures from the tuning wiki page:

> A 2TB MDS by default will have 400M inodes. A 10TB MDT would hold 2B inodes. With an MDS inode ratio of 1536 bytes per inode, a 3TB MDS would hold 2B inodes, and a 6TB MDS would hold 4B inodes, which is the maximum number of inodes currently supported by ldiskfs.

So if you allocate ~5 `inodes` for every asset then you're effectively reducing the number of assets you can store by a factor of 5. 

Given those numbers above (although the size of the MDS is not the size of the cluster, it's just metadata) it's not going to have any significant impact on how many assets we can keep in scratch space, but it does add unnecessary overhead to filesystem operations that could be avoided (e.g., if you need to delete an asset from scratch you need to recursively remove the entire subtree, and similarly for filesystem walks).

However, on the other side of the problem: large directories make fast file enumeration unresponsive. This isn't a problem for building the initial view of the cache since it doesn't need to be responsive and it isn't a problem for asset lookup since the full path is deterministic as shown above, but is a consideration if there is any code relying on directory listings of the cache.

If there is no scenario where you need to answer a query such as "what assets are cached for customer 10 and space 20" without consulting the scavenger service then I'd aim to keep file count low but keep some organization with something like /customer/space/asset-identifier/data.

docs/adr/0003-storage-use-tracking.md
garyttierney: That's right. In bash you'd get the full path via 

```shell
> $ "$(lfs fid2path $parent_FID)/$target_name"
```

tomcrane:
The file system approach in Deliverator is encountered elsewhere in digital repositories in cultural heritage (usually called a "PairTree" - see https://ocfl.io/1.0/implementation-notes/#storage-root-hierarchy). It made perfect sense for Deliverator for two reasons:

 - the source filenames were GUIDs with an assumed even distribution of pairs
 - the tree-walking approach suited having directories like that

Neither of these are true. In fact in the Wellcome case a pairtree approach is massively skewed because of file naming conventions.

> If there is no scenario where you need to answer a query such as "what assets are cached for customer 10 and space 20" without consulting the scavenger service

I don't think there is any code that would require a directory listing?

It sounds like we should just use the same form on Lustre as the S3 keys and URIs of assets; /customer/space/asset - even if that means Wellcome has (potentially) 50m files in /2/2/ (if they wanted to increase the Lustre volume size, say).

This is also simple and easy for some other service to do something (test presence of, read) with a lustre file, without any additional knowledge - it's going to be where you think it is, if it's there.

docs/adr/0003-storage-use-tracking.md
```diff
@@ -0,0 +1,101 @@
...
+### Record filesystem changes with `inotify`
+
+`inotify` is a Linux kernel mechanism for recording filesystem events.
+It is only capable of recording events for local VFS devices (i.e. you will only get events for the machine you're using `inotify` on) and is inapplicable here.
+
+## Additional Consideration
+
+### Priority Assets
+
+It may be desirable to provide a set of assets that should never or rarely be evicted, regardless of time or frequency of access.
```
tomcrane, line 100: Could we arrange the file system as

/a/<files>
and 
/b/<files>

where the process only evicts from /a/ but orchestrator can orchestrate to both, deciding which to orchestrate to based on additional metadata?

Or, we just have another lustre volume that our scavenging process does not run on. Both volumes are mounted by the image servers and when orchestrator proxies it sends the location to the image server as now.

PR #363, opened 2022-08-11
Support `q` and `string1`, `number1` etc params on paged Asset API calls

tomcrane:
The Hydra classes gain an `ImageQuery` object which as serialised JSON can be used as the `q` parameter to an asset query.

`/customers/99/spaces/377/images?pageSize=50&q={"number3": 2}`

(NB there is a class like this in DDS which can ultimately be replaced with a ref to this one).

This is called ImageQuery rather than AssetQuery to reflect the rest of the current Hydra API.

However it is parsed to an `AssetFilter` object in the controller. While this looks similar it means that we don't pass the hydra-related on any deeper than the controller.

This AssetFilter can also be obtained directly by parsing out string1=xxx etc from the query string. I did not add these to the action method signature as I felt it would get very messy; there are helpers to examine the request querystring.

If an ImageQuery object is present AND direct params are present, the ImageQuery is parsed first then any direct params added to the AssetFilter - or override the same field on the AssetFilter if already present.

Multiple fields are ANDs, so `/customers/99/spaces/377/images?pageSize=50&q={"number3": 2, "string2": "1-10"}` requires both to match.

### DIFF
--- a/src/protagonist/API.Tests/Integration/AssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/AssetTests.cs
@@ -683,4 +683,67 @@ public class AssetTests :
         asset.Origin.Should().Be("https://protagonist-test-origin.s3.eu-west-1.amazonaws.com/99/1/Post_ImageBytes_Ingests_New_Image");
 
     }
+
+    [Fact]
+    public async void Paged_Assets_Can_Be_Queried()
+    {
+        // arrange
+        await dbContext.Spaces.AddTestSpace(99, 377, "query-tests");
+        for (int i = 1; i <= 20; i++)
+        {
+            var padded = i.ToString().PadLeft(4, '0');
+            await dbContext.Images.AddTestAsset($"99/377/asset-{padded}",
+                customer: 99, space: 377,
+                num1: i, num2: i % 2, num3: i % 3,
+                ref1: $"Asset {padded}",
+                ref2: i < 11 ? "1-10" : "11-20",
+                ref3: i < 16 ? "1-15" : "16-20");
+        }
+        await dbContext.SaveChangesAsync();
+        
+        // Act
+        var assetPage = "/customers/99/spaces/377/images?pageSize=50&q={\"string3\": \"16-20\"}";
+        var response = await httpClient.AsCustomer(99).GetAsync(assetPage);
+
+        // Assert
+        var coll = await response.ReadAsHydraResponseAsync<HydraCollection<JObject>>();
+        coll.Members.Length.Should().Be(5);
+        
+        
+        // Act
+        assetPage = "/customers/99/spaces/377/images?pageSize=50&q={\"string2\": \"1-10\"}";
+        response = await httpClient.AsCustomer(99).GetAsync(assetPage);
+
+        // Assert
+        coll = await response.ReadAsHydraResponseAsync<HydraCollection<JObject>>();
+        coll.Members.Length.Should().Be(10);
+        
+        
+        // Act
+        assetPage = "/customers/99/spaces/377/images?pageSize=50&q={\"number3\": 2}";
+        response = await httpClient.AsCustomer(99).GetAsync(assetPage);
+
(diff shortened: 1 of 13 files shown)


src/protagonist/DLCS.Repository/Assets/AssetQueryX.cs
```diff
@@ -63,4 +64,48 @@ private static LambdaExpression CreateExpression(Type type, string propertyName)
...
+    public static IQueryable<Asset> ApplyAssetFilter(this IQueryable<Asset> queryable, 
+        AssetFilter? assetFilter, bool filterOnSpace = false)
+    {
+        if (assetFilter == null)
+        {
+            return queryable;
+        }
+
+        var filtered = queryable;
+        // is HasText right here? Would we want to select string1 = ""?
```
donaldgray, line 77: `HasText` makes sense, although I can see the value of explicitly wanting empty string. Could update later if it's needed.

src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -83,19 +83,33 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
...
+        if (q.HasText())
+        {
+            assetFilter = Request.GetAssetFilterFromQParam(q);
+            if (assetFilter == null)
+            {
+                return HydraProblem("Could not parse query", null, 400);
+            }
+        }
+        // Now augment this assetFilter, or create one if we don't have one yet
+        assetFilter = Request.UpdateAssetFilterFromQueryStringParams(assetFilter);
```
donaldgray, line 109: would lines 99-109 be cleaner as a method call as there's already a few things going on in that method. making it easier when scanning. `AssetFilter? assetFilter = GetAssetFilter();` (personal opinion!)

src/protagonist/API.Tests/Integration/AssetTests.cs
```diff
@@ -683,4 +683,67 @@ public async Task Post_ImageBytes_Ingests_New_Image()
         asset.Origin.Should().Be("https://protagonist-test-origin.s3.eu-west-1.amazonaws.com/99/1/Post_ImageBytes_Ingests_New_Image");
 
     }
+
+    [Fact]
```
donaldgray, line 687: Is this better as?

```cs
[Theory]
[InlineData("/customers/99/spaces/377/images?pageSize=50&q={\"string3\": \"16-20\"}", 5)]
[InlineData("/customers/99/spaces/377/images?pageSize=50&q={\"string2\": \"1-10\"}", 10)]
...
```

Having them as separate cases would make it easier to see at a glance if, e.g. 2 of the 5 fail. As is the whole case will fail if any of them fail - so could be 1 or all of them.

Could add the images in ctor, or append something to the Id's to make them unique per inline data.

src/protagonist/DLCS.Repository/Assets/AssetQueryX.cs
```diff
@@ -63,4 +64,48 @@ private static LambdaExpression CreateExpression(Type type, string propertyName)
 
         return Expression.Lambda(body, param);
     }
+
+    public static IQueryable<Asset> ApplyAssetFilter(this IQueryable<Asset> queryable, 
+        AssetFilter? assetFilter, bool filterOnSpace = false)
```
donaldgray, line 69: `bool filterOnSpace = false` is always `false`, is it worth removing?

donaldgray (approved):

src/protagonist/DLCS.Repository/Assets/AssetQueryX.cs
tomcrane: It won't be in the two other calls that also take a query

src/protagonist/API/Features/Image/ImageController.cs
tomcrane: Have rewritten to improve

PR #364, opened 2022-08-11
Changes after orchestrator review

donaldgray:
A couple of different changes based on feedback from review.

Main change is to _not_ set orchestration status until image needs to be orchestrated. This mostly involved changes to `MemoryAssetTracker`, adding a `bool requireOrchestrationStatus` flag to some methods. Also updated `ImageOrchestrator` to refresh the asset if status is `Unknown`.

Refactored the `AsyncKeyedLock` class to extract an interface. ~~Previous implementation held a lock per instance, extracted interfaces to allow easier refactoring. Introduced `OrchestrationLock` which is a decorator around `AsyncKeyedLock` this is registered as a singleton and allows the `FileBasedStatusProvider` class to easily share a locker with the `ImageOrchestrator` and allow it to check if status is "Orchestrating".~~. The initial work to alter how locks are handled led to a degradation in performance so I have removed it but left the slight refactoring of `AsyncKeyedLock` to use interfaces as it could lead to alternative implementations in the future.

Added some controller extension methods to tidy common tasks like error handling, cache-header setting etc.

Final change is removal of `DefaultHttpTransformer` - this was a remnant of when redirecting some queries directly to deliverator.


donaldgray:
Closing this for now - taking a different approach using LazyCache as discussed with @tomcrane

PR #375, opened 2022-08-23
Add ADR proposal with dependabot processes

garyttierney:


donaldgray (approved):
LGTM - I think both approaches have their merits but I personally prefer _Automatically merge all incoming dependency changes to a dependency staging branch for manual review_ as I think fewer/larger "update dependency" commits are better than lots of small churning commits to update patch versions.
It's not a strongly held opinion so happy to go with other approach if @tomcrane and @garyttierney  prefer.

docs/adr/0004-dependabot-upgrade-process.md
```diff
@@ -0,0 +1,69 @@
...
+#### Negative Consequences
+
+* Requires adherence to SemVer at a basic level for a majority of upstream dependencies. This varies between ecosystems.
+
+### Automatically merge all incoming dependency changes to a dependency staging branch for manual review
+
+Funnels all incoming dependabot changes into a single `dependencies` staging branch.
+This branch will be reviewed by developers on a given schedule (e.g. weekly basis) and the changes from that period of time will be merged to `develop`
+
+[Example from @stephenwf's capture model work](https://github.com/digirati-co-uk/capture-models/blob/master/.github/workflows/update-dependencies.yml)
```
stephenwf, line 57: A pain point worth mentioning here is the need to rebase:

https://github.com/digirati-co-uk/capture-models/blob/master/.github/workflows/sync-dependencies.yml

And one extra positive is that you can choose which version to auto-merge into that staging branch

https://github.com/digirati-co-uk/capture-models/blob/master/.github/workflows/automerge-dependencies.yml

donaldgray:
@garyttierney - can this be merged?

PR #380, opened 2022-09-01
Handle PUT and PATCH for timebased assets

donaldgray:
General changes were to handle PUT + PATCH for Timebased assets, raising a SQS notification rather than directly calling engine. Main points:

* Split `AssetTests` out to `GetAssetTests` and `ModifyAssetTests`. The latter was simplified by using a mock `IEngineClient` rather than a controllable http message handler. There are no new tests in the former.
* Move `AssetPreparer` from DLCS.Model -> API project. I can't see this being used elsewhere but if it is we can move back.
* Renamed `PutOrPatchImage` request to `CreateOrUpdateImage`, felt wrong to tie it to specific Http verbs. Also, in here:
  * If `T` asset, create a batch and queue message.
* Added the following to DLCS.AWS:
  * `IQueueLookup` + `SqsQueueLookup` - takes `AssetFamily` and whether priority and returns the queue name. This is similar to `IStorageKeyGenerator` for S3 - keeps all references to specific queues within AWS project.
  * `IQueueSender` + `SqsQueueSender` - takes queue name from above, looks up queue URL and sends message, returning success/failure boolean.
* Added `DlcsSettings.UseLegacyEngineMessage` - this controls whether the deliverator or protagonist style message is sent to Engine (works for sync + async).
* Added `BatchRepository` - this creates a batch and saves it to DB to generate a new pk. Once it has the PK it sets the `Asset.Batch` value for any asset in that batch. Used Dapper for saving Batch as there may be other outstanding changes in the dbContext that we may not want to commit at that time.
* Slight refactoring of `DapperRepository` to allow it to use either `IConfiguration` or `DlcsContext` to get connection string.
* Small tweaks like lowering some logging statements and reordering args so that `CancellationToken` is always last.
* Added `IEngineClient` - this has 2 methods for sending sync or async. It also has logic for checking `UseLegacyEngineMessage`, having it in single place means that we only need to change `EngineClient` code when retiring legacy message format handling.
* Refactored `AssetNotificationSender`:
  * Takes an instance of `IEngineClient` rather than `HttpClient`. I _think_ we may have run into some lifecycle issues with the latter as `AssetNotificationSender` was registered as a singleton rather than via `AddHttpClient<IAuthServicesRepository, DapperAuthServicesRepository>` so I'm not sure how the httpClient would be disposed/reused. Another option would have been to pass an `IHttpClientFactory` but having `IEngineClient` removes the requirement for `AssetNotificationSender` to need `DlcsSettings` and allows all engine specific stuff like transforming to legacy format in one place.
  * Takes an `Asset` rather than an `IngestAssetRequest`, it converts to latter before passing down to actual handler. This was for thinking ahead when `AssetNotificationSender` may be used for raising Delete/CacheInvalidation events - if it takes an `Asset` only then the caller doesn't need to worry about what the actual sent message format is (e.g. `DeleteAssetNotification`).
  * Uses `IEngineClient` for raising both sync + async. This keeps the implementation thin, it can convert to the appropriate format and call other services for raising notificatinos, e.g. `ISnsClient` or `ISqsClient` etc.
  * Moved legacy json methods out to separate class to make it easier to remove eventually.
  * Deleted `AssetNotificationTests` as these are now covered by `EngineClientTests`.


src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -93,219 +63,180 @@ public class AssetTests :
...
+        
+        response.StatusCode.Should().Be(HttpStatusCode.TooManyRequests);
+        response.Headers.Location.Should().BeNull();
+        var asset = await dbContext.Images.FindAsync(assetId.ToString());
+        asset.Id.Should().Be(assetId.ToString());
     }
     
-    [Fact]
-    public async Task Get_Assets_In_Space_Returns_NotFound_For_Missing_Space()
+    [Fact(Skip = "Is this an expected behaviour?")]
```
donaldgray, line 119: @tomcrane - should we set the `Asset.Error` property if a) The call to enqueue or ingest fails AND b) the `Asset.Error` property is empty? 

I don't think deliverator does this now and always assumes that the engine request will succeed, however I think it could be useful. Without it an image could sit in a "ingesting" limbo

PR #394, opened 2022-10-05
Customer image API endpoints

donaldgray:
Implemented `/customer/{customer}/allImages` and `/customer/{customer}/deleteImages`.

Added new `CustomerImagesController` within `Customer` feature for these endpoints. 

Both of these take a `HydraCollection<IdentifierOnly>`. Identifier only is the minimum class to just have `{"id": "_value_" }`

Followed same approach as current Deliverator, `/deleteImages` returns `200: {"message": "images deleted"}` on success - this should be something better. Also maintained use of `POST` but should arguably be `DELETE`. A slight change in behaviour is that a 400 is returned if nothing was deleted, deliverator returns 200.

`DeleteImage` handler uses `DeleteFromQueryAsync()` from new EntityFramework+ nuget package. This runs a sql DELETE without needing to load assets first. The downside of this is that the handler raises asset modified notification for _all_ asset ids sent to handler but there's a chance 1 or more of these didn't delete anything.

Handlers for both endpoints share similar validation - namely checking that requested Ids are valid asset Ids and they are all for same customer. This was pulled out to `ImageIdListValidation` which is run within controller in addition to FluentValidator validation of actual request.

Added a new `HandleHydraRequest` handler to `HydraController` this avoids needing to have the following error handling in multiple places:

```cs
try
{
    // run code
}
catch (APIException apiEx)
{
    return this.HydraProblem(apiEx.Message, null, apiEx.StatusCode ?? 500, apiEx.Label);
}
catch (Exception ex)
{
    return this.HydraProblem(ex.Message, null, 500, errorTitle);
}
```

Modified `AssetId.FromString` to return a `FormatException` with consistent message as it makes handling it easier. The default message if customer or space was non-numeric is `"Input string was not in a correct format."`

### DIFF
--- a/src/protagonist/API.Tests/Features/Customer/Validation/AllImagesValidatorTests.cs
+++ b/src/protagonist/API.Tests/Features/Customer/Validation/AllImagesValidatorTests.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using API.Features.Customer.Validation;
+using API.Settings;
+using DLCS.Model;
+using FluentValidation.TestHelper;
+using Hydra.Collections;
+using Microsoft.Extensions.Options;
+
+namespace API.Tests.Features.Customer.Validation;
+
+public class AllImagesValidatorTests
+{
+    private readonly AllImagesValidator sut;
+
+    public AllImagesValidatorTests()
+    {
+        var apiSettings = new ApiSettings { MaxImageListSize = 4 };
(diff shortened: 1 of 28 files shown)


donaldgray:
Closed and opened to get GH action to pick up

tomcrane (approved):

PR #396, opened 2022-10-12
Implement last required endpoints for Wellcome

donaldgray:
Implement the following API endpoints:

## POST `/customers/{customerId}/queue/batches/{batchId}/test`

All logic done in request handler, using DlcsContext.

## DELETE `/customers/{customerId}/resources/pdf/{named-query}?args=`

This involved moving around a number of NamedQuery related classes, primarily from Orchestrator to DLCS.Repository or DLCS.Models. The classes that remain in Orchestrator are all related to creating projections, the common classes are related to parsing NQ requests and generating relevant results object (containing `IQueryable<Asset>`, this is only materialised when creating projections).

Sizable refactoring to replace `CustomerPathElement` in `ParsedNamedQuery` with `int CustomerId`. `CustomerPathElement`is not using in API and it's only required when projecting to IIIF manifest.

Introduced `NamedQueryStorageService`, which consists of methods that have been removed from `StoredNamedQueryManager`. The methods are for interacting with stored projection and control files stored S3, regardless of the type of projection. All required paths are in the `StoredParsedNamedQuery` object passed to methods.

This keeps all _what bucket/key is this projection stored at_ logic together and is what the `DeletePdf` handler uses for removing stored assets.

## GET `/customers/{customerId}/spaces/{spaceId}/images/{imageId}/metadata`

Refactored some existing ElasticTranscoder classes. Moved logic for writing MetadataObject to `ElasticTranscoderWrapper` and added method for reading it again. This keeps the logic for reading + writing metadata in single place.

When fetching job data it is converted to custom `TranscoderJob` type - this very closely matches the returned ET type with a few fields ommited. This copies the return type from Deliverator but ommits the ET job ARN as this seems unnecessary.

## Other

I also tidied dependency registrations in API `Startup` class, grouping together related classes into `AddCaching()`, `AddAws()` and `AddDataAccess()` extension methods. As part of doing this I refactored some small areas:
* Renamed `DeliveratorApiAuth` -> `DlcsApiAuth` as it's used by both Protagonist + Deliverator
* Remove usage of `SHA256Managed` class as it's obsolete, instead using base `SHA256` class.
* Extends global usings
* Remove `DlcsSettings.PortalThumb` setting as not used
* Remove `ApiSettings.ApiClientSettings` as not used
* Tweak `HostAssetAtOrigin` to dispose of stream, this broke tests which relied on checking stream length so updated these to use LocalStack.


donaldgray:
Note on history - changed base branch then closed + reopened to kick off a build.

tomcrane (approved):

PR #398, opened 2022-10-13
API Tweaks/Tidy

donaldgray:
Opening this as stopping API work for time being. It's a number of small changes, initially brought on as part of some bugfixes when running through API testing in dev environment. Some of the changes are purely for consistency.

* Fix XML comments for swagger. All of our comments started with `/// <HTTP-VERB> <Route>`  but this looked odd in Swagger as it already shows the verb + route. Remove those and added description and sample requests in `<remarks>` where made sense. Fuller use of swagger could serve as living documentation.
* Moved converters to feature folders, with the exception of `AssetConverter` as that is used in multiple places.
* Removed API referencing API.Client project as it's not used. Also removed nuget package that wasn't used.
* Remove unused usings
* Add/alter xml comments outwith controllers.

I also removed a few comment from controllers similar to:
```
/// This controller does not do any data access; it creates Mediatr requests and passes them on.
/// It converts to and from the Hydra form of the DLCS API.
```
as this describes how things are done, rather than what the controller is/does. We should probably create an ADR to outline the above.


donaldgray:
Closed + reopened to start a build

tomcrane (approved):

PR #414, opened 2022-11-17
Update thumbReorganiser to handle thumbnailPolicy larger than source

donaldgray:
Previous impl always created bounded thumbs in accordance to thumbnailPolicy sizes, this fix is to create based on available size, even if that is smaller. This matches logic in s.json and engine.

E.g. if policy was 1024,400,200,100 and source was 426x780, previously reorg would have created:

* `/open/1024.jpg`
* `/open/400.jpg`
* `/open/200.jpg`
* `/open/100.jpg`

but now it'll create

* `/open/780.jpg`
* `/open/400.jpg`
* `/open/200.jpg`
* `/open/100.jpg`

s.json and output sizes on info.json remain unchanged but mean requests for `/426,/` or `/,780/` will pass as advertised.

### DIFF
--- a/DLCS.Repository.Tests/Assets/ThumbReorganiserTests.cs
+++ b/DLCS.Repository.Tests/Assets/ThumbReorganiserTests.cs
@@ -204,6 +204,64 @@ namespace DLCS.Repository.Tests.Assets
                         "application/json"))
                 .MustHaveHappened();
         }
+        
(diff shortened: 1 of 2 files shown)


donaldgray:
This has been deployed to dlc.services and has resolved their issues.

tomcrane (approved):

Issue #393, opened 2022-10-04
Assets may be available in multiple ways


tomcrane:
When registering, or patching an asset you should be able to specify multiple ways in which the DLCS will make it available.

**At the moment:**

- For `File` family assets, they will always (for now at least) be served with their registered mime type at the `/file/...` path.
- For `Image` family assets, a IIIF Image Service will be made available at the path /iiif-img/ and there is no way of obtaining the original.
- For `TimeBased` family assets, the configured transcoding pipelines will be run and the derivatives made available under the /iiif-av/ path.

To meet #62 and wider requirements, **you should be able to**:

 - if you register a `TimeBased` asset you should be able to specify which transcoding pipelines run (by DLCS friendly name, not directly exposing Elastictranscoder details), including NO pipelines (existing `ImageOptimisationPolicy` property but allowing for nothing to be available at /iiif-av/ - explicit named "nothing" policy rather than default policy)
 - if you register a `TimeBased` _or_ `Image` asset you should be able to specify that the original file is _also_ available at the /file/ path. So, for example, you can register a JP2, have it available as an image service but also have it as the original JP2 (which you can provide in the Manifest as a `rendering` property).
 - If you want to serve a registered AV file as-is (#62), you register it with the flag to serve as File and specify no pipelines - but the Asset is still TimeBased, it has a duration (and w,h too if video), it appears as a `painting` anno in Manifest named queries etc.
 
This doesn't mean multiple registrations per asset.
The Orchestrator would see a request for a TimeBased asset (for example), but on the /file/ path. If the Asset record has this flag, it allows the request and just serves as a file.

This also leaves the door open for `File` types to have their own `ImageOptimisationPolicy` - for Wellcome we initially register all File assets with a "nothing" policy, but later could introduce PDF conversion, html, richtext etc without changing this model. It doesn't make  sense right now to register a File and not make it available at the /file/ path because there's no other way of seeing it, but later a File could have transformations and be explicitly NOT available.
This might also see the AssetFamily enum gain new members, e.g., `TextDocument` as something that could conceivably be "transcoded" to text/plain or PDF, leaving File as the catch-all for untransformed files - but that would mean some big updating.

I think that in terms of the model, `Asset` just gains a boolean `AsOriginal`.
This is true by default when registering `File` AssetFamily but false by default for Image and TimeBased (?)
For Wellcome we could patch the Asset table for images to make the JP2 available - there's no Engine work available here.

[...]

Issue #389, opened 2022-09-15
Serving File based assets


donaldgray:
Current implementation of `/file/` doesn't apply access-control logic like other famililes - this needs to happen.

Once #384 has been implemented we could look at using Yarp to handle file requests.

Issue #384, opened 2022-09-06
Engine processing of File family


tomcrane:
At the moment, Assets of type `F` for File do not touch Engine at all; they are fetched from their origin at runtime when required for orchestration.

But this means that:

 - If the origin disappears we can no longer serve the file, even though it wasn't an _optimised_ origin
 - there's nowhere to hook any processing in future.

At the very least we should apply the same rules as for #64 and copy files to the storage bucket. So we need to notify engine, and engine needs to copy, if the origin is not optimised.

See also 

#259 #271

PR #431, opened 2022-12-01
Delivery Channel engine handling

donaldgray:
Introduced `DeliveryChannel` concept as a way to control how an asset is available. This is based on [RFC-009](https://github.com/dlcs/protagonist/blob/develop/docs/rfcs/009-asset-family-improvements.md). The supported values are:

* `file` - source file is available to download on `/file/`
* `iiif-img` - image-service and thumbs are available
* `iiif-av` - transcoded Audio and Video files are available.

`AssetFamily` is still very much part of the internal workings of the codebase, to remove it would have been a mammoth task.

This PR covers the following tickets (the latter was to allow testing of file delivery):

* https://github.com/dlcs/protagonist/issues/393
* https://github.com/dlcs/protagonist/issues/384

Added `Asset.DeliveryChannel`, which added a migration. The migration will set the value in `DeliveryChannel` depending on the current `AssetFamily`. Below outlines the changes by application:

## API

Altered validation to use require `mediaType`, removing any rules around `AssetFamily`.

If `DeliveryChannel` is provided the `AssetPreparer` will overwrite the `AssetFamily` using the provided channels. If a payload has both an AssetFamily and DeliveryChannel value set the resulting AssetFamily value in the database will be the one determined by the DeliveryChannel.

Introduced a collection of appSettings that represent presets for an assetFamily - these use `AssetFamily` and mediaType to find presets for ImageOptimisationPolicy, ThumbnailPolicy and DeliveryChannel values. These presets are only used if the corresponding value is empty in the payload.

Any logic around the API _not_ sending `F` assets to the Engine have now been removed. These are treated similar to `I` assets - synchronously processed if PUT, else put on queue. Introduced a new 'file' queue for this.

## Engine

Sizable changes in the engine to cope with the delivery-channel logic, as the logic is no longer `File || Timebased || Image`. This also handles #384

Added a new listener for File queue, the handler for this is the same as Image and Timebased ingestion (ie `IngestHandler`). 

The main new moving parts are:
* `IAssetIngesterWorker` - this has a single `Ingest` method and there are 3 implementations for the 3 main delivery-channels: `FileChannelWorker`, `TimebasedIngesterWorker` and `ImageIngesterWorker` (details below).
* `IAssetIngesterPostProcess` - this has a single `PostIngest` method that allows any clean-up, or additional processing to be carried out. The `ImageIngesterWorker` is the only class that implements this to orchestrate after ingestion (if required).
* `IWorkerBuilder` - this works out which `IAssetIngesterWorker` are required to handle the ingestion of a single asset. As detailed in that class the ordering is important.
* `IngestExecutor` - this calls the `IWorkerBuilder`, iterates through the different `IAssetIngesterWorker` implementations, handles result sets and updates asset in database upon completion.

The above classes should allow us to add further delivery-channels in the future, adding different implementations and logic to the `IWorkerBuilder` for when to use them.

> Clean up of old assets is not done - this will be carried out in another ticket, e.g. if an asset was available on "file" and "iiif-img" channel but changes to "iiif-img" only, we won't clean up the copied origin file.

### Workers

As detailed above there are 3 main workers for each of the different channels, below is an outline of how these work.

They all share some common logic for checking if storage size exceeded and bailing out if so etc.

#### `FileChannelWorker`

This will always run first. This is to handle situations where we have both "file" and "iiif-img" or "iiif-av" channels as we need some knowledge from this handler in the respective channel handlers.

This is the simplest worker. If the COS is optimised it's a no-op. Else..
* file is uploaded from origin to dlcs storage, at key `/{cust}/{space}/{asset}/original` _this is not the key that we would have chosen but provides backwards compat with deliverator_
* `IngestionContext` asset size is increased with the size of the origin file.
* `IngestionContext.StoredObjects` dictionary is updated. This stores the `/{cust}/{space}/{asset}/original` path and the size of the file.

#### `ImageIngesterWorker`

This downloads the file from Origin to local disk then hands off to `AppetiserClient` (unchanged). However, `AppetiserClient` has a bit more logic now (some of the below is not new but outlining full logic now).

Appetiser generation logic:
* If file is `use-original` AND it's image/jp2 appetiser will generate thumbnails only.
* Else appetiser will generate a derivative AND thumbnails. (If `use-original` derivative is only used to generate thumbnails, it's disregarded from then on).

For storing files:
* If file is `use-original` AND optimised, no image-server file is stored. `ImageStorage.Size = 0`, `ImageLocation.S3 = {path to origin}`.
* Else if file is `use-original` AND not optimised, we store the _origin_ file at `/{cust}/{space}/{asset}/original`. `ImageStorage.Size = {size of origin}`. `ImageLocation.S3 = {path to /original}`. _`IngestionContext.StoredObjects` is checked at this stage to verify if `FileChannelWorker` has already stored the file to avoid unnecessary work_.
* Else, we store the _origin_ file at `/{cust}/{space}/{asset}`. `ImageStorage.Size = {size of origin}`, `ImageStorage.S3 = {normal path to derivative}`.

> Notes
> In all cases we store thumbnails in DLCS storage and record that value in the ImageStorage.
> If not optimised AND "file" delivery-channel we can end up with an asset at both `/{cust}/{space}/{asset}/original` and a derivative at `/{cust}/{space}/{asset}`. `ImageStorage` record will show combination of both sizes.

#### `TimebasedIngesterWorker`

This transfers the file from Origin to ElasticTranscoder input bucket and creates a new ElasticTranscoder job. Each ET job has a `UserMetadata` dict that is echoed back when job is complete. We use this to store the `AssetId` for knowing which record to update when finished. We now add a key of `"storedOriginSize"` - this is taken from the `IngestionContext.StoredObjects` record. This will allow us to know accurately know what the `ImageStorage` record should show - whether it's the total size of the generated derivatives, or the generated derivatives + the size of the origin file.

When this worker finished the database record isn't updated, that happens in `TimebasedIngestorCompletion` which handles the callback from ET.

> Notes
> If origin not optimised AND "file" delivery-channel we can end up with an asset at both `/{cust}/{space}/{asset}/original` and a derivatives at `/{cust}/{space}/{asset}/{audio or video path}` plus `/{cust}/{space}/{asset}/metadata`. `ImageStorage` record store combination of all data with the exception of the metadata object as it's only a few bytes of XML.

## Orchestrator

The `OrchestrationAsset` record stored in memory was altered - previously this was aligned with `T`, `I` and `F` records but this is not possible now. The main changes are:
* Added flags enum `AvailableDeliveryChannel` which stores the channels available. All delivery-channels check this and 404 if incorrect channel.
* Added `string? Origin` property to cached asset. This is only used for File delivery channel and is used to determine where and how to download the file.

The `GetFile` handler was radically changed, it uses above `Origin` property to determine what COS to use to download the file (e.g. download from Origin if optimised or DLCS storage if not). Given that the DLCS storing files is a new concept I added a `StreamMissingFileFromOrigin` appSetting - when this is `True` we will attempt to stream from DLCS (new behaviour) and if it's not found we will stream from origin (previous behaviour).

> We don't do anything 'nice' with range headers etc when requesting files - it's a basic stream operation.
> File handling still doesn't have auth, that is covered in https://github.com/dlcs/protagonist/issues/389


donaldgray:
Reminder: This PR contains a migration

donaldgray:
PR question reminder - what should the names of the image and av channels be? This needs reflected in code + ADR

PR #438, opened 2023-01-12
Implement delete handler

donaldgray:
New project that listens to SQS queue and, on receipt of message, will delete:

1. Thumbnails
2. Image from "Storage" bucket (JP2)
3. JP2 from local storage.

Each of the above steps will only run if the relevant appSetting value is populated, this is to allow flexibility in how we deploy - e.g. we may have 1 instance that has NAS mounted so deletes from fast-storage and X Fargate instances handling 1+2.

`AssetDeletedHandlerTests` are unlike the other integration tests in sln as we cannot use `WebApplicationFactory<>` method as it uses different hosting model - use mocks instead.

This handles messages in format `{ "id": "1/2/foo" }`. Added `QueueMessage.GetMessageContents()` to avoid handlers needing to know how the image arrived (via SNS->SQS with our without 'raw message delivery' or directly to SQS).

Reused SQS classes from `DLCS.AWS` but added some helpers as this project has a single listener, rather than different possible listeners like Engine.

This is currently used for deliverator so will need expanded to handle Protagonist resources (e.g. `T` assets, info.json etc).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -32,13 +32,13 @@ Utils/
 **/Properties/launchSettings.json
 
 # build results
-[Dd]ebug/
-[Dd]ebugPublic/
-[Rr]elease/
-[Rr]eleases/
-x64/
-x86/
-bld/
-[Bb]in/
-[Oo]bj/
-[Ll]og/
+**/[Dd]ebug/
+**/[Dd]ebugPublic/
+**/[Rr]elease/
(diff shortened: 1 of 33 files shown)


donaldgray:
See https://github.com/digirati-co-uk/dlcs-spinup/pull/33 for sample TF changes for standard setup

tomcrane (approved):

PR #453, opened 2023-02-01
maxUnauthorised change becomes a requiresReingest operation

tomcrane:
Adds maxunauthorised as something that will cause an error if changed in a batch patch operation.

This is more likely to catch people out as it's an int, so more likely to arrive in a payload with a default `0` value when the caller didn't intend to set it.

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/AssetPreparer.cs
+++ b/src/protagonist/DLCS.Model/Assets/AssetPreparer.cs
@@ -107,6 +107,14 @@ public static class AssetPreparer
             {
                 requiresReingest = true; // YES, because we've changed the way this image should be processed
             }
+
+            // This is ONLY true if we need engine to set the permissions on S3 for redirects.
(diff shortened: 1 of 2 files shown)


src/protagonist/DLCS.Model/Assets/AssetPreparer.cs
```diff
@@ -107,6 +107,14 @@ public static class AssetPreparer
             {
                 requiresReingest = true; // YES, because we've changed the way this image should be processed
             }
+
+            // This is ONLY true if we need engine to set the permissions on S3 for redirects.
+            // It is not true if orchestrator proxies requests for open AV.
+            // See https://github.com/dlcs/protagonist/issues/452
```
donaldgray, line 113: Is this one of those comments that'll seem odd on 2 years if it's still there? Is it enough to have the issue in the backlog?

donaldgray (approved):

PR #455, opened 2023-02-07
Don't notify engine if no assets in list, AND save batch!

tomcrane:
```
 assetNotificationSender.SendIngestAssetsRequest(assetNotificationList ... )
```

This assumed assetNotificationList had members, which if they are all File family, it won't.

Also, the batch was never being saved here.

### DIFF
--- a/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
+++ b/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
@@ -141,10 +141,18 @@ public class CreateBatchOfImagesHandler : IRequestHandler<CreateBatchOfImages, M
         }
         else
         {
-            // Raise notifications
-            logger.LogDebug("Batch {BatchId} created - sending engine notifications", batch.Id);
-            await assetNotificationSender.SendIngestAssetsRequest(assetNotificationList, request.IsPriority,
-                cancellationToken);
+            if (assetNotificationList.Count > 0)
+            {
+                // Raise notifications
+                logger.LogDebug("Batch {BatchId} created - sending engine notifications", batch.Id);
+                await assetNotificationSender.SendIngestAssetsRequest(assetNotificationList, request.IsPriority,
+                    cancellationToken);
+            }
+            else
+            {
+                logger.LogDebug("There are no assets to ingest (was this batch all File?)");
+            }
+            await dlcsContext.SaveChangesAsync(cancellationToken);
         }
         
         return ModifyEntityResult<Batch>.Success(batch, WriteResult.Created);


donaldgray (changes requested):
There's no need to explicitly save the Batch here, it's taken care of higher up via the Transaction, so `await dlcsContext.SaveChangesAsync(cancellationToken);` can be removed. Explanation...

The batch is created at line 78:

```cs
var batch = await batchRepository.CreateBatch(request.CustomerId, request.Assets, cancellationToken);
```

At this point it has `Count`, `Customer` and `Submitted` values set and is saved to DB to generate the id value. If we have all `I` or `T` values this is all that is needed as it's then passed to Engine to update `Completed`/`Error` counts as necessary.

However, if we have at least 1 `F` value then the `Batch.Complete` count is incremented because it never goes to the Engine (which is changing in near future).

Once we have iterated over all Assets posted we check to see if `batch.Completed` has been incremented and if so mark the batch as modified in current context so that it is updated when the current Transaction is committed (line 115):

```cs
if (batch.Completed > 0)
{
    dlcsContext.Batches.Attach(batch);
    dlcsContext.Entry(batch).State = EntityState.Modified;
}

if (!updateFailed)
{
    await transaction.CommitAsync(cancellationToken);
}
```

tomcrane:
I wonder if what I was seeing here was a result of the transaction isolation.

> Shared locks are held while the data is being read to avoid dirty reads, but the data can be changed before the end of the transaction, resulting in non-repeatable reads or phantom data.

If the dashboard page is generated while this update is taking place, it could be reading data that is then removed on transaction rollback.

tomcrane:
There are problems resulting from timeouts, too.

This can be observed in the debugger, if you let things take too long.

The incoming operation (as in Http operation) has a 30s timeout; if the whole API interaction takes longer than this, the operation will be cancelled.

This causes a failure here, setting `updateFailed` to true:

```
var processAssetResult =
                    await assetProcessor.Process(asset, false, true, true, cancellationToken: cancellationToken);
                if (!processAssetResult.IsSuccess)
                {
                    logger.LogDebug("Processing asset {AssetId} failed, aborting batch. Error: '{Error}'", asset.Id,
                        processAssetResult.Result.Error);
                    updateFailed = true;
                    failureMessage = processAssetResult.Result.Error;
                    break;
                }
```

So then we get into this section:

```
        if (updateFailed)
        {
            await transaction.RollbackAsync(cancellationToken);
            
            dlcsContext.Batches.Remove(batch);
            await dlcsContext.SaveChangesAsync(cancellationToken);
            
            return ModifyEntityResult<Batch>.Failure(failureMessage, WriteResult.Error);
        }
```

BUT we can't roll back the transaction - `await transaction.RollbackAsync(cancellationToken);` throws an exception, "The operation was cancelled". The whole `Handle` method bails out with an exception.

donaldgray:
Does this commit resolve some of the issues? https://github.com/dlcs/protagonist/pull/466/commits/56e488e3261f8c0c1a12f9f65c047a11c44b8773

> I wonder if what I was seeing here was a result of the transaction isolation.
> 
> > Shared locks are held while the data is being read to avoid dirty reads, but the data can be changed before the end of the transaction, resulting in non-repeatable reads or phantom data.
> 
> If the dashboard page is generated while this update is taking place, it could be reading data that is then removed on transaction rollback.

The non-repeatable reads are during the transaction. The dashboard won't be able to read any changes made as part of the until it's committed. The isolation level could be an issue but would only be so if something changed an Image row outside of the transaction, but between 2 different reads of that row in the transaction. See https://stackoverflow.com/questions/4034976/difference-between-read-commited-and-repeatable-read-in-sql-server

> BUT we can't roll back the transaction - await transaction.RollbackAsync(cancellationToken); throws an exception, "The operation was cancelled". The whole Handle method bails out with an exception.

Updated logic to only call Rollback if cancellation not required. Rollback is implicitly called if transaction is aborted so arguably we don't need to. This _should_ only have resulted in an incorrect result though, the Rollback should still have been rolledback.

PR #466, opened 2023-03-28
Weco snagging

donaldgray:
Various fixes identified when testing weco deployment

## info.json sizes

Use thumbnail sizes in infojson. 

These overwrite the suggestions returned from image-server. 

If thumbnails cannot be returned the image-server values are used. Add **all** thumbnail sizes, regardless of open or auth. These are read from the `s.json` file stored in S3 rather than the thumbnailPolicy, so - "as-is" rather than "should-be".

## Increment entity counters

EntityCounters were not being incremented when assets are being ingested. Altered save logic in `ApiAssetRepository` to increment `customer-space` and `space-images` counters on asset creation.

Added `KnownEntityCounters` which has the used entity-counter types. These are strings for now - if we want to persist when EC logic then we should have helper methods (ie `IncrementSpaceCounter`) rather than relying on caller knowing the correct `Customer` and `Scope` values to use per type.

Also removed unused methods in EntityCounter repo as these had been copied from Deliverator but are not being used - we can add them back in if needed in the future.

## Handle NullRef exception if no open thumbs

Default cached `OrchestrationImage.OpenThumbs` to empty list if not thumbs found. Avoids nullRefException in orchestrator.

## InfoJson links

Updated `InfoJson` and `ThumbnailInfoJson` fields on Asset API hydramodel to contain `/info.json` prefix as without this it'll result in a 302 response.

## Transaction / DB Changes

Rework `DapperRepository` helpers to ensure that transactions started in EF work with Dapper.

Tweak `EngineAssetRepository` to only create a Transaction when updating Asset + Batch if there is currently no transaction running. This worked okay as it was but nested transactions are not ideal.

## Update Statements

Reworked some `UPDATE tbl SET col = col + 1` style logic, both direct SQL queries and loading + updating, to use [UpdateFromQuery](https://entityframework-extensions.net/update-from-query) method from linked nuget package. Noticed issue when ingesting 20+ items, where `Batch.Complete` would be less than expected, issue was due to batch being loaded + updated rather than in 1 query. Using LINQ will make transition to proper cases postgres table syntax easier.

Using this in `CustomerQueueRepository` removed need for repository.

## Simplify Batch creation

The above change allowed `CreateBatchOfImages` to use EF for batch creation within a transaction. This simplified rollback/commit logic.

Modified to only call `transaction.Rollback()` if `cancellationToken` **not** cancelled. Ultimately this is unnecessary because it's implicitly called when disposing of the transaction but it helps to keep it explicit.

## Elastic Transcoder Failures

Better handling of ElasticTranscoder failure notifications - log each outputs failure status in addition to general failure. This can result in a long error message but better for troubleshooting.

This change requires ET error notifications to be configured in AWS.

### DIFF
--- a/src/protagonist/DLCS.Repository/Processing/CustomerQueueRepository.cs
+++ b/src/protagonist/DLCS.Repository/Processing/CustomerQueueRepository.cs
@@ -54,32 +54,25 @@ q.""Customer"", q.""Size"", q.""Name"", b.""BatchesWaiting"", b.""ImagesWaiting"
 
     private async Task ChangeQueueSize(int customer, string name, int amount, CancellationToken cancellationToken)
     {
-        await using var transaction =
-            await DlcsContext.Database.BeginTransactionAsync(IsolationLevel.ReadCommitted, cancellationToken);
-
         try
         {
-            var queue = await DlcsContext.Queues.SingleOrDefaultAsync(q => q.Customer == customer && q.Name == name,
-                cancellationToken: cancellationToken);
+            var updateCount = await DlcsContext.Queues
+                .Where(q => q.Customer == customer && q.Name == name)
+                .UpdateFromQueryAsync(q => new Queue { Size = Math.Max(0, q.Size + amount) }, cancellationToken);
 
-            if (queue == null)
+            if (updateCount == 0)
             {
+                logger.LogInformation("Updating queue {QueueName} for {Customer} customer returned 0 results. Creating",
+                    name, customer);
                 await DlcsContext.Queues.AddAsync(
                     new Queue { Customer = customer, Name = name, Size = Math.Max(0, amount) }, cancellationToken);
+                await DlcsContext.SaveChangesAsync(cancellationToken);
             }
-            else
-            {
-                queue.Size += amount;
-                if (queue.Size < 0) queue.Size = 0;
-            }
-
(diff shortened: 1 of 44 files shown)


src/protagonist/DLCS.Repository/Processing/CustomerQueueRepository.cs
```diff
@@ -54,32 +54,25 @@ public CustomerQueueRepository(DlcsContext dlcsContext, ILogger<CustomerQueueRep
...
     {
-        await using var transaction =
-            await DlcsContext.Database.BeginTransactionAsync(IsolationLevel.ReadCommitted, cancellationToken);
-
         try
         {
-            var queue = await DlcsContext.Queues.SingleOrDefaultAsync(q => q.Customer == customer && q.Name == name,
-                cancellationToken: cancellationToken);
+            var updateCount = await DlcsContext.Queues
+                .Where(q => q.Customer == customer && q.Name == name)
```
tomcrane, line 60: Would SingleOrDefault(..) be more expressive here?

donaldgray: It would but the `UpdateFromQueryAsync()` method is an extension method for `IQueryable<T>`, and `SingleOrDefault()` would return a `T` only.

tomcrane (approved):

Issue #397, opened 2022-10-13
Update Orchestrator to Reingest assets using Protagonist


donaldgray:
The orchestrator has an implementation of `IDlcsApiClient` that uses implementation `DeliveratorApiClient` - update this to be `ProtagonistApiClient` OR update to use the `IDlcsClient` from API.Client project. 

If using the former, there should be very little (if any) changes required - maybe just renaming the class.

If going down the latter route then need to consider how auth works, currently it takes a `ClaimsPrincipal` which assumes there's already a principal with the relevant claims set. This suits the portal but doesn't necessarily suit making ad-hoc requests using API credentials that aren't part of the original request. `DeliveratorApiClient` currently looks in DB to get API keys for relevant customer and sets them when making a request.

PR #481, opened 2023-04-13
Rename DeliveratorApiClient to ApiClient

donaldgray:
For #397. Chose to just rename the current `DeliveratorApiClient` and leave implementation intact.

Didn't reuse the `API.Client/DlcsClient` as it is unnecessary to pull in that full interface just to call reingest. As detailed in #397, the `DlcsClient` uses a `ClaimsPrincipal`, which doesn't fit how Orchestrator handles auth separate interface was cleaner.

### DIFF
--- a/src/protagonist/API.Client/DlcsClient.cs
+++ b/src/protagonist/API.Client/DlcsClient.cs
@@ -217,11 +217,6 @@ public class DlcsClient : IDlcsClient
         return patched;
     }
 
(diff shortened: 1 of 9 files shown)


tomcrane (approved):
Early on, when we were still in hybrid mode the early protagonist was still able to call the Deliverator API assuming there was still a Deliverator API running somewhere. I assume there is no longer any example of new API calling old API (I can't see one)

PR #485, opened 2023-04-18
Handle images smaller than thumbnail-policy

### LINKED ISSUE #32: Ensure images smaller than thumbnail policy don't break anything
API consumers shouldn't have to worry if an image in their workflow is smaller than a configured thumbnail policy.
DLCS should just handle this gracefully - not try to make the bigger images, or advertise them as available.

This affects Orchestrator, Thumbs, Engine...

### LINKED ISSUE #415: Thumbnail layout
PR #414 was a hotfix for an issue related to pregenerated thumbnail layout in S3. The fix aligns the thumbnails `EnsureNewThumbnailLayout` logic with deliverator Protagonist.

This layout need to be reflected in how new thumbs reorganises thumbnails _and_ how the engine sets them up in the first instance.

donaldgray:
Resolves #415 and resolves #32

Fix for #415 was to replicates a hotfix made straight to main branch (#414) where the logic in `ThumbReorganiser` was updated to not look at the expected sizes from the thumbnailPolicy, but instead to look at what sizes are actually available.

Main change is to ensure that we handle an image with dimensions smaller than a required size in thumbnail policy. If this happens we will create as large an image as we can, even if it doesn't match a thumbnail size, without scaling up.

E.g. policy `800,400,200,100` for image with w,h of 300,150 would create thumbs of size:
* 300,150
* 200,100
* 100,50

Handling for this was in a few different places, detailed below.

**`AssetX.GetAvailableThumbSizes()`**

Previously this was returning a number of sizes equal to the number of sizes in thumbnail policy. E.g., for above example it would return `[[300,150],[300,150],[200,100],[100,50]]` and always returned a `maxBoundedSize` equal to the largest thumbnail (so `800` for above).

Modified this to not return duplicate sizes and return the actual max size.

**`ThumbCreator.CreateNewThumbs()`**

This now calls the above method to get a list of all possible sizes. It then iterates through the returned sizes from Appetiser, for each one it gets the sizes from above list to avoid rounding error previously seen between dotnet + python code.

Similar to how `GetAvailableThumbSizes()` previously worked, Appetiser returns an array of size equal to number of sizes requested. This can lead to duplicates which are ignored.

`s.json` will reflect the actual size of thumbnails uploaded (based on results of `GetAvailableThumbSizes()`) - this will in turn be used to generate info.json and advertise these sizes as the ones available.

### DIFF
--- a/src/protagonist/DLCS.Core.Tests/Guard/GuardXTests.cs
+++ b/src/protagonist/DLCS.Core.Tests/Guard/GuardXTests.cs
@@ -32,6 +32,31 @@ public class GuardXTests
         actual.Should().Be(val);
     }
     
+    [Fact]
+    public void ThrowIfNull_NullableStruct_Throws_IfArgumentNull()
+    {
+        // Act
+        Action action = () => GuardX.ThrowIfNull<int?>(null, "foo");
+        
+        // Assert
+        action.Should()
+            .Throw<ArgumentNullException>()
+            .WithMessage("Value cannot be null. (Parameter 'foo')");
+    } 
+    
+    [Fact]
+    public void ThrowIfNull_NullableStruct_ReturnsProvidedValue_IfNotNull()
+    {
+        // Arrange
+        int? val = 12345;
+        
+        // Act
+        var actual = val.ThrowIfNull(nameof(val));
+        
+        // Assert
+        actual.Should().Be(val);
+    }
+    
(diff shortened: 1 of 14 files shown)


tomcrane (approved):
Something about saving the low.jpg, and the rounding error stuff... We might return to thumbs in future I feel. But seems OK for now.

Issue #463, opened 2023-03-27
What happens when a legacy API call is supported?


donaldgray:
With the removal of asset family in #431, `mediaType` is now a lot more important. If there is no `mediaType` provided during ingest the DLCS will throw a 400 exception.

For backwards compatibility we may need to allow `mediaType`-less ingests. If that's the case a config value should enable/disable this per customer and some sensible defaults configured. E.g. looking at URL and trying to workout from the file name.

tomcrane:
> per customer

(v important!) - and customer can control this in portal.

donaldgray:
Regarding above comment, see https://github.com/dlcs/protagonist/issues/528

tomcrane:
If acceptLegacyApiCalls = true:


# MediaType

* By API, using (for example) file extension?
* By Engine, using mediaType of HTTP response from origin?

If the latter and there is missing mediaType and acceptLegacyPayloads=true then:

* API sets a dummy mediaType "dlcs/unknown" for Engine to correct.

When processed by Engine:

if "dummy" mediaType (set by API)
 - work out mediaType from origin HTTP response
 - if still unknown, or `application/*` or similar, use https://github.com/dlcs/protagonist/blob/main/src/protagonist/DLCS.Core/MIMEHelper.cs

# MaxUnauthorised 

In API:

 - Replace any `0` with `-1` if roles is empty 
 
(this is also the release-day DB update)

# Delivery Channels / Family

In API (not engine):

If `origin` only supplied, assume it is an image
 - Set Family to `I`

(protagonist already implements the default delivery channels from here, whether family is `I`, `T` or `F`)

PR #559, opened 2023-07-19
initial commit for adding the flag

JackLewis-digirati:
resolves https://github.com/dlcs/private-protagonist/issues/70

Adds a custom flag to reenable legacy support to appesttings.

These flags are set like this:

```
{
    "DefaultLegacySupport": true, // default if not defined here
    "CustomerOverrides": {
        "2": {
            "LegacySupport": true, // whether to enable legacy suport
            "NovelSpaces": [1] // if legacy support is enabled, this setting disables legacy support on these spaces
        },
        "4": {
            "LegacySupport": false
        }
    }
}
```

Currently, setting these flags won't do anything, the logic that takes advantage of this will be completed under #463

### DIFF
--- a/src/protagonist/API.Tests/Settings/ApiSettingsTests.cs
+++ b/src/protagonist/API.Tests/Settings/ApiSettingsTests.cs
@@ -1,0 +1,54 @@
+﻿using System.Collections.Generic;
+using API.Settings;
+
+namespace API.Tests.Settings;
+
+public class ApiSettingsTests
+{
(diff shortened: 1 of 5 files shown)


donaldgray (approved):
I think this is fine for now - depending on how it's used it might be useful to have a `ShouldSupportLegacy(int customer, int space);` to contain all of the logic in a single method.

tomcrane (approved):

PR #561, opened 2023-07-20
adding in ability to put/post an image using legacy formats

JackLewis-digirati:
This PR adds in the ability to use protagonist in a similar manner to deliverator via the use of settings.

This is disabled by default using the `DefaultLegacySupport` app setting. Individual customers can also be controlled using the appsettings like the following:

```json
"CustomerOverrides": {
    "2": {
      "LegacySupport": true,
      "NovelSpaces": ["1","2","4"]
    }
```

`NovelSpaces` are to specifically set spaces that are handled in a protagonist manner when legacy support is enabled.  If legacy support is disabled, there is no use to setting `NovelSpaces`

### DIFF
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -95,6 +95,56 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
         asset.ImageOptimisationPolicy.Should().Be("fast-higher");
     }
     
+    [Fact]
+    public async Task Put_NewImageAsset_FailsToCreateAsset_whenMediaTypeAndFamilyNotSet()
+    {
+        var assetId = new AssetId(99, 1, nameof(Put_NewImageAsset_FailsToCreateAsset_whenMediaTypeAndFamilyNotSet));
+        var hydraImageBody = $@"{{
+  ""@type"": ""Image"",
+  ""origin"": ""https://example.org/{assetId.Asset}.tiff""
+}}";
+        A.CallTo(() =>
+                EngineClient.SynchronousIngest(
+                    A<IngestAssetRequest>.That.Matches(r => r.Asset.Id == assetId), false,
+                    A<CancellationToken>._))
+            .Returns(HttpStatusCode.OK);
+        
+        // act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(99).PutAsync(assetId.ToApiResourcePath(), content);
+        
+        // assert
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+    }
+    
+    [Fact]
+    public async Task Put_NewImageAsset_CreatesAsset_whenMediaTypeAndFamilyNotSetWithLegacyEnabled()
+    {
+        const int customer = 325665;
+        const int space = 2;
+        var assetId = new AssetId(customer, space, nameof(Put_NewImageAsset_CreatesAsset_whenMediaTypeAndFamilyNotSetWithLegacyEnabled));
+        await dbContext.Customers.AddTestCustomer(customer);
+        await dbContext.Spaces.AddTestSpace(customer, space);
+        await dbContext.SaveChangesAsync();
+        
+        var hydraImageBody = $@"{{
+  ""@type"": ""Image"",
+  ""origin"": ""https://example.org/{assetId.Asset}.tiff""
+}}";
+        A.CallTo(() =>
+                EngineClient.SynchronousIngest(
+                    A<IngestAssetRequest>.That.Matches(r => r.Asset.Id == assetId), false,
+                    A<CancellationToken>._))
+            .Returns(HttpStatusCode.OK);
+        
+        // act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customer).PutAsync(assetId.ToApiResourcePath(), content);
+        
+        // assert
+        response.StatusCode.Should().Be(HttpStatusCode.Created);
+    }
+    
     [Theory]
     [InlineData("audio/mp3", AssetFamily.Timebased)]
     [InlineData("video/mp4", AssetFamily.Timebased)]
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -143,6 +143,48 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
         
         // assert
         response.StatusCode.Should().Be(HttpStatusCode.Created);
+        
+        var asset = await dbContext.Images.FindAsync(assetId);
+        asset.Id.Should().Be(assetId);
+        asset.MediaType.Should().Be("image/tiff");
+        asset.Family.Should().Be(AssetFamily.Image);
+        asset.ImageOptimisationPolicy.Should().Be("fast-higher");
+        asset.ThumbnailPolicy.Should().Be("default");
+    }
+    
+    [Fact]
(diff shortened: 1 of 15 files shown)


src/protagonist/API/Converters/LegacyModeConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+            
+            if (image.Origin is not null && image.Family is null && image.DeliveryChannels.IsNullOrEmpty())
+            {
+                  image.Family = AssetFamily.Image;
+            }
+        }
+
+        if (image.MaxUnauthorised is null or 0)
+        {
+            image.MaxUnauthorised = -1;
```
donaldgray, line 35: We only want to set to `-1` if Roles property is null or empty. (`Replace any 0 with -1 if roles is empty`).

The reason for this change is that the logic to work out if an asset requires auth has changed from `asset.MaxUnauth = 0 && hasRoles` to `asset.MaxUnauth == 0 || hasRoles`, so if a user provides `{ "maxUnauthorised": 0, "roles": "my-role" }` then that's a valid configuration to say that nothing should be displayed unless current user has `my-role`

src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -23,10 +23,13 @@ namespace API.Features.Image;
 [ApiController]
 public class ImageController : HydraController
 {
+    private readonly ApiSettings _apiSettings;
+    
     public ImageController(
         IMediator mediator,
         IOptions<ApiSettings> options) : base(options.Value, mediator)
     {
+        _apiSettings = options.Value;
```
donaldgray, line 32: For consistency with wider sln, private members shouldn't start with `_`. We should have this enforced config file but we don't (yet).

src/protagonist/API/Settings/ApiSettings.cs
```diff
@@ -55,4 +55,15 @@ public CustomerOverrideSettings GetCustomerSettings(int customerId)
             {
                 LegacySupport = DefaultLegacySupport
             };
+    
+    /// <summary>
+    /// Get whether legacy mode is enabled for the
```
donaldgray, line 60: Comment not finished

src/protagonist/API/Converters/LegacyModeConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+    private const string DefaultMediaType = "dlcs/unknown";
+    
+    /// <summary>
+    /// Converts from legacy format to new format
+    /// </summary>
+    /// <param name="image">The image to convert</param>
+    /// <returns>A converted image</returns>
+    public static Image VerifyAndConvertToModernFormat(Image image)
+    {
+        if (image.MediaType is null)
```
donaldgray, line 21: check for `string.IsNullOrEmpty()` ?

src/protagonist/API.Tests/Settings/ApiSettingsTests.cs
```diff
@@ -51,4 +51,94 @@ public void ApiSettings_LegacySupportEnabledForSingleCustomer_WhenDefaultLegacyD
...
+        settings.LegacyModeEnabled(1, 1).Should().BeTrue();
+        settings.LegacyModeEnabled(2, 1).Should().BeFalse();
+        settings.LegacyModeEnabled(2, 2).Should().BeTrue();
+    }
+
+    [Fact]
+    public void LegacyModeEnabled_LegacySupportDisabledForCustomer_WhenDefaultLegacyEnabled()
+    {
+        // Arrange
+        var settings = new ApiSettings();
```
donaldgray, line 132: Reading this test - is the default set in code for `LegacySupport = true`? If so I think it should be `false`. 

A newly spun up DLCS without any prior customers shouldn't support legacy formats. We will want to set it to `true` as a default in some instances, but that should be via config for existing installs with existing customers. Apologies, I must have missed that on previous PR!

src/protagonist/API/Features/Queues/CustomerQueueController.cs
```diff
@@ -90,6 +93,17 @@ public async Task<IActionResult> GetCustomerQueue([FromRoute] int customerId, Ca
         [FromServices] QueuePostValidator validator,
         CancellationToken cancellationToken)
     {
+        if (images.Members != null)
```
donaldgray, line 96: I wonder if we could shortcut this somehow if there's no changes that LegacyMode could be enabled for a customer? Would it be worth it?

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -95,6 +95,56 @@ public async Task Put_NewImageAsset_Creates_Asset()
...
+                    A<IngestAssetRequest>.That.Matches(r => r.Asset.Id == assetId), false,
+                    A<CancellationToken>._))
+            .Returns(HttpStatusCode.OK);
+        
+        // act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customer).PutAsync(assetId.ToApiResourcePath(), content);
+        
+        // assert
+        response.StatusCode.Should().Be(HttpStatusCode.Created);
```
donaldgray, line 145: Is it worth validating that the updates values are persisted in DB? More so for `dlcs/unknown` mediatype as Engine will require that when it picks up the asset.

donaldgray (changes requested):
What's here looks good and is a lot less code than I had envisaged!

The Engine side of #463 is missing (detecting the dummy media type + setting content-type from origin in its place).

One thing to check is that `ImageOptimisationPolicy` and `ThumbnailPolicy` are set on the asset in the database - could add that to one of the tests. I think at least one of these look at the mediaType 'type' value and only set some if it is `image/*`.

src/protagonist/API.Tests/Settings/ApiSettingsTests.cs
JackLewis-digirati: yeah, must have misinterpreted! - will get it changed so it's false by default

src/protagonist/API/Features/Queues/CustomerQueueController.cs
JackLewis-digirati: it's possible but I think it adds a fair amount of additional complexity. For example - 

```csharp
if (images.Members != null)
{
    bool  legacyEnabled = false;
    int checkedSpace = 0;
    
    for (int i = 0; i < images.Members.Length; i++)
    {
       var space = images.Members[i].Space;
       if ((legacyEnabled && checkedSpace == space)  || (legacyEnabled = apiSettings.LegacyModeEnabled(customerId, space)))
       {
           images.Members[i] = LegacyModeConverter.VerifyAndConvertToModernFormat(images.Members[i]);
       }
     checkedSpace = space;
    }
}
```

Probably needs some work, but I think this gets the idea across?

donaldgray: I was more thinking of checking whether it's possible for the customer to support Legacy (something like `apiSettings.IsItPossibleForCustomerToSupportLegacy(customer)` but less badly named). 

For example we won't need to do any `LegacyModeEnabled` checks for Wellcome as we have already updated their integrations to use Protagonist models, also for new customers in the future. If there's no configuration permutation where they support Legacy then the above is unnecessary.

JackLewis-digirati: ah sure, I get it now - yeah I can add a method to enable this

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
JackLewis-digirati: hi @donaldgray, was just looking at the section on `imageOptimisationPolicy` and `thumbnailPolicy`. From what I can see, if the media type can't be inferred, the `thumbnailPolicy` is set to an empty string (otherwise looks like it's `default`) - is this ok?

donaldgray (approved):
lgtm, does this still need deployed + tested on dev? (prior to merging)

Issue #542, opened 2023-07-03
Delete a Space via API


tomcrane:
Other space operations are already supported.

What does it mean to delete a space?
Can you delete a space that has assets in it or do you have to remove them first?
If yes, are there any limits on the number in the space to be deleted?

- [ ] DELETE `/customers/{customer}/spaces/{space}`

## Acceptance criteria

MVP

 - You can delete a space that has no assets in it
 - It's removed from the DB so you could make another space with same `id` later (via PUT)
 - EntityCounters are updated

PR #563, opened 2023-07-24
adding in delete space method

JackLewis-digirati:
This pull request adds the `deleteSpace` method

### DIFF
--- a/src/protagonist/DLCS.Repository.Tests/Spaces/SpaceRepositoryTests.cs
+++ b/src/protagonist/DLCS.Repository.Tests/Spaces/SpaceRepositoryTests.cs
@@ -10,6 +10,8 @@ using Microsoft.Extensions.Options;
 
 namespace DLCS.Repository.Tests.Spaces;
 
+[Trait("Category", "Database")]
+[Collection(DatabaseCollection.CollectionName)]
 public class SpaceRepositoryTests
 {
     private readonly SpaceRepository sut;


src/protagonist/API/Infrastructure/HydraController.cs
```diff
@@ -77,6 +78,35 @@ protected UrlRoots GetUrlRoots()
...
+    /// <summary>
+    /// Handles a deletion
+    /// </summary>
+    /// <param name="request">The request/response to be sent through Mediatr</param>
+    /// <param name="errorTitle">The title of the error</param>
+    /// <param name="cancellationToken">Current cancellation token</param>
+    /// <typeparam name="T">The type of entity that was to be deleted</typeparam>
+    /// ActionResult generated from DeleteResult. This will be 204 on success. Or a Hydra
+    /// error and appropriate status code if failed.
+    protected async Task<IActionResult> HandleDelete<T>(
```
donaldgray, line 91: `<T>` isn't required here.

src/protagonist/API/Infrastructure/HydraController.cs
```diff
@@ -77,6 +78,35 @@ protected UrlRoots GetUrlRoots()
...
+        IRequest<(DeleteResult deleteReuslt, string errorMessage)> request,
```
donaldgray, line 92: Typo in `deleteReuslt`

src/protagonist/API/Infrastructure/HydraController.cs
```diff
@@ -77,6 +78,35 @@ protected UrlRoots GetUrlRoots()
...
+        return await HandleHydraRequest(async () =>
+        {
+            var result = await Mediator.Send(request, cancellationToken);
+
+            return result.deleteReuslt switch
+            {
+                DeleteResult.NotFound => this.HydraNotFound(),
+                DeleteResult.Error => this.HydraProblem(result.errorMessage, null, 500,
+                    "Delete failed"),
+                _ => NoContent()
```
donaldgray, line 106: It might be worth having `DeleteResult.Deleted => NoContent()` explicitly and a default handler throw an `ArgumentOutOfRangeException`.

With the above we could add a new enum value (`DeleteResult.Conflict`) and it would return a 204 if the above wasn't updated.

src/protagonist/DLCS.Repository/Spaces/SpaceRepository.cs
```diff
@@ -207,4 +208,37 @@ private async Task<int> GetIdForNewSpace(int requestCustomer)
...
+
+        if (space != null)
+        {
+            var images = await dlcsContext.Images.AsNoTracking().Where(i => 
+                i.Customer == customerId && i.Space == spaceId).CountAsync(cancellationToken);
+            
+            if (images == 0)
+            {
+                dlcsContext.Spaces.Attach(space);
+                dlcsContext.Spaces.Remove(space);
```
donaldgray, line 228: To avoid needing to `.Attach()` and then `.Remove()` the space prior to deleting. Alter the query fetching space to remove `.AsNoTracking()`.

```cs
var space = await dlcsContext.Spaces.SingleOrDefaultAsync(s =>
            s.Customer == customerId && s.Id == spaceId, cancellationToken: cancellationToken);
```

donaldgray (changes requested):
LGTM, couple of comments in addition to those below

1. Update the PR name/description as these will be pulled through into the release notes

2. It might be worth using something more concrete than a tuple for `(DeleteResult, string)`, as this might be used in quite a few places.
There is a generic `ResultStatus<T>` which represents a `bool success` and `T value`, we could add something like that (`ResultMessage<T>`?) or a more specific `ModifyEntityResult<T>` which is specific for modifications and contains message/success/entity.

src/protagonist/API.Tests/Integration/SpaceTests.cs
```diff
@@ -379,5 +380,62 @@ private async Task EnsureSpaces(int customerId, int numberOfSpaces)
...
+        // Arrange & Act
+        int? customerId = await EnsureCustomerForSpaceTests("Patch_Space_Updates_Name");
+        var deleteResponse = await httpClient.AsCustomer().DeleteAsync($"/customers/{customerId}/spaces/456453");
+        
+        // Assert
+        deleteResponse.StatusCode.Should().Be(HttpStatusCode.NotFound);
+    }
+    
+    [Fact]
+    public async Task DeleteSpace_InternalServerError_WhenDeletingSpaceWithImages()
```
donaldgray, line 418: 409 / Conflict is maybe more descriptive than 500?

> The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request. The server SHOULD generate content that includes enough information for a user to recognize the source of the conflict.

src/protagonist/API/Infrastructure/HydraController.cs
JackLewis-digirati: this is removed

donaldgray (approved):

donaldgray:
Resolves #542

Issue #539, opened 2023-07-03
Manage Custom Headers via API


tomcrane:
(Split from https://github.com/dlcs/protagonist/issues/24)

NB These are not currently documented in https://dlcs-book.readthedocs.io/en/latest/API_Reference/introduction.html

- [x] GET `/customers/{customer}/customHeaders` - lists all available custom headers
- [x] POST `/customers/{customer}/customHeaders`
- [x] GET `/customers/{customer}/customHeaders/{customheader}`
- [x] PUT `/customers/{customer}/customHeaders/{customheader}`
- [x] DELETE `/customers/{customer}/customHeaders/{customheader}`

https://github.com/dlcs/protagonist/blob/main/src/protagonist/DLCS.Model/Assets/CustomHeaders/CustomHeader.cs

## Acceptance Criteria

- An API Client can list, create, update and delete custom headers
- `key` and `value` are required, `role` and `space` are optional. role and space are not validated (don't have to exist)
- `role` is a single value not a comma-delimited list (so you need multiple custom headers for multiple roles) 
- POST - the request body does not include an `@id` or `id`, DLCS mints a GUID for `id`; POST is the only way to create a custom header
- POST never produces 409 conflict
- POST that includes an `id` will be 400 Bad Request
- Successful POST => 201 Created with Location header; returns the new resource in the body
- PUT is update-only because DLCS mints the identifier
- PUT allows you to update the `key`, `value`, `role`, `space` fields
- DELETE removes from DB

PR #569, opened 2023-07-31
Allow Custom Headers to be managed via the API

griffri:
Resolves https://github.com/dlcs/protagonist/issues/539


src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            Customer = customerId,
+            Key = "test-key",
+            Value = "test-value"
+        };
+        
+        await dlcsContext.CustomHeaders.AddAsync(customHeader);
+        await dlcsContext.SaveChangesAsync();
+        
+        // Act
+        var response = await httpClient.AsCustomer(customerId).DeleteAsync(Path.Combine(path, customHeader.Id));
```
donaldgray, line 89: I think it's cleaner to have the entire path evident in the `path` variable, rather than building it where `httpClient` is called, that way at a quick scan of the tests it's really evident what URL is being tested.

Initially I thought this was testing `DELETE /customers{customerId}/customHeaders` and didn't initially clock the Id was at the end.

e.g.

```cs
[Fact]
public async Task Delete_CustomHeader_204()
{
    // Arrange
    const int customerId = 92;
    var path = $"customers/{customerId}/customHeaders/{customHeader.Id}`

    // other stuff

    var response = await httpClient.AsCustomer(customerId).DeleteAsync(path);
```

```cs
[Fact]
public async Task Delete_CustomHeader_204()
{
    // Arrange
    const int customerId = 92;
    var path = $"customers/{customerId}/customHeaders";

    // other stuff
    var response = await httpClient.AsCustomer(customerId).DeleteAsync(Path.Combine(path, customHeader.Id));
```

src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        
+        var model = await response.ReadAsHydraResponseAsync<HydraCollection<CustomHeader>>();
+        model.Members.Should().HaveCount(2);
+    }
+    
+    [Fact]
+    public async Task Post_CustomHeader_201()
```
JackLewis-digirati, line 120: Do you need tests around the hydra validations? I.e. testing failure - same with what happens when you try and add a duplicate header?  Additionally, what happens when you try and post to a space which doesn't exist?

src/protagonist/API/Features/CustomHeaders/Requests/DeleteCustomHeader.cs
```diff
@@ -0,0 +1,50 @@
...
+    private readonly DlcsContext dbContext;
+
+    public DeleteCustomHeaderHandler(DlcsContext dbContext)
+    {
+        this.dbContext = dbContext;
+    }
+
+    public async Task<ResultMessage<DeleteResult>> Handle(DeleteCustomHeader request, CancellationToken cancellationToken)
+    {
+        var deleteResult = DeleteResult.NotFound;
```
donaldgray, line 34: `deleteResult` only ever has 2 different values, is it cleaner to use use the enum value directly rather than having a mutating variable? Using a specific enum value at the place it's used saves looking up in the code to see the current value.

E.g., in these it's evident what the return value is:
```cs
if (customHeader == null) return new ResultMessage<DeleteResult>("CustomHeader not found", DeleteResult.NotFound);

return new ResultMessage<DeleteResult>("CustomHeader deleted", DeleteResult.Deleted);
```

whereas here we need to look around to see what the current value of `deleteResult` is
```cs
if (customHeader == null) return new ResultMessage<DeleteResult>(message, deleteResult);

return new ResultMessage<DeleteResult>(message, deleteResult);
```

src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
```diff
@@ -0,0 +1,200 @@
...
+    [Fact]
+    public async Task Put_CustomHeader_404_IfNotFound()
+    {
+        // Arrange
+        const int customerId = 96;
+        var path = $"customers/{customerId}/customHeaders/{Guid.Empty}";
+        const string updatedCustomHeaderJson = @"{
+          ""key"": ""test-key-2"",
+          ""value"": ""test-value-2"",
+          ""space"": 2,
```
JackLewis-digirati, line 190: what happens when you set this value to a space which doesn't exist? Likely also needs verification that nothing changes on failure

src/protagonist/API/Features/CustomHeaders/Converters/CustomHeaderConverter.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace API.Features.CustomHeaders.Converters;
+
+public static class CustomHeaderConverter
```
JackLewis-digirati, line 3: this class needs comments

src/protagonist/API.Tests/Features/CustomHeaders/Validation/HydraCustomHeaderValidator.cs
```diff
@@ -0,0 +1,55 @@
...
+        var result = sut.TestValidate(customHeader);
+        result.ShouldHaveValidationErrorFor(ch => ch.Key);
+    }
+    
+    [Fact]
+    public void NewCustomHeader_Requires_Value()
+    {
+        var customHeader = new CustomHeader();
+        var result = sut.TestValidate(customHeader);
+        result.ShouldHaveValidationErrorFor(ch => ch.Value);
```
JackLewis-digirati, line 53: might be worth validating that the values you haven't set are null in this set of tests?

donaldgray (approved):
Added a couple of comments.

If you change `"Related to"` in comment to one of the GH linking words (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) it'll auto-link the issue and close it when you close this.

src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
griffri: Agreed. This has now been resolved in 17a80f5a6d1196fde444c6c9a291829124344379

src/protagonist/API/Features/CustomHeaders/Requests/DeleteCustomHeader.cs
```diff
@@ -0,0 +1,50 @@
...
+        var message = string.Empty;
```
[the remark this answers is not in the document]
griffri: This should return a message if it cannot find a name query to delete - resolved in 01ce2bfb5832e9ea6641bb84dadc0d8725d65fba

src/protagonist/API/Features/CustomHeaders/Requests/DeleteCustomHeader.cs
griffri: That's a good point, I've made it a little hard to follow. This has been changed in 95770330c2a1a080c057d35aab57ae96ae800260

src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
griffri: https://github.com/dlcs/protagonist/pull/569#discussion_r1280842836

src/protagonist/API/Features/CustomHeaders/Converters/CustomHeaderConverter.cs
griffri: Good call - I missed the fact that the other converters have comments. I'll get to work on that 👍

src/protagonist/API.Tests/Features/CustomHeaders/Validation/HydraCustomHeaderValidator.cs
griffri: Good idea. Added in df51740576117eeabff930b0a854996b3ed7c0aa

src/protagonist/API.Tests/Integration/CustomHeaderTests.cs
JackLewis-digirati: I mainly mean within the controller itself and while the tests themselves are probably not that useful, if you were to output a code coverage report, it would show
```csharp
if (!validationResult.IsValid)
{
    return this.ValidationFailed(validationResult);
}
```
is not covered

In terms of keys, is it worth writing a test to show you can push multiple keys in with the same id? Mainly because if I see key and value, I'd assume it's a dictionary which would mean the keys are unique - having a test like this just makes it cleared that the logic is intentional

And if you're allowing multiple of the same keys, is there a verified order that a key is returned in `GetCustomHeader`? I.e.: is it FIFO or LIFO? and is it worth writing a test to show this?

griffri: Tests for validation have been added in ddc6b699f722ac3d85cc181ef0fa1930f19baf2b

donaldgray: > And if you're allowing multiple of the same keys, is there a verified order that a key is returned in GetCustomHeader? I.e.: is it FIFO or LIFO? and is it worth writing a test to show this?

There is an order of precedence in how the keys are applied in orchestrator but nothing specified about how they should be returned here. Leaving it to DB order is fine for now.

griffri: a2bba53e489d958f5fa4d980f16cf2ad850df51b adds a test for multiple custom headers using the same ID

Issue #335, opened 2022-07-11
Code housekeeping - eliminate build warnings


tomcrane:
While we can eliminate all build warnings, most of them stem from inconsistent `nullable` annotations on arguments passed between components.

PR #574, opened 2023-08-03
Adding ability to delete folders from S3 and invalidate cache

JackLewis-digirati:


griffri (approved):
Looks good!

src/protagonist/DLCS.AWS/Cloudfront/CacheInvalidator.cs
```diff
@@ -0,0 +1,57 @@
...
+        };
+        
+        try
+        {
+            logger.LogDebug("invalidating cloud front. {Paths}", invalidationPaths);
+            var invalidationResult = await client.CreateInvalidationAsync(invalidationRequest, cancellationToken);
+
+            return invalidationResult.HttpStatusCode.IsSuccess();
+        }
+        catch (Exception ex)
```
donaldgray, line 50: I think it's worth catching other exceptions in addition to general `Exception` (https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/CloudFront/MCloudFrontCreateInvalidationCreateInvalidationRequest.html#exceptions)

`NoSuchDistributionException` will never work so it's worth rethrowing.
`TooManyInvalidationsInProgressException` is something we'll be specifically interested i (if it covers the wildcard issue)

src/protagonist/DLCS.AWS/Cloudfront/CacheInvalidator.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            logger.LogError(ex, "Error invalidating the cache for  distribution {Distribution}",
```
donaldgray, line 52: Extra whitespace in `for  distribution`

src/protagonist/DLCS.AWS/Cloudfront/CacheInvalidator.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            DistributionId = cloudfrontSettings.DistributionId,
+            InvalidationBatch = new InvalidationBatch
+            {
+                Paths = new Paths
+                {
+                    Quantity = invalidationPaths.Count,
+                    Items = invalidationPaths
+                },
+                CallerReference = DateTime.Now.Ticks.ToString()
```
donaldgray, line 39: I wonder how long this lasts and whether it's worth putting `assetId` in it somewhere to avoid duplicate requests that are close together. Maybe assetId and sqs messageId or something. (probably not worth doing now, can revisit if we're seeing a lot of duplicates).

src/protagonist/DLCS.AWS/Cloudfront/CacheInvalidator.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Amazon.CloudFront;
+using Amazon.CloudFront.Model;
+using DLCS.AWS.Settings;
+using DLCS.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace DLCS.AWS.Cloudfront;
+
+public class CacheInvalidator : ICacheInvalidator
```
donaldgray, line 10: Small thing, is it worth naming this `CloudfrontInvalidator`, in the future we may support other CDNs (e.g. `FastlyInvalidator : ICacheInvalidator`)

src/protagonist/DLCS.AWS.Tests/SQS/Models/QueueMessageXTests.cs
```diff
@@ -70,4 +75,59 @@ public void GetMessageContents_ReturnsNull_IfSnsOrigin_AndBodyNotJson()
...
+    public void GetMessageContentsWithType_ThrowsException_WhenCalledWithObjectThatCannotBeDeserialized()
+    {
+        // Arrange
+        var queueMessage = new QueueMessage
+        {
+            Body = new JsonObject { ["id"] = "foo" }
+        };
+
+        // Act and Assert
+        Assert.Throws<JsonException>(() =>queueMessage.GetMessageContents<Asset>());
```
donaldgray, line 115: Idiomatic way to do this with FluentAssertions is to use an `Action` and assert that

```cs
Action action = () => queueMessage.GetMessageContents<Asset>();
action.Should().Throw<JsonException>(); // or .ThrowExactly<JsonException>()
```

src/protagonist/DLCS.AWS.Tests/SQS/Models/QueueMessageXTests.cs
```diff
@@ -70,4 +75,59 @@ public void GetMessageContents_ReturnsNull_IfSnsOrigin_AndBodyNotJson()
...
+        var queueMessage = new QueueMessage
+        {
+            Body = new JsonObject { ["id"] = "foo" }
+        };
+
+        // Act
+        var exception = Record.Exception(() => queueMessage.GetMessageContents<Asset>(false));
+        
+        // Assert
+        Assert.Null(exception);
```
donaldgray, line 131: Similar to above, idiomatic way is `Action` and `.Should().NotThrow<Exception>();`. Add assertion for return type `null` too, to document behaviour.

src/protagonist/DLCS.AWS.Tests/SQS/Models/QueueMessageXTests.cs
```diff
@@ -70,4 +75,59 @@ public void GetMessageContents_ReturnsNull_IfSnsOrigin_AndBodyNotJson()
...
+    }
```
donaldgray, line 132: Probably worth putting another test in here to document that this will handle SQS message in the SNS shape as well as raw.

src/protagonist/DLCS.AWS/S3/IStorageKeyGenerator.cs
```diff
@@ -13,6 +13,13 @@ public interface IStorageKeyGenerator
     /// <returns><see cref="ObjectInBucket"/> for tile-ready asset</returns>
     RegionalisedObjectInBucket GetStorageLocation(AssetId assetId);
     
+    /// <summary>
```
donaldgray, line 16:
```
    /// <summary>
    /// Get <see cref="ObjectInBucket"/> for root location of stored delivery assets (tile-ready or original)
    /// </summary>
    /// <param name="assetId">Unique identifier for Asset</param>
    /// <returns><see cref="ObjectInBucket"/> for delivery asset root</returns>
```

src/protagonist/DLCS.AWS/S3/IStorageKeyGenerator.cs
```diff
@@ -146,4 +153,18 @@ public interface IStorageKeyGenerator
     /// </summary>
     /// <returns><see cref="ObjectInBucket"/> for specified asset's metadata file</returns>
     ObjectInBucket GetTimebasedMetadataLocation(AssetId assetId);
+
+    /// <summary>
+    /// Get <see cref="ObjectInBucket"/> root location for the origin bucket
+    /// </summary>
+    /// <param name="assetId">asset id the request is for</param>
+    /// <returns><see cref="ObjectInBucket"/> for specified asset's metadata file</returns>
+    ObjectInBucket GetOriginBucketRoot(AssetId assetId);
```
donaldgray, line 162: `GetOriginRoot(AssetId assetId);` as it's the root for the asset rather than root of the bucket (same with `GetOutputBucketRoot`, below)

src/protagonist/DLCS.AWS/Settings/S3Settings.cs
```diff
@@ -13,29 +13,29 @@ public class S3Settings
     /// <summary>
     /// Name of bucket storing pre-generated thumbnails
     /// </summary>
-    public string ThumbsBucket { get; set; }
+    public string? ThumbsBucket { get; set; }
```
donaldgray, line 16: I think these are better handled as part of #335

src/protagonist/DLCS.Repository/Messaging/AssetNotificationSender.cs
```diff
@@ -99,13 +98,13 @@ public async Task<bool> SendIngestAssetRequest(Asset assetToIngest, Cancellation
...
+        var request = new CleanupAssetNotificationRequest()
         {
-            Id = assetToCleanup.Id.ToString(),
+            Asset = assetToCleanup,
             CustomerPathElement = customerPathElement
         };
-
-        return await topicPublisher.PublishToAssetModifiedTopic(JsonConvert.SerializeObject(notificationRequest, Formatting.None, jsonSerializerSettings), ChangeType.Delete, cancellationToken);
+        
+        return await topicPublisher.PublishToAssetModifiedTopic(JsonConvert.SerializeObject(request, Formatting.None, jsonSerializerSettings), ChangeType.Delete, cancellationToken);
```
donaldgray, line 107: It shouldn't make a difference but as we're using STJ to deserialise these would it make more sense to serialise with STJ too?

donaldgray (approved):
LGTM, approving but added a few comments.

src/protagonist/DLCS.Repository/Messaging/AssetNotificationSender.cs
JackLewis-digirati: I initially used newtonsoft as it's what the rest of the API uses, however git is faster, so I've modified this to use STJ

PR #575, opened 2023-08-04
removing redundant code

JackLewis-digirati:
Removes redundant code accidentally left as part of refactor for #574


donaldgray (approved):
LGTM, maybe expand description for if it's looked back on in a few months?

> Removes redundant code accidentally left as part of refactor for #pr-whatever

PR #579, opened 2023-08-08
Emit IIIF Auth2 services on manifests

### LINKED ISSUE #523: Emit Auth2 services
Emit appropriate Auth2 services on generated Manifests (single-item, NQ and info.json).

donaldgray:
Resolves #523

Update info.json, named-query manifest and single-item manifests to output Auth services. The rules are:

* `info.json` - include both IIIF Auth v1 and v2 services. V1 will be removed in the future but this is to aid rolling out and remove need for lockstep deployments with viewers. Same pattern of main info.json being persisted and `"id"` paths being rewritten on way out is followed.
* NQ - Include IIIF Auth v2 for IIIF Presentation 3 only. IIIF Presentation 2 resources will continue to omit auth services. `AuthAccessService2` is included in main `Manifest.Services` element, relevant `ImageServices` include `AuthProbeService2` and reference the above `AuthAccessService2`.
* Single-item manifests - as NQ's above.

Rough steps to achieve this:
* Rename `IIIFAuthBuilder` to `IIIFAuth1Builder` (this reads Auth services from DLCS database) and extract `IIIIFAuthBuilder` interface from this.
* Add `IIIFAuth2Client` implementation of above interface, which uses [iiif-auth-v2](https://github.com/dlcs/iiif-auth-v2) to generate services.
* Introduced new class, `IIIFManifestBuilder`, which is used by single-item manifest + NQ to generate manifests.
  * This could make a lot of requests to iiif-auth-v2 service, for now I've left the implementation as doing a bulk fan out but may need to batch requests to avoid a self-imposed DDoS to iiif-auth-v2

To handle Path rewrites I introduced `TypedPathTemplateOptions`, this is similar to `PathTemplateOptions` but covers cases where simple path replacements aren't possible and we need an extra level, which I have termed 'type'. E.g. For Auth1 (`"AuthPathRules"` below) we can map all possible auth services with a single rule per host but this doesn't work for auth2 (`"Auth2PathRules"` below) hence the need for additional model.

```json
{
  "AuthPathRules": {
    "Default": "`/auth/{customer}/{behaviour}`",
    "Overrides": {
      "my-proxy.com": "/auth/{behaviour}"
    }
  },
  "Auth2PathRules": {
    "Defaults": {
      "AuthProbeService2": "/auth/v2/probe/{assetId}",
      "AuthAccessService2": "/auth/v2/access/{customer}/{accessService}",
      "AuthAccessTokenService2": "/auth/v2/access/{customer}/token",
      "AuthLogoutService2": "/auth/v2/access/{customer}/{accessService}/logout"
    },
    "Overrides": {
      "my-proxy.com": {
        "AuthProbeService2": "/probe/{assetId}",
        "AuthAccessService2": "/access/{customer}/{accessService}",
        "AuthAccessTokenService2": "/access/{customer}/token",
        "AuthLogoutService2": "/access/{customer}/{accessService}/logout"
      }
    }
  },
}
```

> [!NOTE]  
> NQ and single item manifests don't respect auth rewrite rules, even if called via an alternative path. The assumption is NQ's will only for the base of a manifest and single-item mani is for testing only.


src/protagonist/Orchestrator.Tests/Features/IIIF/IIIFAuth1BuilderTests.cs
```diff
@@ -1,254 +1,254 @@
-﻿using System;
```
JackLewis-digirati, line 1: I think this file has different line endings - could just be the filename has changed though

src/protagonist/Orchestrator/Infrastructure/IIIF/IIIFAuth1Builder.cs
```diff
@@ -1,105 +1,106 @@
...
+using AuthCookie0 = IIIF.Auth.V0.AuthCookieService;
+using AuthCookie1 = IIIF.Auth.V1.AuthCookieService;
+
+namespace Orchestrator.Infrastructure.IIIF;
+
+/// <summary>
+/// Implementation of <see cref="IIIIFAuthBuilder"/> for IIIF Auth 0.9 and 1.0
+/// </summary>
+/// <remarks>This is legacy and will be retired in the future</remarks>
+public class IIIFAuth1Builder : IIIIFAuthBuilder
```
JackLewis-digirati, line 22: everywhere else seems to use `Auth` for v0.9/1 and `Auth2` for v2, whereas it seems there's `Auth1` for v1 - is it worth doing this here to keep consistency?

src/protagonist/Orchestrator/Infrastructure/IIIF/IIIFManifestBuilder.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        return manifest;
+    }
+
+    private async Task<Dictionary<AssetId, AuthProbeService2>?> GetProbeServices(IReadOnlyCollection<Asset> assets, CancellationToken cancellationToken)
+    {
+        var assetsRequiringAuth = assets.Where(a => a.RequiresAuth && !string.IsNullOrEmpty(a.Roles)).ToList();
+
+        var assetsRequiringAuthCount = assetsRequiringAuth.Count;
+        if (assetsRequiringAuthCount == 0) return null;
```
JackLewis-digirati, line 121: what's the reason for returning null over an empty dictionary?

JackLewis-digirati (approved):
LGTM

src/protagonist/Orchestrator.Tests/Features/IIIF/IIIFAuth1BuilderTests.cs
donaldgray: Yeah it looks like it could have but when I've changed it locally and add to git index it's always showing as a noop.

src/protagonist/Orchestrator/Infrastructure/IIIF/IIIFManifestBuilder.cs
donaldgray: The vast majority of assets won't have any items requiring auth (ie `assetsRequiringAuthCount == 0`) so it avoids instantiating a new object in these cases, only for it to be disposed of.

griffri (approved):

src/protagonist/Orchestrator/Infrastructure/IIIF/IIIFAuth1Builder.cs
JackLewis-digirati: no, this makes sense now - thanks for letting me know!

tomcrane (changes requested):
ImageV2 info.json

 - The `@context` declaration needs to pull in the Auth2 context - see https://iiif-auth2-server.herokuapp.com/img/01_Icarus_Breughel.jpg/info.json (technically we should have an auth1 context in there too for the auth1 code block but we never did before)
 - The `label` and `description` for the auth v1 services which are just empty strings atm - is that just because the schema hasn't been populated? If so doesn't matter as we'll be removing that later
- comparing with a live example https://iiif.wellcomecollection.org/image/b33058258_0001.jp2/info.json where we have AuthCookieService0, AuthLogoutService0 and AuthTokenService1, they are all AuthXxxx1 - but I'm fine with that, an auth 1 client will be using the `profile` anyway to identify them.
- BUT - the profiles are also all ../1/... on this version. I don't know whether the work page will pick this up. But if it doesn't it should be a simple change and probably better to be consistent here (although again, it will all go soon).

ImageV3 info.json

- I should have been clearer in Slack... I don't think we need authV1 emitted here, just authV2 - Wellcome manifests only advertise imageV2 so imageV3 API for Wellcome is unadvertised/unsupported, and no new usage of authV1 will be encouraged
- otherwise same as v2 -  need the auth2 `@context`

PresentationV2 manifest

 - LGTM, no auth

PresentationV3 manifest 

 - LGTM - does have auth context and emits only auth2 refs

NQs - same as above

This might not be possible yet because we only have clickthrough but we'll need an example of a NQ that has multiple auth services - e.g., 3 clickthrough and 4 something else - to validate that we get two entries in the `services` and they are correctly referenced on the individual assets.

We'll also need to see, in the single asset and NQ manifests, auth services on non-image-services - AV and files.

donaldgray:
> The `label` and `description` for the auth v1 services which are just empty strings atm - is that just because the schema hasn't been populated? If so doesn't matter as we'll be removing that later

Yes - these were minimally populated in DB, I'll updated to show fully populated example.

> comparing with a live example https://iiif.wellcomecollection.org/image/b33058258_0001.jp2/info.json where we have AuthCookieService0, AuthLogoutService0 and AuthTokenService1, they are all AuthXxxx1 - but I'm fine with that, an auth 1 client will be using the `profile` anyway to identify them.

Whether services are `AuthCookieService0` or `AuthCookieService1` etc is determined by the `profiile` in database - I will create a new example which shows v0 + v1 examples.

> This might not be possible yet because we only have clickthrough but we'll need an example of a NQ that has multiple auth services - e.g., 3 clickthrough and 4 something else - to validate that we get two entries in the `services` and they are correctly referenced on the individual assets.

I should be able to fake this by creating 2 'clickthrough' roles in iiif-auth-v2 service with different access-services. They'll effectively be the same but would offer different Ids

> We'll also need to see, in the single asset and NQ manifests, auth services on non-image-services - AV and files.

Will create a new ticket to cover this, changes in this PR to introduce `IIIFManifestBuilder` will simplify this.

donaldgray:
@tomcrane

Updated `label` and `description` for auth V1 services, was missing data in DB:
* https://dlcs.digirati.io/iiif-img/v2/7/4/523_singleauth/info.json

`@context` added to v2 + v3
* https://dlcs.digirati.io/iiif-img/v2/7/4/523_auth_v0/info.json - v2 + v0
* https://dlcs.digirati.io/iiif-img/v2/7/4/523_singleauth/info.json - v2 + v1
* https://dlcs.digirati.io/iiif-img/7/4/523_singleauth/info.json - v2

Re: AuthCookieService0 or AuthCookieService1 etc, this is DB config driven:
* https://dlcs.digirati.io/iiif-img/v2/7/4/523_singleauth/info.json - V1
* https://dlcs.digirati.io/iiif-img/v2/7/4/523_auth_v0/info.json - V0

NQ with multiple different access-services:
* https://dlcs.digirati.io/iiif-resource/7/slf/523/4

tomcrane (approved):

Issue #548, opened 2023-07-04
Add ProbeService to Orchestrator


donaldgray:
Add probe-service endpoint to Orchestrator. This will delegate the call to the internal probe-service (#547) and return that result. 

The Orchestrator will only call the probe-service if authentication is required an a Bearer token has been provided.

PR #588, opened 2023-08-11
Refactor InfoJsonConstructor to have separate impls for versions

donaldgray:
Related to #548.

This was originally written as part of PR #587 but felt cleaner to leave in a separate PR. Rather than a single `InfoJsonConstructor`, this refactors that out to use template-pattern and introduce a `InfoJsonConstructorTemplate<T>`, this has all shared logic in it and implementing classes (`InfoJson2Constructor` and `InfoJson3Constructor`) only implement relevant logic.

The original caller of `InfoJsonConstructor` now calls a delegate resolver/factory to get appropriate implementation depending on ImageApi version.


src/protagonist/Orchestrator/Features/Images/ImageServer/InfoJson2Constructor.cs
```diff
@@ -0,0 +1,84 @@
...
+using IIIFAuth2 = IIIF.Auth.V2;
+using Version = IIIF.ImageApi.Version;
+
+namespace Orchestrator.Features.Images.ImageServer;
+
+/// <summary>
+/// Service responsible for orchestrating image, calling IImageServerClient to get info.json and update with
+/// required information that image-server will be unaware of (e.g. Auth, Id).
+/// </summary>
+public class InfoJson2Constructor : InfoJsonConstructorTemplate<ImageService2>
```
JackLewis-digirati, line 19: is it worth adding to this class description what the difference is between this, `InfoJsonConstructor` and `InfoJson3Constructor` - assuming the difference is with the `AuthBuilder`

JackLewis-digirati (approved):

src/protagonist/Orchestrator/Features/Images/ImageServer/InfoJson2Constructor.cs
donaldgray: Done, this was copy/paste from original `InfoJsonConstructor`

griffri (approved):

PR #591, opened 2023-08-14
Allowing id to be set to a GUID when empty on POST to a queue

### LINKED ISSUE #558: Mint GUID `id` for assets where no `id` supplied
This applies to new assets POSTed to the queue.

This matches Deliverator behaviour.

## Acceptance criteria

- [ ] If the caller supplies the `id` (not `@id`) in the Asset in a HydraCollection POSTed to the queue, the DLCS should use that value as the `Asset` part of https://github.com/dlcs/protagonist/blob/main/src/protagonist/DLCS.Core/Types/AssetId.cs
- [ ] If the caller does not supply `id` the DLCS will mint a GUID for the Asset part using `Guid.NewGuid()`

JackLewis-digirati:
Resolves #558 

This PR allows the API to mint GUIDS for an `id` if no `id` is set in a queue request
i.e.:
```
{
  "members": [
    {
      "id": "foo", // has id
      "space": 213,
      "origin": "https://images.com/1.jpg"
    },
    {
      "space": 213, // No-id, create GUID
      "origin": "https://images.com/2.jpg"
    }
  ]
}
```

will produce asset id's of `/{customerId}/213/foo` and `/{customerId}/213/{guid}`

*NOTE:* I did leave the hydra validator in for the id as it covers a couple extra scenarios (such as an asset id of " "), but I can remove this if required

### DIFF
--- a/src/protagonist/API/Features/Queues/CustomerQueueController.cs
+++ b/src/protagonist/API/Features/Queues/CustomerQueueController.cs
@@ -1,10 +1,10 @@
-﻿using API.Converters;
+﻿using System.Collections.Generic;
+using API.Converters;
 using API.Features.Queues.Converters;
 using API.Features.Queues.Requests;
 using API.Features.Queues.Validation;
 using API.Infrastructure;
 using API.Settings;
-using DLCS.Core.Collections;
 using DLCS.Core.Strings;
 using DLCS.Model.Assets;
 using DLCS.Model.Processing;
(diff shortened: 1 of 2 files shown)


src/protagonist/API/Features/Queues/CustomerQueueController.cs
```diff
@@ -106,13 +107,24 @@ public async Task<IActionResult> GetCustomerQueue([FromRoute] int customerId, Ca
                 }
             }
         }
-        
+
+        if (images.Members != null)
+        {
+            // ReSharper disable once ReturnValueOfPureMethodIsNotUsed
+            images.Members.Select(c =>
```
donaldgray, line 114: `.Select()` with side effects seems odd, is it better to filter those with `string.IsNullOrEmpty(id)` and only update those?

donaldgray (changes requested):
This same logic needs to be applied to the priority queue too (`/customers/{customerId}/queue/priority`).

donaldgray (approved):

PR #592, opened 2023-08-15
Modifying customer storage to take into account the size of the previous asset

### LINKED ISSUE #386: CustomerStorage management for reingesting asset
When an asset is reingested we need to validate that the maximum file size allowed for that customer isn't exceeded.

The logic needs to change

```cs
// Current logic
var canSave = customerStorage.TotalSizeOfStoredImages + proposedNewFileSize <= policy.MaximumTotalSizeOfStoredImages;

// Needs to be something like
var canSave = (customerStorage.TotalSizeOfStoredImages - oldFileSize) + proposedNewFileSize <= policy.MaximumTotalSizeOfStoredImages;
```

The `oldFileSize` can be read from an existing `ImageStorage` record.

JackLewis-digirati:
Resolves #386

This PR updates a check on how much space a customer has to take into account the size of a reingested asset so that a customer can more accurately have an image uploaded if they are close to their storage limit


src/protagonist/Engine.Tests/Ingest/IngestExecutorTests.cs
```diff
@@ -18,7 +18,8 @@ public IngestExecutorTests()
     {
         workerBuilder = A.Fake<IWorkerBuilder>();
         repo = A.Fake<IEngineAssetRepository>();
-        sut = new IngestExecutor(workerBuilder, repo, new NullLogger<IngestExecutor>());
+        var assetSizeCheck = A.Fake<IAssetIngestorSizeCheck>();
+        sut = new IngestExecutor(workerBuilder, repo, assetSizeCheck ,new NullLogger<IngestExecutor>());
```
donaldgray, line 22: Should there be a test that uses `assetSizeCheck`?

Tiny thing but spacing around comma when passing it is off : `assetSizeCheck ,new Null...`

griffri (approved):

src/protagonist/Engine.Tests/Integration/ImageIngestTests.cs
```diff
@@ -122,6 +122,60 @@ public async Task IngestAsset_Success_HttpOrigin_AllOpen()
...
+        // Note - API will have set this up before handing off
+        var origin = $"{apiStub.Address}/image";
+
+        var entity = await dbContext.Images.AddTestAsset(assetId, ingesting: true, origin: origin,
+            imageOptimisationPolicy: "fast-higher", mediaType: "image/tiff", width: 0, height: 0, duration: 0,
+            deliveryChannels: imageDeliveryChannels);
+        var asset = entity.Entity;
+        await dbContext.Customers.AddTestCustomer(customerId);
+        await dbContext.Spaces.AddTestSpace(customerId, 2);
+        dbContext.ImageStorages.Add(new ImageStorage()
```
donaldgray, line 142: Worth adding an `AddTestImageStorage()` helper?

src/protagonist/Engine.Tests/Integration/ImageIngestTests.cs
```diff
@@ -269,6 +323,59 @@ public async Task IngestAsset_Error_ExceedAllowance()
         storage.Should().BeNull();
     }
     
+    [Fact]
+    public async Task IngestAsset_Error_ExceedAllowanceOnReingest()
```
donaldgray, line 327: What causes this to fail when the other new test passes?

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -85,6 +85,15 @@ public EngineAssetRepository(DlcsContext dlcsContext, ILogger<EngineAssetReposit
     public ValueTask<Asset?> GetAsset(AssetId assetId, CancellationToken cancellationToken = default)
         => dlcsContext.Images.FindAsync(new object[] { assetId }, cancellationToken);
 
+    public async Task<long?> GetImageSize(AssetId assetId, CancellationToken cancellationToken = default)
+    {
+        var imageSize = await dlcsContext.ImageStorages.AsNoTracking().FirstOrDefaultAsync(
```
donaldgray, line 90: As only `Size` is used there's no need to select the entire object (not a huge perf saving in this instance as it's a small table).

```cs
var imageSize = await dlcsContext.ImageStorages
  .Where(i => i.Id == assetId)
  .Select(i => i.Size)
  .FirstOrDefaultAsync();
```

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -27,6 +32,14 @@ public class IngestExecutor
         var workers = workerBuilder.GetWorkers(asset);
         
         var context = new IngestionContext(asset);
+        long? preIngestionAssetSize = null;
```
donaldgray, line 35: Declaring the var outside isn't necessary as it's only used if the `if` statement is `true`

```cs
if (!assetIngestorSizeCheck.CustomerHasNoStorageCheck(asset.Customer))
{
    var preIngestionAssetSize = await assetRepository.GetImageSize(asset.Id, cancellationToken);
    context.WithPreIngestionAssetSize(preIngestionAssetSize);
}
```

src/protagonist/Engine/Ingest/Persistence/AssetMoverBase.cs
```diff
@@ -15,10 +15,10 @@ protected AssetMoverBase(IStorageRepository storageRepository)
         StorageRepository = storageRepository;
     }
     
-    protected async Task<bool> VerifyFileSize(AssetId assetId, long size)
+    protected async Task<bool> VerifyFileSize(AssetId assetId, long size, long oldFileSize)
     {
         var storageMetrics = await StorageRepository.GetStorageMetrics(assetId.Customer, CancellationToken.None);
-        var customerHasEnoughSize = storageMetrics.CanStoreAssetSize(size);
+        var customerHasEnoughSize = storageMetrics.CanStoreAssetSize(size, oldFileSize );
```
donaldgray, line 21: Additional whitespace after parameter `, oldFileSize );`

donaldgray (approved):
LGTM, couple of small comments

src/protagonist/Engine.Tests/Integration/ImageIngestTests.cs
JackLewis-digirati: it's to do with the size of the previous asset in the db - on success it's 950, with an overall storage of 950, a max of 1000 and a 'new' asset size of 1000, meaning the logic is 950 - 950 + 1000 <= 1000 

in the failure test it's 500 with an overall storage of 950, so it produces 950 - 500 + 1000 <= 1000

src/protagonist/Engine.Tests/Integration/ImageIngestTests.cs
JackLewis-digirati: The `imageStorage` passed in does change between the tests, so I'm not sure this is neccessary

PR #597, opened 2023-08-30
Add PUT endpoint to Spaces API

### LINKED ISSUE #562: PUT can create a space as well as update it
## Acceptance criteria

 - [ ] You can create a space for any valid integer `id` (as long as not already taken)

e.g., 

```
PUT {{dlcs-api-base}}customers/{{customer_id}}/spaces/333
{
    "name":"Space 333 created"
}
```

griffri:
Resolves #562 

Additions:
- `PUT` endpoint for `SpaceController.cs` + tests
- `PutSpace` request that can create and update an existing space

Changes:
- `DlcsDatabaseFixture.CleanUp()` now includes queues when clearing old test data
- `GetSpace` and `PatchSpace` have been rewritten to use the the `FetchEntityResult` and `ModifyEntityResult` helper classes, respectively

Notes:
- It might be worth reviewing how `CreateSpace` and `GetPageOfSpaces` are implemented and possibly update them so that they match current DLCS practices, e.g

  - Both `CreateSpace` and `GetPageOfSpaces` validate the calling user before performing the request, but this behaviour isn't seen in other requests
https://github.com/dlcs/protagonist/blob/47bfbf0455d1a60fa76b119758751ef39b320315/src/protagonist/API/Features/Space/Requests/CreateSpace.cs#L57 https://github.com/dlcs/protagonist/blob/39abf53f6cf369caba4501ca79d81c5769736fca/src/protagonist/API/Features/Space/Requests/GetPageOfSpaces.cs#L45
  - The `CreateSpace` route in `SpaceController` has a check that ensures that the created space has been assigned an ID. Is this still necessary?
https://github.com/dlcs/protagonist/blob/39abf53f6cf369caba4501ca79d81c5769736fca/src/protagonist/API/Features/Space/SpaceController.cs#L103

### DIFF
--- a/src/protagonist/API/Features/Space/Requests/PutSpace.cs
+++ b/src/protagonist/API/Features/Space/Requests/PutSpace.cs
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Generic;
+using API.Infrastructure.Requests;
+using DLCS.Core;
 using DLCS.Core.Strings;
 using DLCS.Model.Spaces;
 using MediatR;
@@ -8,7 +10,7 @@ namespace API.Features.Space.Requests;
 /// <summary>
 /// Create or update an existing space
 /// </summary>
-public class PutSpace : IRequest<PutSpaceResult>
+public class PutSpace : IRequest<ModifyEntityResult<DLCS.Model.Spaces.Space>>
 {
     public int CustomerId { get; set; }
     public int SpaceId { get; set; }
@@ -19,14 +21,7 @@ public class PutSpace : IRequest<PutSpaceResult>
     public int? MaxUnauthorised { get; set; }
 }
 
-public class PutSpaceResult
-{
-    public DLCS.Model.Spaces.Space? Space { get; set; }
-    public List<string> ErrorMessages { get; } = new();
-    public bool Conflict { get; set; }
-}
-
-public class PutSpaceHandler : IRequestHandler<PutSpace, PutSpaceResult>
+public class PutSpaceHandler : IRequestHandler<PutSpace, ModifyEntityResult<DLCS.Model.Spaces.Space>>
 {
     private readonly ISpaceRepository spaceRepository;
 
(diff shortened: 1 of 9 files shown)


src/protagonist/API/Features/Space/SpaceController.cs
```diff
@@ -140,27 +138,31 @@ public async Task<IActionResult> DeleteSpace(int customerId, int spaceId)
...
+            request,
+            s => s.ToHydra(GetUrlRoots().BaseUrl),
+            errorTitle: "Failed to get space",
+            cancellationToken: cancellationToken
+        );
     }
 
     /// <summary>
-    /// Make a partial update of an existing space
+    /// Patch a space within this customer. DLCS assigns identity.
```
donaldgray, line 157: The previous xml docs were more accurate, `DLCS assigns identify` isn't relevant here as this is a partial update to an existing space so we'd be unaware how the id was assigned.

src/protagonist/API/Features/Space/SpaceController.cs
```diff
@@ -140,27 +138,31 @@ public async Task<IActionResult> DeleteSpace(int customerId, int spaceId)
...
     /// </summary>
     /// <remarks>
     /// Sample request:
     ///
-    ///     PATCH: /customers/1/spaces/5
+    ///     POST: /customers/1/spaces
```
donaldgray, line 162: This uses `PATCH` verb

src/protagonist/API/Features/Space/Requests/PutSpace.cs
```diff
@@ -0,0 +1,55 @@
...
+
+/// <summary>
+/// Create or update an existing space
+/// </summary>
+public class PutSpace : IRequest<ModifyEntityResult<DLCS.Model.Spaces.Space>>
+{
+    public int CustomerId { get; set; }
+    public int SpaceId { get; set; }
+    public string? Name { get; set; }
+    public string? ImageBucket { get; set; } = string.Empty;
```
donaldgray, line 18: Should this be `string` if it's assigned a default value?

src/protagonist/API/Features/Space/Requests/PutSpace.cs
```diff
@@ -0,0 +1,55 @@
...
+                return ModifyEntityResult<DLCS.Model.Spaces.Space>.Failure($"The space name '{request.Name}' is already taken.",
+                    WriteResult.Conflict);
+        }
+        
+        var putSpaceResult = await spaceRepository.PutSpace(
+            request.CustomerId, request.SpaceId, request.Name, request.ImageBucket,
+            request.MaxUnauthorised, request.Tags, request.Roles,
+            cancellationToken);
+       
+        return ModifyEntityResult<DLCS.Model.Spaces.Space>.Success(putSpaceResult);
```
donaldgray, line 53: This should also return `WriteResult.Created` or `WriteResult.Updated`, depending on what happened.

This enum value will be used by the `HydraController` to return a 200 or 201 w/ Location header

donaldgray (changes requested):
LGTM, only real change I can see required is to get the 200/201 response for PUT.

Based on slack chat, the if statements should include braces.

It's good to move some of these methods to use `FetchEntityResult` + `ModifyEntityResult`

donaldgray:
>The CreateSpace route in SpaceController has a check that ensures that the created space has been assigned an ID. Is this still necessary?

I can't see why that would be necessary

donaldgray:
>`int? customerId = request.CustomerId ?? principal.GetCustomerId(); `

This can go, `request.CustomerId` is always present. I wonder if this is from API + Portal sharing the same mediatr request at some point.

JackLewis-digirati (approved):

src/protagonist/API/Features/Space/SpaceController.cs
griffri: I'm not sure why I edited it - I think it might've been by mistake while I was working on the comments for PUT. Thanks for spotting it!

donaldgray (approved):
LGTM

PR #606, opened 2023-09-11
Return correct `@id` value in `GET /imageOptimisationPolicies`

### LINKED ISSUE #525: GET `/imageOptimisationPolicies` returning wrong `@id` value for global policies
GET `/imageOptimisationPolicies` returns a list of imageOptimisationPolicies. ImageOptimisationPolicies can be customer specific, or global. If global they are assigned `customer=1` (Admin customer). 

This is rendering an `@id` property of `"https://api.dlcs.io/customers/1/imageOptimisationPolicies/audio-max"` rather than `"https://api.dlcs.io/imageOptimisationPolicies/audio-max"`. The former doesn't work as the API expects credentials to be provided when viewing a `/customer/{customerId}/` path, however the latter does.

We need to render the `@id` property _without_ customer if it's global but _with_ customer if it is customer-specific.

I suspect this is to do with the Hydra modelling and how it generates uri's.

griffri:
Fixes #525

This changes the `ImageOptimisationPolicy` Hydra model so that global polices do not contain the customer's ID in the `@id` path.

### DIFF
--- a/src/protagonist/API.Tests/Integration/PolicyTests.cs
+++ b/src/protagonist/API.Tests/Integration/PolicyTests.cs
@@ -1,5 +1,9 @@
-﻿using System.Net;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
 using System.Net.Http;
+using System.Text.Json;
 using System.Threading.Tasks;
 using API.Client;
 using API.Tests.Integration.Infrastructure;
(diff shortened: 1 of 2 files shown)


donaldgray (approved):
LGTM - is it worth adding a test to verify the fix?

griffri:
> LGTM - is it worth adding a test to verify the fix?

This might not be such a bad idea seeing as this changes a little bit of the default behaviour of the Hydra models. I'll add one soon.

JackLewis-digirati (approved):

donaldgray (approved):

PR #644, opened 2023-10-11
Allow users to view thumbnails in the Portal image viewer

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/114
Resolves https://github.com/dlcs/private-protagonist/issues/116

This PR adds a new panel for previewing an image's thumbnails:

thumbnailviewer

Additionally, it adds the option to view the single asset manifest of an image: 

image


src/protagonist/DLCS.Core/Settings/DlcsSettings.cs
```diff
@@ -28,6 +28,11 @@ public class DlcsSettings
     /// </summary>
     public string SpaceManifestQuery { get; set; }
     
+    /// <summary>
+    /// URL format for generating manifests for single assets
+    /// </summary>
+    public string SingleAssetManifestTemplate { get; set; }
```
donaldgray, line 34: This value will be unlikely to change between deployments so should have a default value set in `appSettings.json` to remove the need to set it whenever it's deployed.

griffri: Good point, changed in [0c3a4e0](https://github.com/dlcs/protagonist/pull/644/commits/0c3a4e09c0bb34121a6616d968002d04bc5610b7)

PR #646, opened 2023-10-16
Allow queues to be viewed in Portal

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/117

This PR adds pages for viewing the "recent" and "active" batch queues:

image
image


src/protagonist/Portal/Pages/Queue/Index.cshtml.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        this.dlcsClient = dlcsClient;
+    }
+
+    public async Task OnGetAsync(
+        [FromRoute] string queueType, 
+        [FromQuery] int page = 1, 
+        [FromQuery] int pageSize = PagerViewComponent.DefaultPageSize)
+    {
+        Queue = await dlcsClient.GetQueue();
```
donaldgray, line 32: Values from `Queue` is only used `@if (Model.QueueType == IndexModel.BatchQueueType.Active)` so only needs to be called if it's active queue.

src/protagonist/Portal/Pages/Queue/Index.cshtml.cs
```diff
@@ -0,0 +1,54 @@
...
+        
+        switch (queueType)
```
donaldgray, line 34: `switch` seems unnecessary here if there's only 2 possible values?

donaldgray (changes requested):
The `IndexModel` should make a mediator request to gather the required values to render the view.

It may seem overkill but we've got a couple of mediatr [Behaviours](https://github.com/jbogard/MediatR/wiki/Behaviors) that implement cross cutting concerns to log requests + create an audit log of which user carried out each action. See https://github.com/dlcs/protagonist/blob/72b2d2e83e254ae4b354b45d8ec2051aa563b064/src/protagonist/Portal/Startup.cs#L83-L84

src/protagonist/Portal/Pages/Queue/Index.cshtml
```diff
@@ -0,0 +1,88 @@
...
+            </tr>
+        </thead>
+        <tbody>
+            @foreach (var batch in Model.Batches.Members.OrderByDescending(batch => batch.Id))
+            {
+                var percentComplete = Math.Round(batch.Count > 0 ? (double)batch.Completed / (double)batch.Count * 100.0 : 1.0);
+                var progressPct = batch.Finished.HasValue ? 100 : percentComplete;
+
+                <tr>
+                    <td>@batch.Id?.Split('/').Last()</td>
```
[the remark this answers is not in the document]
griffri: I wasn't aware of this. Much appreciated!

src/protagonist/Portal/Pages/Queue/Index.cshtml.cs
griffri: Good catch. Will fix.

JackLewis-digirati (approved):

PR #650, opened 2023-10-23
Allow batches to be viewed individually in Portal

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/118

This PR adds a page for viewing individual batches in Portal:

image

### DIFF
--- a/src/protagonist/Portal/Features/Batches/Requests/GetBatch.cs
+++ b/src/protagonist/Portal/Features/Batches/Requests/GetBatch.cs
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 using API.Client;
 using DLCS.Core.Settings;
 using DLCS.HydraModel;
+using DLCS.Web.Auth;
 using Hydra.Collections;
 using MediatR;
 using Microsoft.Extensions.Options;
@@ -28,13 +30,13 @@ public class GetBatchHandler : IRequestHandler<GetBatch, GetBatchResult>
 {
     private readonly DlcsSettings dlcsSettings;
     private readonly IDlcsClient dlcsClient;
-    private readonly HttpClient httpClient;
+    private readonly string customerId;
     
-    public GetBatchHandler(IDlcsClient dlcsClient, HttpClient httpClient, IOptions<DlcsSettings> dlcsSettings)
+    public GetBatchHandler(IDlcsClient dlcsClient, ClaimsPrincipal currentUser, IOptions<DlcsSettings> dlcsSettings)
     {
         this.dlcsClient = dlcsClient;
-        this.httpClient = httpClient;
         this.dlcsSettings = dlcsSettings.Value;
+        customerId = (currentUser.GetCustomerId() ?? -1).ToString();
     }
 
     public async Task<GetBatchResult> Handle(GetBatch request, CancellationToken cancellationToken)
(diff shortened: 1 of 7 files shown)


src/protagonist/Portal/Pages/Batches/Index.cshtml
```diff
@@ -0,0 +1,91 @@
...
+}
+
+<div class="card w-100">
+    <div class="card-body">
+        @if (Model.Batch.Errors > 0)
+        {
+            <div class="row px-2">
+                <div class="alert alert-danger" role="alert">
+                    <i class="me-2" data-feather="alert-circle"></i><strong>Error!</strong>
+                    This batch contains @Model.Batch.Errors image@(Model.Batch.Errors > 1 ? "s" : null) that could not be processed
```
donaldgray, line 20: I think we should stick with the old wording of `has errors` rather than `could not be processed` to avoid confusion. `could not be processed` sounds like the DLCS couldn't generate derivatives but it could be the case that we couldn't fetch the image in the first instance

src/protagonist/Portal/Pages/Batches/Index.cshtml
```diff
@@ -0,0 +1,91 @@
...
+        }
+    </div>
+</div>
+<div class="row m-1 justify-content-center">
+    <div class="row">
+        @foreach (var image in Model.Images.Members)
+        {
+            <div class="p-4 d-flex flex-column" style="width:20%; min-width: 128px;">
+                <a asp-page="/Images/Index" asp-route-space="@image.Space" asp-route-image="@image.ModelId" class="flex-grow-1 align-items-center d-flex mb-1">
+                    @if (!string.IsNullOrEmpty(image.Error) || !Model.Thumbnails.ContainsKey(image.ModelId))
```
donaldgray, line 64: There are a couple of missing cases here that are in deliverator.

```
if 'file' asset
  .. render glyphicon-floppy-disk ..
if 'timebased' asset
  .. render glyphicon-film ..
if hasRoles
  .. render glyphicon-lock ..
```

src/protagonist/Portal/Pages/Batches/Index.cshtml.cs
```diff
@@ -0,0 +1,50 @@
...
+using Portal.Features.Spaces.Requests;
+
+
+namespace Portal.Pages.Batches;
+
+[BindProperties]
+public class Index : PageModel
+{
+    private readonly IMediator mediator;
+    private const int ImageReferenceMaxLength = 32;
```
donaldgray, line 19: Variable isn't used

src/protagonist/Portal/Pages/Batches/Index.cshtml.cs
```diff
@@ -0,0 +1,50 @@
...
+        var batchResult = await mediator.Send(new GetBatch{ BatchId = batch });
+        Batch = batchResult.Batch;
+        Images = batchResult.Images;
+        Thumbnails = batchResult.Thumbnails;
+        return Page();
+    }
+    
+    public string GetImageReference(Image image)
+    {
+        var reference = string.Format("{0}/{1}/{2}/{3}/{4}/{5}",
```
donaldgray, line 40: Elsewhere in the solution string interpolation is generatlly favoured instead of `string.Format()`. 

Null coalescing to `string.Empty` isn't required

donaldgray (changes requested):

src/protagonist/Portal/Features/Batches/Requests/GetBatch.cs
```diff
@@ -0,0 +1,59 @@
...
+    public async Task<GetBatchResult> Handle(GetBatch request, CancellationToken cancellationToken)
+    {
+        var batch = await dlcsClient.GetBatch(request.BatchId);
+        var images = await dlcsClient.GetBatchImages(request.BatchId);
+        var thumbnails = new Dictionary<string, string>();
+        
+        foreach (var image in images.Members)
+        {
+            var thumbnailSrc = new Uri(dlcsSettings.ResourceRoot, 
+                $"thumbs/27/{image.Space}/{image.ModelId}/full/full/0/default.jpg").ToString();
```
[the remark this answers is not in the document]
griffri: Woops! I must have left this in during testing. Good catch!

JackLewis-digirati (approved):

donaldgray (approved):

PR #653, opened 2023-10-27
Allow batches to uploaded via CSV files in Portal

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/108

This PR adds the ability to create new batches of images in the portal with a CSV file:
image

### DIFF
--- a/src/protagonist/Portal/Features/Batches/CsvUploadController.cs
+++ b/src/protagonist/Portal/Features/Batches/CsvUploadController.cs
@@ -9,24 +9,22 @@ using Portal.Features.Batches.Requests;
 
 namespace Portal.Features.Batches;
 
+[Route("[controller]/[action]")]
 public class CsvUploadController : Controller
 {
-    private readonly ClaimsPrincipal currentUser;
     private readonly IMediator mediator;
     private readonly IDlcsClient dlcsClient;
 
-    public CsvUploadController(IMediator mediator, IDlcsClient dlcsClient, ClaimsPrincipal currentUser)
+    public CsvUploadController(IMediator mediator, IDlcsClient dlcsClient)
     {
         this.mediator = mediator;
         this.dlcsClient = dlcsClient;
-        this.currentUser = currentUser;
     }
 
     [HttpPost]
-    [Route("[controller]/{customer}/{space}/[action]")]
-    public async Task<IActionResult> Upload(int customer, int space, List<IFormFile> file)
+    public async Task<IActionResult> Upload(List<IFormFile> file)
     {
-        var request = await mediator.Send(new ParseCsv(){ File = file[0]});
+        var request = await mediator.Send(new IngestFromCsv(){ File = file[0]});
         
         if (!request.IsSuccess)
         {
--- a/src/protagonist/Portal/Features/Batches/CsvUploadController.cs
+++ b/src/protagonist/Portal/Features/Batches/CsvUploadController.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using API.Client;
+using MediatR;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Portal.Features.Batches.Requests;
+
+namespace Portal.Features.Batches;
+
+public class CsvUploadController : Controller
+{
+    private readonly ClaimsPrincipal currentUser;
+    private readonly IMediator mediator;
+    private readonly IDlcsClient dlcsClient;
+
+    public CsvUploadController(IMediator mediator, IDlcsClient dlcsClient, ClaimsPrincipal currentUser)
+    {
+        this.mediator = mediator;
+        this.dlcsClient = dlcsClient;
+        this.currentUser = currentUser;
+    }
+
+    [HttpPost]
+    [Route("[controller]/{customer}/{space}/[action]")]
+    public async Task<IActionResult> Upload(int customer, int space, List<IFormFile> file)
+    {
+        var request = await mediator.Send(new ParseCsv(){ File = file[0]});
+        
+        if (!request.IsSuccess)
+        {
+            return BadRequest(request);
+        }
+        
+        return Ok();
+    }
+}
--- a/src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
+++ b/src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
@@ -81,6 +81,7 @@ public class ParseCsvHandler : IRequestHandler<ParseCsv, ParseCsvResult>
             {
                 try
                 {
+                    
                     var record = csv.GetRecord<ImageIngestModel>();
                     if (record.AssetType.ToLower() != "image") continue;
                     if (distinctRows.ContainsKey(record.Line.Value)) continue;
@@ -117,9 +118,9 @@ public class ParseCsvHandler : IRequestHandler<ParseCsv, ParseCsvResult>
                 }
                 catch (TypeConverterException typeEx)
                 {
+                    var fieldIndex = typeEx.Context.Reader.CurrentIndex - 1;
                     var currentLine = csv.GetField(1);
-                    var fieldName = ImageIngestModel.FieldNames[typeEx.Context.Reader.CurrentIndex];
-                    readErrors.Add($"Line {currentLine}: could not parse {fieldName} value '{typeEx.Text}'");
+                    readErrors.Add($"Line {currentLine}: could not parse {ImageIngestModel.FieldNames[fieldIndex]} value ({typeEx.Text})");
                 }
                 catch (Exception ex)
                 {
--- a/src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
+++ b/src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
@@ -25,7 +25,6 @@ namespace Portal.Features.Batches.Requests;
 
 public class ParseCsv : IRequest<ParseCsvResult>
 {
-    public int? SpaceId { get; set; }
     public IFormFile File { get; set; }
 }
 
@@ -184,7 +183,7 @@ public class ImageIngestModel
     [Index(2)] public int Space { get; set; }
     [Index(3)] public string Id { get; set; }
     [Index(4)] public string Origin { get; set; }
-    [Index(5)] public string InitialOrigin { get; set; }
+    [Index(5), NullValues("")] public string? InitialOrigin { get; set; }
     [Index(6)] public string String1 { get; set; }
     [Index(7)] public string String2 { get; set; }
     [Index(8)] public string String3 { get; set; }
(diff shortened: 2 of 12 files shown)


src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        var distinctRows = new Dictionary<int, Image>();
+        var readErrors = new List<string>();
+        
+        using (var stream = new StreamReader(request.File.OpenReadStream()))
+        using (var csv = new CsvReader(stream, CultureInfo.InvariantCulture))
+        {
+            csv.Read();
+            csv.ReadHeader();
+            while (csv.Read())
```
donaldgray, line 78: Rider is showing that there's an async overload of `csv.Read()` that is probably worth using, particularly for larger files

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+                        Number3 = record.Number3,
+                        InitialOrigin = record.InitialOrigin,
+                        Family = AssetFamily.Image,
+                        MediaType = "image/jp2"
+                    });
+                }
+                catch (BadDataException badDataEx)
+                {
+                    return ParseCsvResult.Failure(
+                        $"CSV read error: bad data found in file at line {badDataEx.Context.Parser.Row}, row {badDataEx.Context.Reader.CurrentIndex}");
```
donaldgray, line 110: These error messages all mention `line` but I think this is row index - rather than `Line` value specified in the spreadsheet. It might be worth using different wording to differentiate? e.g. `bad data found in file at row `

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+                catch (MissingFieldException missingFieldEx)
+                {
+                    return ParseCsvResult.Failure(
+                        $"CSV read error: line {missingFieldEx.Context.Parser.Row} in file is missing fields");
+                }
+                catch (TypeConverterException typeEx)
+                {
+                    var fieldIndex = typeEx.Context.Reader.CurrentIndex - 1;
+                    var currentLine = csv.GetField(1);
+                    readErrors.Add($"Line {currentLine}: could not parse {ImageIngestModel.FieldNames[fieldIndex]} value ({typeEx.Text})");
```
donaldgray, line 121: Didn't notice this when I wrote above comment, this _does_ mean `Line` value so I think it's worth using different wording in the other errors.

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+            try
+            {
+                await dlcsClient.CreateBatch(collection);
+            }
+            catch (DlcsException dlcsEx) // Forward errors from the API
+            {
+                // Get the range of images where the exception may have occured:
+                var imagesStart = (maxBatchSize * i) + 1;
+                var imagesEnd = (maxBatchSize * i) + batches[i].Length;
+                return ParseCsvResult.Failure($"DLCS Error in images {imagesStart}-{imagesEnd}: {dlcsEx.Message}");
```
donaldgray, line 160: Similar to above messaging, this looks like `Lines` rather than "images"

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+using CsvHelper.TypeConversion;
+using DLCS.Web.Auth;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Portal.Settings;
+using MissingFieldException = CsvHelper.MissingFieldException;
+
+namespace Portal.Features.Batches.Requests;
+
+public class ParseCsv : IRequest<ParseCsvResult>
```
donaldgray, line 26: Better name here? `ParseCsv` is only part of what this does

src/protagonist/Portal/Portal.csproj
```diff
@@ -60,4 +61,8 @@
       <Folder Include="wwwroot\img" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Compile Remove="Features\Batches\Requests\PostBatch.cs" />
```
donaldgray, line 65: What's this for?

src/protagonist/Portal/Pages/Queue/Upload.cshtml
```diff
@@ -0,0 +1,58 @@
...
+
+@section header
+{
+    <link href="https://cdn.jsdelivr.net/npm/dropzone@5.9.2/dist/min/dropzone.min.css" rel="stylesheet" />
+    <link href="~/assets/dist/css/jquery-ui.min.css" rel="stylesheet" />
+    <link href="~/assets/dist/css/jquery-ui.structure.min.css" rel="stylesheet" />
+}
+
+<div>On this page you can upload a CSV of information about images for the DLCS to ingest.</div>
+<a href="https://legacyportal.dlcs.digirati.io/Support/sample.csv">Click here to download a sample .csv file.</a>
```
donaldgray, line 16: This shouldn't reference the old portal as it's a dev environment and will be removed once we roll out Protagonist.

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+        
+        return ParseCsvResult.Success();
+    }
+}
+
+public class ImageIngestModel
+{
+    public static readonly string[] FieldNames =
```
donaldgray, line 175: These `FieldNames` don't line up with `ImageIngestModel` - should they? There are 14 fields but 15 properties on model

src/protagonist/Portal/Pages/Queue/Upload.cshtml
```diff
@@ -0,0 +1,58 @@
...
+<div class="alert alert-danger my-2" id="upload-error-box" role="alert" hidden>
+    <i class="me-2" data-feather="alert-circle"></i><strong>Error!</strong>
+    <ul class="my-1" id="errors">
+    </ul>
+</div>
+<div class="my-2">
+    <form action="@Url.Action("Upload", "CsvUpload", new { customer = 27, space = 1 })" class="dropzone" id="dropzoneForm"></form>
```
donaldgray, line 23: Hardcoded route values

src/protagonist/Portal/Features/Batches/CsvUploadController.cs
```diff
@@ -0,0 +1,38 @@
...
+        this.mediator = mediator;
+        this.dlcsClient = dlcsClient;
+        this.currentUser = currentUser;
+    }
+
+    [HttpPost]
+    [Route("[controller]/{customer}/{space}/[action]")]
+    public async Task<IActionResult> Upload(int customer, int space, List<IFormFile> file)
+    {
+        var request = await mediator.Send(new ParseCsv(){ File = file[0]});
```
donaldgray, line 29: In js `maxFiles: 32,` property is set but only `file[0]` is used here. What happens if someone drag/drops multiple files? Is it easier to set `maxFiles: 1`?

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+    
+    public ParseCsvHandler(IDlcsClient dlcsClient, IOptions<PortalSettings> portalSettings, ILogger<ParseCsvHandler> logger, ClaimsPrincipal currentUser)
+    {
+        this.dlcsClient = dlcsClient;
+        this.logger = logger;
+        customerId = (currentUser.GetCustomerId() ?? -1);
+        maxBatchSize = portalSettings.Value.MaxBatchSize;
+    }
+
+    public async Task<ParseCsvResult> Handle(ParseCsv request, CancellationToken cancellationToken)
```
donaldgray, line 68: There's a lot happening in this method. It might be easier to read if it were split down into more meaningful methods - makes it easier to see what it does a glance (ParsesCsv then UploadsToDlcs) and if you need to see details you can look into methods.

```cs
public async Task<ParseCsvResult> Handle(ParseCsv request, CancellationToken cancellationToken)
{
  var (distinctRows, readErrors) = ParseCsv(request.File);

  if (readErrors.Any()) 
  {
    return ParseCsvResult.Failure(readErrors.ToArray());
  }

  var parseCsvResult = await UploadImagesToDlcs(distinctRows);
  return parseCsvResult;
}
```

A note on the last `var parseCsvResult` - returning that from a method that doesn't parse highlights that name is a little out.

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+        
+        // Order images
+        var images = distinctRows
+            .OrderBy(image => image.Key)
+            .Select(image => image.Value)
+            .ToList();
+        
+        // Split the read images into multiple batches if they surpass the maximum batch size
+        var batches = images.Chunk(maxBatchSize).ToList();
```
donaldgray, line 144: Internally `.Chunk()` uses [`yield return`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/yield) to return next element from the iterator. By calling `.ToList()` you are getting all results at once but looking at the code in the `for` statement you might only use the first chunk (e.g. if the first chunk fails then processing is aborted).

You can avoid this by iterating over the `.Chunk()` results, which means you'll only do the work to get the next Chunk of items when they are required. If you abort after chunk 1 of 10, you've only fetched the first rather than all prematurely.

```cs
foreach (var batch in images.Chunk(maxBatchSize))
{
  var collection = new HydraCollection<Image> { Members = batch };
```

donaldgray (changes requested):

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
griffri: Good idea, I'll replace this one with `bad data found in file at row X, column Y`

src/protagonist/Portal/Pages/Queue/Upload.cshtml
griffri: I overlooked this - I'll stick a copy of the file in wwwroot and link to there instead

src/protagonist/Portal/Features/Batches/Requests/ParseCsv.cs
griffri: Good catch! It's missing `InitialOrigin`, will fix asap

src/protagonist/Portal/Pages/Queue/Upload.cshtml
griffri: I forgot about these. They're not used any more either, so I'll just remove them

donaldgray (approved):
LGTM

PR #660, opened 2023-11-02
adding ability to set model id from the @id property

### LINKED ISSUE #654: protagonist as deliverator does not handle the @id property correctly
in deliverator the following payload can be successfully POSTed: 
```
{
  "@type": "Collection",
  "member": [
    {
      "@id": "https:<dlcs api>/customers/<customer id>/spaces/2/images/someId",
      "space" : 2,
      "origin": "https://www.nasa.gov/sites/default/files/thumbnails/image/as11-44-6552.jpeg",
    },
    {
      "@id": "https:<dlcs api>/customers/<customer id>/spaces/2/images/someId2",
      "space" : 2,
      "origin": "https://www.nasa.gov/sites/default/files/thumbnails/image/hap_brennecke_1.jpg",
    },
    {
      "@id": "https:<dlcs api>/customers/<customer id>/spaces/2/images/someId3",
      "space" : 2,
      "origin": "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerBlazes.mp4",
    }
  ]
}
```

However, when POSTing to protagonist (or protagonist as deliverator) we get a `500` error back due to the `@id` property.  In order to support the migration to protagonist, the `@id` property needs to be handled in the same way as deliverator when in the protagonist as deliverator mode

JackLewis-digirati:
Resolves #654

when creating a batch of images, using the `@id` property field makes the API return a `500` error.  When running in deliverator mode, this is incorrect and should allow a user to create a batch using this property to set the id.  

Ultimately, this issue is caused by the `LegacyModeConverter` defaulting a `ModelId` when one isn't supplied. However, a later check also checks that the `@id` field is equivalent to the `ModelId` field and throws an error if it doesn't match.  This fix makes the `LegacyModeConverter` take into account the `@id` field when defaulting the `ModelId` field

### DIFF
--- a/src/protagonist/API.Tests/Converters/LegacyModeConverterTests.cs
+++ b/src/protagonist/API.Tests/Converters/LegacyModeConverterTests.cs
@@ -99,7 +99,7 @@ public class LegacyModeConverterTests
     public void VerifyAndConvertToModernFormat_ModelIdSet_WhenNoModelId()
     {
         // Arrange
-        var hydraImage = new Image{ Id = "https://test/someId" };
+        var hydraImage = new Image{ Id = "https://test/someId", MediaType = "something"};
         
         // Act
         var convertedImage = LegacyModeConverter.VerifyAndConvertToModernFormat(hydraImage);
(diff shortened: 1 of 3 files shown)


JackLewis-digirati:
added an integration test per comments

donaldgray (approved):

PR #661, opened 2023-11-02
Bugfix - Remove items when /deleteImages called

### LINKED ISSUE #656: the /deleteImages endpoint doesn't actually delete images
links to #574 

when using the `/deleteImages` endpoint a request is made to delete but the images aren't actually deleted - this is due to the below:

```
// went from this
var deletedRows = await dlcsContext.Images.AsNoTracking()
  .Where(i => i.Customer == request.CustomerId && assetIds.Contains(i.Id))
  .DeleteFromQueryAsync(cancellationToken);

// to this
var assetsFromDatabase = dlcsContext.Images
   .Where(i => i.Customer == request.CustomerId && assetIds.Contains(i.Id)).ToList();
dlcsContext.Images.RemoveRange(assetsFromDatabase);
```

`DeleteFromQueryAsync()` is a helper that executes the delete statement in the DB, without needing to read the values back.
`Images.RemoveRange(assetsFromDatabase)` marks them for deletion in the context but `.SaveChanges()` isn't called further on so they're never actually removed

donaldgray:
Fixes #656

As detailed in the original ticket there was a bug where `.SaveChanges()` wasn't called when bulk deleting images via `/deleteImages` endpoint. This is the main reason for this fix but while making the change I implemented batch raising of SNS notifications as some bulk deletion may be quite large and raising batch SNS notifications is much more efficient.

Summary of changes:
* Renamed existing `AssetNotificationSender` -> `IngestNotificationSender`. The latter now only has methods for raising ingest notifications to the Engine.
* Introduced a new `AssetNotificationSender` class, in API project, to handle raising notifications resulting from `Asset` status changes.
* Moved `SendAssetModifiedNotification()` from `IngestNotificationSender` to new `AssetNotificationSender` class. Simplified the signature to `SendAssetModifiedMessage(AssetModificationRecord notification)` because:
  * `AssetModificationRecord` contains before/after `Asset` properties and `ChangeType`. Used static creation methods to remove complex validation logic from original `SendAssetModifiedMessage` method.
  * `CustomerPathElement` is no longer passed in, it's only used for Delete notifications and is purely to allow the notification to be handled by downstream recipient so felt cleaner to be a messaging/notification concern than having caller look it up.
* Added overload of `SendAssetModifiedNotification` that takes a collection of `AssetModificationRecord` classes.
* Updated `TopicPublisher` to take a list of `AssetModifiedNotification` records, these contain serialised message and `ChangeType`. Method will now either publish a single topic, or publish topics in batches.

Note that the new `AssetNotificationSender` class only raises notifications for Deletes but I left in plumbing for Create + Update so that these can be utilised later.

### DIFF
--- a/src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
+++ b/src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
@@ -1,109 +1,143 @@
-﻿using Amazon.SimpleNotificationService;
+﻿using System.Net;
+using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 using DLCS.AWS.Settings;
 using DLCS.AWS.SNS;
 using DLCS.Model.Messaging;
+using FakeItEasy;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json;
-using Test.Helpers.Integration;
 
 namespace DLCS.AWS.Tests.SNS;
 
-[Collection(LocalStackCollection.CollectionName)]
-[Trait("Category", "Manual")]
 public class TopicPublisherTests
 {
     private readonly IAmazonSimpleNotificationService snsClient;
+    private readonly TopicPublisher sut;
 
-    public TopicPublisherTests(LocalStackFixture localStackFixture)
+    public TopicPublisherTests()
     {
-        snsClient = localStackFixture.AWSSNSFactory();
+        snsClient = A.Fake<IAmazonSimpleNotificationService>();
         
-        snsClient.CreateTopicAsync(new CreateTopicRequest("someTopic"));
-    }
-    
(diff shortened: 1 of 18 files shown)


src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
```diff
@@ -1,109 +1,98 @@
...
+using DLCS.AWS.Settings;
+using DLCS.AWS.SNS;
+using DLCS.Model.Messaging;
+using FakeItEasy;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+
+namespace DLCS.AWS.Tests.SNS;
+
+public class TopicPublisherTests
```
JackLewis-digirati, line 12: is it worth writing a negative test, i.e.: what happens when the publish fails?

JackLewis-digirati (approved):

PR #662, opened 2023-11-06
Adding ability to disable delivery channels

### LINKED ISSUE #659: Disable delivery-channel by feature flag
`deliveryChannel` can be specified when ingesting an asset. It's current form is a reduced form of how we want to eventually handle deliveryChannels (see #625) for permutations of payloads.

Current `deliveryChannel` handling was added for wellcome only - we don't want other integrations using this field as it is being repurposed. To avoid it being used we should have an `AllowDeliveryChannel` feature flag. When `false` (default) - delivery-channels are ignored or return a 400 (which is best?). When `true` delivery-channels are respected.

JackLewis-digirati:
Resolves #659 

This pull request adds a setting called `DeliveryChannelsDisabled` to the API.  When this property is enabled, if a request is sent with a `DeliveryChannel`, a `400` response will be returned, otherwise the API will work as intended

### DIFF
--- a/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
+++ b/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
@@ -13,7 +13,7 @@ public class HydraImageValidatorTests
 
     public HydraImageValidatorTests()
     {
-        var apiSettings = new ApiSettings();
+        var apiSettings = new ApiSettings() { DeliveryChannelsEnabled = true};
         sut = new HydraImageValidator(Options.Create(apiSettings));
     }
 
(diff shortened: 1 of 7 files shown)


src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
```diff
@@ -223,10 +226,30 @@ public void DeliveryChannel_CanContainKnownValues(string knownValues)
...
         result.ShouldHaveValidationErrorFor(a => a.DeliveryChannels);
     }
+    
+        
+    [Fact]
+    public void DeliveryChannel_ValidationError_WhenDeliveryChannelsDisabled()
+    {
+        var apiSettings = new ApiSettings() { DeliveryChannelsDisabled = true};
+        var imageValidator = new HydraImageValidator(Options.Create(apiSettings));
+        var model = new DLCS.HydraModel.Image { DeliveryChannels = new[] { "iif-img" } };
```
donaldgray, line 242: typo `"iiif-img"`

donaldgray (approved):
As per ticket, you should need to opt-in to allow API to handle delivery-channels rather than having them on by default.

PR #665, opened 2023-11-07
update to priority queue to allow family or delivery channel to not be set

### LINKED ISSUE #657: priority queue requires a family or delivery channel to be set
While writing tests for the priority queue, the following response was observed:

```
{
    "@type": "Error",
    "statusCode": 400,
    "status": 400,
    "title": "Create priority batch failed: Validation failed",
    "detail": "Priority queue only supports image assets",
    "instance": "https://api.dlcs-stage.digirati.io",
    "description": "Priority queue only supports image assets"
}
```

After checking all assets in the payloads, it appears as if the payload requires either a image family or a delivery channel to be set to be handled as an image.  This differs from other calls such as the normal batch call to work.

The code that breaks is likely due to these lines being called before defaulting the value in `CreateBatchOfImages.cs`:

```
if (request.Assets.Any(a =>
                    a.Family != AssetFamily.Image && !a.HasDeliveryChannel(AssetDeliveryChannels.Image)))
            {
                return ModifyEntityResult<Batch>.Failure("Priority queue only supports image assets",
                    WriteResult.FailedValidation);
            }
```

JackLewis-digirati:
Resolves #657

This pull request updates the priority queue so that it can handle not having a `family` or `deliveryChannel` set, which brings it inline with `batch POST` .  This works by checking the `MIMEtype` in addition to the other 2 properties.

### DIFF
--- a/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
+++ b/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
@@ -943,7 +943,7 @@ public class CustomerQueueTests : IClassFixture<ProtagonistAppFactory<Startup>>
         await dbContext.CustomerStorages.AddTestCustomerStorage(customerId);
         await dbContext.SaveChangesAsync();
 
-        // a batch of 3 images - 1 with Family, 1 with DC and 1 with both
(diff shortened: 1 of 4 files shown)


donaldgray (approved):
LGTM - does pr desc need updated to line up with last change?

PR #666, opened 2023-11-07
checking for image location to allow deletion of assets without an image location

### LINKED ISSUE #636: Unable to delete assets that failed to ingest
The API currently returns `404 Not Found` if you attempt to delete an asset that cannot be ingested (i.e, the file type is unrecognized) but it should be deletable regardless of the state it's in.

JackLewis-digirati:
Resolves #636

This pull request updates checks found in the `AssetRepository` class to include checking that the asset exists in the image location table.  The reason for this is that file assets don't contain an `imageLocation` record, causing the check to error out and a `404` to be returned to the user.

### DIFF
--- a/src/protagonist/API.Tests/Features/Assets/ApiAssetRepositoryTests.cs
+++ b/src/protagonist/API.Tests/Features/Assets/ApiAssetRepositoryTests.cs
@@ -293,7 +293,7 @@ public class ApiAssetRepositoryTests
         var result = await sut.DeleteAsset(assetId);
         
         // Assert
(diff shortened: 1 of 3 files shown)


donaldgray (approved):
LGTM

The logic inside `catch (DbUpdateConcurrencyException dbEx)` can be deleted now as it's not required - this was added to return a 404 if `ImageLocation` doesn't exist. Would you be able to remove it? It's the type of code that's not particularly self-explanatory so can go on living when it's not required anymore

PR #669, opened 2023-11-15
Change Portal time format to `yyyy-MM-dd hh:mm:ss`

griffri:
This PR changes the time format used in the portal to `yyyy-MM-dd hh:mm:ss`, like the previous version:

image

### DIFF
--- a/src/protagonist/Portal/Pages/Batches/Index.cshtml
+++ b/src/protagonist/Portal/Pages/Batches/Index.cshtml
@@ -52,7 +52,7 @@
         </div>
         @if (Model.Batch.Finished != null)
         {
-            <span>Completed: @Model.Batch.Finished</span>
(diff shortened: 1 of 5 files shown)


donaldgray (commented):
Space details page is showing default date format: `Spaces/Details/:spaceId`.

Some of these use `.Value` on a nullable `DateTime` - are those always guaranteed to have a value?

griffri:
> Some of these use `.Value` on a nullable `DateTime` - are those always guaranteed to have a value?

I think so. I've yet to have seen the API not provide a creation date timestamp upon creating an image. The timestamps for completing a batch seem pretty safe, too.
Would it be good idea to change this just in case?

donaldgray:
Yeah I think so. The extension could be for `DateTime?` and handle null values. There are a lot of batches in DB without a finished date so these could surface at some point

donaldgray (approved):

PR #673, opened 2023-11-22
Add RFC for new composite-handler storage keys

donaldgray:


docs/rfcs/013-pdfs-as-input-storage.md
```diff
@@ -0,0 +1,69 @@
...
+      "incrementSeed": 0,
+      "compositeId": "my-internal-identifier-123"
+    }
+  ]
+}
+```
+
+### Side effects
+
+If a PDF is resubmitted and has decreased in size, say from 200 -> 150 pages, there could still be orphaned images stored in S3. This number is likely to be negligable unless there are huge variations in size.
```
tomcrane, line 67: A separate process could still identify orphans and delete them, e.g., an orphan check lambda runs after a PDF is processed.

tomcrane (approved):

PR #674, opened 2023-11-22
Ignore case when checking for existing user e-mail addresses in Portal

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/120

Changes in this PR:

- Case-insensitive string comparison is now used when verifying that a prospective user's e-mail address isn't already in use
- The expiry date for signup links is now converted to UTC upon creation to fit current DB rules
- The "copy link to clipboard" button that appears after creating a new signup link now copies over the correct URL
- Hints in the signup form now display correctly whenever validation fails (previously, they would display to the right of the validation messages rather than below)

### DIFF
--- a/src/protagonist/API/Features/Customer/Requests/CreatePortalUser.cs
+++ b/src/protagonist/API/Features/Customer/Requests/CreatePortalUser.cs
@@ -57,9 +57,11 @@ public class CreatePortalUserHandler : IRequestHandler<CreatePortalUser, CreateP
             return new CreatePortalUserResult { Error = "Email address is invalid" };
         }
 
-        var userWithEmail = await dbContext.Users.SingleOrDefaultAsync(
-            u => u.Email == request.PortalUser.Email, cancellationToken: cancellationToken);
-        if (userWithEmail != null)
+        // get all locally for more string comparison support
+        var allUsers = await dbContext.Users.ToListAsync(cancellationToken);
+        var userWithEmail = allUsers.Any(user => user.Email.Equals(request.PortalUser.Email, 
+            StringComparison.InvariantCultureIgnoreCase));
+        if (userWithEmail)
         {
             return new CreatePortalUserResult { Conflict = true, Error = "Email address already in use." };
         }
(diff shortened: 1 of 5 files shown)


src/protagonist/API/Features/Customer/Requests/CreatePortalUser.cs
```diff
@@ -57,9 +57,11 @@ public async Task<CreatePortalUserResult> Handle(CreatePortalUser request, Cance
             return new CreatePortalUserResult { Error = "Email address is invalid" };
         }
 
-        var userWithEmail = await dbContext.Users.SingleOrDefaultAsync(
-            u => u.Email == request.PortalUser.Email, cancellationToken: cancellationToken);
-        if (userWithEmail != null)
+        // get all locally for more string comparison support
+        var allUsers = await dbContext.Users.ToListAsync(cancellationToken);
```
donaldgray, line 61: This is inefficient - not ideal but `.toLower()` both should cover this use case. A better long term solution would be to sort the collation used in DB

src/protagonist/Portal/Pages/Account/SignupFromLink.cshtml.cs
```diff
@@ -97,7 +97,9 @@ public async Task<IActionResult> OnPostAsync(string signupCode)
 
     private async Task<bool> EmailInUse(string inputEmail)
     {
-        return await dbContext.Users.AnyAsync(user => user.Email == inputEmail);
+        // get all locally for more string comparison support
+        var allUsers = await dbContext.Users.ToListAsync();
+        return allUsers.Any(user => user.Email.Equals(inputEmail, StringComparison.InvariantCultureIgnoreCase));
```
donaldgray, line 102: as `CreatePortalUser`

donaldgray (changes requested):

src/protagonist/API/Features/Customer/Requests/CreatePortalUser.cs
griffri: Will change. I think it's also worth mentioning that this is currently how it's handled in `CreateCustomer`
https://github.com/dlcs/protagonist/blob/7ea6586c51dac5ed21ecc8dc1bfd099f61446682/src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs#L137

donaldgray (approved):

PR #679, opened 2023-11-23
validate customer storage allowance before ingestion

### LINKED ISSUE #353: Maintain CustomerStorage counts
When an image is ingested into the Engine we verify that the size of the incoming asset won't exceed the allowance in the storage policy (via `ICustomerStorageRepository.VerifyStoragePolicyBySize`).

An `ImageStorage` record is written when saving both `I` and `T` family assets, the values set are:
* `ThumbnailSize` - `T` this is always 0. For `I` this is the total size of the thumbnail jpgs on disk, calculated before upload to S3.
* `Size` - For `T` this is the total size in S3 of all transcoded derivatives. For `I` it is the size of the generated derivative on disk, calculated before upload to S3.

However, the check to verify whether the file has been exceeded is done based on the size of the _origin_, rather than the derivatives that are actually stored which doesn't seem correct (e.g. a 2GB lossles audio file would be a lot smaller as an MP3).

The overall `CustomerStorage` table isn't updated either, this needs to be done as part of the ingest save (or periodically as part of the save). This would need done for both for the individual Space, and for the Customer. The alternative if doing it periodically would need to be weighted up with any potential issues resulting from users over exceeding their limit.

JackLewis-digirati:
Resolves #353 

This PR contains updates to check there is free storage in the bucket prior to ingestion, and removes a check on the origin for size limits

### DIFF
--- a/src/protagonist/Engine/Ingest/IngestExecutor.cs
+++ b/src/protagonist/Engine/Ingest/IngestExecutor.cs
@@ -2,6 +2,7 @@
 using DLCS.Model.Customers;
 using DLCS.Model.Storage;
 using Engine.Data;
+using Engine.Ingest.Models;
 
 namespace Engine.Ingest;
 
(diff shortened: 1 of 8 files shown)


src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -32,35 +36,46 @@ public class IngestExecutor
...
 
         if (!assetIngestorSizeCheck.CustomerHasNoStorageCheck(asset.Customer))
         {
             var preIngestionAssetSize = await assetRepository.GetImageSize(asset.Id, cancellationToken);
             context.WithPreIngestionAssetSize(preIngestionAssetSize);
+            
+            var counts = await storageRepository.GetStorageMetrics(asset.Customer, cancellationToken);
+            
+            if (!counts.CanStoreAssetSize(MinimumAssetSize, 0))
+            {
```
donaldgray, line 49: Is it cleaner to short circuit here and return? Makes it easier to see at a glance that processing stops if true and there's nothing more to do.

```cs
if (!counts.CanStoreAssetSize(MinimumAssetSize, 0))
{
  logger.LogDebug("Something informative");
  asset.Error = "StoragePolicy size limit exceeded";
  var dbSuccess = await CompleteAssetInDatabase(context, true, cancellationToken);
  return new IngestResult(asset, dbSuccess ? IngestResultStatus.StorageLimitExceeded : IngestResultStatus.Failed);
}
```

`"StoragePolicy size limit exceeded"` is used here and `AssetIngestorSizeCheckBase` - worth adding a const?

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -32,35 +36,46 @@ public class IngestExecutor
...
         
         var context = new IngestionContext(asset);
+        var overallStatus = IngestResultStatus.Unknown;
 
         if (!assetIngestorSizeCheck.CustomerHasNoStorageCheck(asset.Customer))
         {
             var preIngestionAssetSize = await assetRepository.GetImageSize(asset.Id, cancellationToken);
             context.WithPreIngestionAssetSize(preIngestionAssetSize);
+            
+            var counts = await storageRepository.GetStorageMetrics(asset.Customer, cancellationToken);
```
donaldgray, line 46: Can this check go before `var preIngestionAssetSize = await assetRepository.GetImageSize(...)`? If so it could save a DB hit to get `ImageStorage` record in the event of storage being exceeded.

donaldgray (approved):
LGTM, couple of comments but overall looks good

PR #682, opened 2023-11-27
Getting retrieve space by name to cache correctly

### LINKED ISSUE #598: Fix space cache key
There are 2 `GetSpace` methods in `SpaceRepository` - one for fetching by name and one for id. Internally they both call `GetSpaceInternal`. The former passing -1 for Id.

`GetSpaceInternal` caches using `var key = $"space:{customerId}/{spaceId}";`

Searching for space with name "foo" then later for "bar" could result in "foo" space being returned for "bar" as it'll be cached for spaceId -1. This could result in a PUT or PATCH failing as it looks like the name will already exist.

Also searching by Id uses `.AsNoTracking()` but name doesn't - these should be consistent.

JackLewis-digirati:
Resolves #598 

This fix updates the `GetSpace` by name method to cache with a key like `space:{customer id}/{name of space}`, in similarity to the `GetSpace` by id method

### DIFF
--- a/src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
+++ b/src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
@@ -11,17 +11,7 @@ public interface ISpaceRepository
     Task<Space?> GetSpace(int customerId, int spaceId, CancellationToken cancellationToken);
     
     Task<Space?> GetSpace(int customerId, int spaceId, bool noCache, CancellationToken cancellationToken);
-    
-    /// <summary>
-    /// Retrieves a space by name
-    /// </summary>
-    /// <param name="customerId">The customer to retrieve a space from</param>
-    /// <param name="name">The name of the space</param>
-    /// <param name="noCache">Whether the value should be cached</param>
-    /// <param name="cancellationToken">The cancellation token</param>
-    /// <returns>A space, or null if it can't be found</returns>
-    Task<Space?> GetSpace(int customerId, string name,  bool noCache, CancellationToken cancellationToken);
-    
+
     /// <summary>
     /// Retrieves a space by name
     /// </summary>
--- a/src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
+++ b/src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
@@ -12,6 +12,23 @@ public interface ISpaceRepository
     
     Task<Space?> GetSpace(int customerId, int spaceId, bool noCache, CancellationToken cancellationToken);
     
+    /// <summary>
+    /// Retrieves a space by name
+    /// </summary>
+    /// <param name="customerId">The customer to retrieve a space from</param>
+    /// <param name="name">The name of the space</param>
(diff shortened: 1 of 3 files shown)


src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
```diff
@@ -12,6 +12,23 @@ public interface ISpaceRepository
     
     Task<Space?> GetSpace(int customerId, int spaceId, bool noCache, CancellationToken cancellationToken);
     
+    /// <summary>
+    /// Retrieves a space by name
+    /// </summary>
+    /// <param name="customerId">The customer to retrieve a space from</param>
+    /// <param name="name">The name of the space</param>
+    /// <param name="noCache">Whether the value should be cached</param>
```
donaldgray, line 20: Results are always cached, `noCache` indicates whether the query can be satisfied by using a cached value or if it should always be fresh from db.

src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
```diff
@@ -12,6 +12,23 @@ public interface ISpaceRepository
...
+    /// <param name="cancellationToken">The cancellation token</param>
+    /// <returns>A space, or null if it can't be found</returns>
+    Task<Space?> GetSpace(int customerId, string name,  bool noCache, CancellationToken cancellationToken);
```
donaldgray, line 23: Is it worth introducing an overload that isn't used? Would it be better to introduce it when/if it's required?

src/protagonist/DLCS.Repository/Spaces/SpaceRepository.cs
```diff
@@ -116,7 +121,7 @@ private async Task<int> GetIdForNewSpace(int requestCustomer)
         CancellationToken cancellationToken, string? name = null,
         bool withApproximateImageCount = false, bool noCache = false)
     {
-        var key = $"space:{customerId}/{spaceId}";
+        var key = $"space:{customerId}/{name ?? spaceId.ToString()}";
```
donaldgray, line 124: We don't add any restrictions on space names so if customer 123 creates 2 spaces: 

```json
[
  { "Id": 1, "Name": "2" }
  { "Id": 2, "Name": "1" }
]
```

Those could both be legitimately cached as either `space:123:1` or `space:123:2` depending on how the request was made.

donaldgray (approved):

src/protagonist/DLCS.Repository/Spaces/SpaceRepository.cs
JackLewis-digirati: to avoid this, I could change the key based on the cache type? i.e.: `space:123:name:2` or `space:123:id:1`

src/protagonist/DLCS.Model/Spaces/ISpaceRepository.cs
JackLewis-digirati: I added this in order to mirror the `id` methods provided, but thinking about it now, it's not being used so probably better to remove it

PR #687, opened 2023-12-04
Publishing to SNS on create and update

### LINKED ISSUE #378: Raise notification when Asset modified
This is stubbed as `IAssetNotificationSender.SendAssetModifiedNotification` and currently logs a message without a notification being raised.

JackLewis-digirati:
Resolves #378 

This PR adds the ability to raise a notification in SNS on create and update, in addition to delete


src/protagonist/API/Infrastructure/Messaging/AssetNotificationSender.cs
```diff
@@ -101,11 +100,15 @@ private async Task<CustomerPathElement> GetCustomerPathElement(int customer)
         return customerPathElement;
     }
     
-    private async Task<bool> SendCleanupAssetRequest(IList<string> change, CancellationToken cancellationToken)
+    private async Task<bool> SendAssetModifiedRequest(Dictionary<ChangeType, List<string>> change, CancellationToken cancellationToken)
     {
         if (change.IsNullOrEmpty()) return true;
 
-        var toSend = change.Select(n => new AssetModifiedNotification(n, ChangeType.Delete)).ToList();
+        var toSend = (
```
donaldgray, line 107: Why use query syntax here? I think this is the only place in the codebase where query syntax is used rather than method syntax

donaldgray (changes requested):
LGTM, change required is AC's state `message includes asset; update has both old and new asset` but currently it'll only send `Before` for update.

PR #693, opened 2023-12-07
Add status code error pages to Portal

griffri:
Resolves https://github.com/dlcs/private-protagonist/issues/142

This PR adds a friendlier looking page for status code errors to the Portal:

image
image

Additionally, the space, batch and image details pages now throw NotFound() if their respective resources cannot be retrieved by the API.


src/protagonist/Portal/Pages/Exception.cshtml
```diff
@@ -0,0 +1,26 @@
...
+@if (Model.ShowRequestId)
+{
+    <p>
+        <strong>Request ID:</strong> <code>@Model.RequestId</code>
+    </p>
+}
+
+<h3>Development Mode</h3>
+<p>
+    Swapping to the <strong>Development</strong> environment displays detailed information about the error that occurred.
```
donaldgray, line 19: It's been a while since I've done anything with custom error handling in razor but I don't think there's a need to copy the markup of the default error page - there is a middleware to handle that. It will enable full stack trace etc to be shown. 

As is would `Swapping to the <strong>Development</strong> environment displays detailed information about the error that occurred.` happen?

donaldgray (changes requested):
The log messages should include customerId as it's not obvious if looking at logs who was having an issue. When using `logger.LogError()` you can pass it the caught exception to get the full stack trace recorded.

However, I'm not sure whether returning `null` is the best way of handling exception as the Portal then loses sight of whether it was a client or server error (e.g. 404 or 503).
The `ReadAsHydraResponseAsync` method will throw a `DlcsException` if a non 2xx status code was returned. Currently `DlcsException` only contains the description of the returned error but it could be extended to store more details from the `Error` object (e.g. the statusCode from the API). This exception could be bubbled up and consumers can handle the different status codes appropriately - I think we'd only want to show a 404 for the Portal but if we wanted more accurate reporting we could do (e.g. "The API is down, try later").

Returning null from `ApiClient` works for Portals needs right now but the idea is that `ApiClient` could be reused by other integrations in the future and just returning `null` removes the richer `Error` type that the API will return.

Unrelated by `"error-page-message"` could be a constant as it's used in a lot of places

src/protagonist/API.Client/DlcsClient.cs
```diff
@@ -81,9 +81,17 @@ public class DlcsClient : IDlcsClient
...
-        var response = await httpClient.GetAsync(url);
-        var space = await response.ReadAsHydraResponseAsync<Space>(jsonSerializerSettings);
-        return space;
+        try
+        {
+            var response = await httpClient.GetAsync(url);
+            var space = await response.ReadAsHydraResponseAsync<Space>(jsonSerializerSettings);
+            return space;
+        }
+        catch (Exception ex)
```
JackLewis-digirati, line 90: I don't think the `ex` is used - might be better to remove it, or pass it into the logger with `logger.LogError(ex, "Failed to retrieve space {SpaceId}", spaceId);`

src/protagonist/Portal/Pages/Error.cshtml.cs
```diff
@@ -1,27 +1,40 @@
...
 
-    public bool ShowRequestId => !string.IsNullOrEmpty(RequestId);
-
-    private readonly ILogger<ErrorModel> _logger;
+    public string? Message { get; set; }
+        
+    public IActionResult OnGet(HttpStatusCode code)
+    {
+        // Return Bad Request if an invalid status code is provided
+        if (code == 0)
```
JackLewis-digirati, line 17: I haven't tested this, but would something like `code.Equals(default(HttpStatusCode))` work? Avoids having a magic number

griffri: It does! I'll patch that in, thanks.

PR #697, opened 2023-12-11
adding cache invalidation flag

### LINKED ISSUE #676: When deleting items should we have a 'clearCaches' flag?
I think it may be useful to be able to specify whether we want to invalidate caches (varnish and CDN) or not when deleting items from DLCS. 

This would allow _Delete this NOW!_ vs _I'm cleaning out old stuff, it can gradually fade_ and could save a lot of churn in caches.

I deleted a large number of items recently and there were ~48K+ items in the varnish delete queue and I think it started affecting varnish performance and seemed to take a while to decrease. It's also easy to hit `TooManyInvalidationsInProgressException` when invalidating CloudFront for a number of images.

JackLewis-digirati:
Resolves #676 

This PR adds flags to `deleteImages` and `delete` that allow for the cache to be cleaned only when using a `deleteFrom` query param with the following options:

- `cdn`
  - performs cloudfront invalidation
- `internalCache`
  - performs varnish `Ban` requests

An example of this call would belike this:

`https://localhost:5002/customers/26/spaces/19/images/assetCreated3?deleteFrom=cdn,internalCache`


donaldgray (approved):
lgtm

Issue #642, opened 2023-10-10
ThumbsCalculator not always serving correct thumbs


donaldgray:
There is a bug in the `ThumbnailCalculator` class. This is used by the Thumbs service to calculate whether the incoming request can be handled using pregenerated thumbs, and also in Orchestrator image-handling to do the same.

The issue is that the w+h are not propertly being checked, if either supplied w or h match the longest edge of the generated thumb, regardless of what dimension it is, that thumbnail will be returned.

Example, for image that is 200w 133h:
* `/full/200,133/0/default.jpg` returns 200x133 thumb (expected)
* `/full/200,190/0/default.jpg` returns 200x133 thumb (error - matches 200 w but doesn't resize)
* `/full/10,190/0/default.jpg` returns 10x190 image (expected)
* `/full/10,200/0/default.jpg` returns 200x133 thumb (error - matches 200 height)

Fix also needs to update the `ThumbnailCalculator` to handle `!`/confined parameter.

donaldgray:
## Acceptance Criteria

Given the above example 200w, 133h thumbnail. The following is what is expected

* `/iiif-img/full/200,133/0/default.jpg` returns 200x133 thumb (`x-proxy-destination: Thumbs`)
* `/iiif-img/full/!200,200/0/default.jpg` returns 200x133 thumb (`x-proxy-destination: Thumbs`)
* `/iiif-img/full/200,190/0/default.jpg` returns 200x133 image from image-server (`x-proxy-destination: SpecialServer`)
* `/iiif-img/full/10,190/0/default.jpg` returns 10x190 image from image-server (`x-proxy-destination: SpecialServer`)
* `/iiif-img/full/10,200/0/default.jpg` returns 200x133 image from image-server (`x-proxy-destination: SpecialServer`)
* `/thumbs/full/200,133/0/default.jpg` returns 200x133 thumb (`x-proxy-destination: Thumbs`)
* `/thumbs/full/!200,200/0/default.jpg` returns 200x133 thumb (`x-proxy-destination: Thumbs`)
* `/thumbs/full/200,190/0/default.jpg` 404
* `/thumbs/full/10,190/0/default.jpg` 404
* `/thumbs/full/10,200/0/default.jpg` 404

PR #699, opened 2023-12-14
`ThumbnailCalculator` bugfixes

griffri:
Resolves https://github.com/dlcs/protagonist/issues/642

This PR fixes a bug with `ThumbnailCalculator` that caused it to serve the wrong pre-generated thumbnail candidate. 

It also fixes the `ProxiesToThumbs` tests in `ImageRequestHandlerTests`, which have been broken by this change. They previously expected a request asking for a confined size of 100x150 to cause Orchestrator to return an image from the thumbnail server, but as the stored thumbnail is 150x150 it received an image from the special server instead.

### DIFF
--- a/src/protagonist/DLCS.Repository/Assets/ThumbnailCalculator.cs
+++ b/src/protagonist/DLCS.Repository/Assets/ThumbnailCalculator.cs
@@ -25,11 +25,21 @@ public static class ThumbnailCalculator
 
     private static SizeCandidate GetLongestEdge(List<Size> sizes, ImageRequest imageRequest)
     {
-        // get the longest dimension of the requested size
-        var max = Math.Max(imageRequest.Size.Width ?? 0, imageRequest.Size.Height ?? 0);
-        
         if (imageRequest.Size.Width > 0 && imageRequest.Size.Height > 0)
         {
+            // get the longest dimension of the requested size
+            var max = Math.Max(imageRequest.Size.Width ?? 0, imageRequest.Size.Height ?? 0);
+            var foundExactSize = sizes.Exists(s => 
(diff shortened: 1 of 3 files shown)


JackLewis-digirati (approved):

src/protagonist/DLCS.Repository/Assets/ThumbnailCalculator.cs
```diff
@@ -25,13 +25,54 @@ public static SizeCandidate GetCandidate(List<Size> sizes, ImageRequest imageReq
...
+            
+            // If the image is confined, are there any sizes that fit?
+            if (imageRequest.Size.Confined)
+            {
+                // Pick the first thumbnail size as a reference for shape
+                var shape = sizes.First().GetShape();
+
+                // If this is a landscape image, max should match its width
+                if (shape == ImageShape.Landscape 
+                    && sizes.Exists(s => s.Width == max && requestHeight >= s.Height))
```
donaldgray, line 53: Removing `&& requestHeight >= s.Height` from this query doesn't fail any tests - it should to avoid it accidentally being removed in the future (same with `&& requestWidth >= s.Width` below)

donaldgray (changes requested):

src/protagonist/DLCS.Repository/Assets/ThumbnailCalculator.cs
griffri: Tests for ensuring that the correct shortest edge is being picked have now been added in https://github.com/dlcs/protagonist/pull/699/commits/6d796dca1c816b35d4f351a044f1c04da903cba6

PR #701, opened 2024-01-02
Adding ability to restrict allowed characters in an asset id and allowing `(` and `)`

### LINKED ISSUE #323: Handle '(' and ')' characters in asset identifiers
AssetId's that contain either `(` or `)` characters are not currently supported as they result in an ingestion error when passed to [`appetiser`](https://github.com/dlcs/appetiser) for derivative generation. This is due to how `appetiser` launches a subprocess to run `kdu_compress`. This will require a fix implemented in `appetiser` to escape or quote `(` + `)`. 

We also need to consider other areas in the DLCS where this may cause issues:
* On orchestration
* When generating URL to pass to downstream server
* Any others?

It may be easier to rewrite on the way in + out so that the only place `(` and `)` are handled is at the boundary of the system (e.g. in `id` values on info.json).

### LINKED ISSUE #340: Enforce restrictions on Asset IDs
Currently these are any string (although see https://github.com/dlcs/protagonist/issues/323)

But we would prefer them to not be susceptible to corruption through escaping when they appear in URLs.

This is related to https://github.com/wellcomecollection/docs/pull/30

The DLCS needs to enforce rules and reject asset IDs that break them - but how restrictive do we want to be? As restrictive as archivematica rules mentioned in that RFC?

And what do we do about existing non-compliant assets?

JackLewis-digirati:
Resolves #340 
Resolves #323

This pull request adds in the ability to restrict the characters allowed in an asset id.  It does this by adding a set of 2 new settings:

```
"RestrictedAssetIdCharacters": "\\ "
"DisableStrictAssetIdChecks": true
```

`RestrictedAssetIdCharacters` is a series of characters that are not allowed within an asset id

`DisableStrictAssetIdChecks`  sets whether the check for strict asset ids is disabled.  As such, this check is *enabled* by default

Additionally, this pull request enables `(` and `)` characters to be used in the image id which would previously cause appetiser to fail.  It manages this by replacing these characters with an alternative.  By default this is an underscore, but can be set to a specific string via settings to be able to reduce the possibility of collisions

### DIFF
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -99,9 +99,12 @@ public class ImageController : HydraController
         {
             hydraAsset = LegacyModeConverter.VerifyAndConvertToModernFormat(hydraAsset);
         }
-        
-        hydraAsset.ModelId = imageId;
 
+        if (hydraAsset.ModelId == null)
+        {
+            hydraAsset.ModelId = imageId;
+        }
+        
         var validationResult = await validator.ValidateAsync(hydraAsset, cancellationToken);
         if (!validationResult.IsValid)
         {
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -66,7 +66,7 @@ public class ImageController : HydraController
     /// <param name="hydraAsset">The body of the request contains the Asset in Hydra JSON-LD form (Image class)</param>
     /// <returns>The created or updated Hydra Image object for the Asset</returns>
     /// <remarks>
-    /// Sample requests:
+    /// Sample request:
     ///
     ///     PUT: /customers/1/spaces/1/images/my-image
     ///     {
@@ -75,13 +75,6 @@ public class ImageController : HydraController
     ///         "origin": "https://example.text/.../image.jpeg",
     ///         "mediaType": "image/jpeg",
     ///         "string1": "my-metadata"
-    ///     }
-    ///
-    ///     PUT: /customers/1/spaces/1/images/my-image
-    ///     {
-    ///         "@type":"Image",
-    ///         "family": "I",
-    ///         "file": "/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAM...."
     ///     }
     /// </remarks>
     [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(DLCS.HydraModel.Image))]
@@ -107,6 +100,8 @@ public class ImageController : HydraController
             hydraAsset = LegacyModeConverter.VerifyAndConvertToModernFormat(hydraAsset);
         }
         
+        hydraAsset.ModelId = imageId;
+
         var validationResult = await validator.ValidateAsync(hydraAsset, cancellationToken);
         if (!validationResult.IsValid)
         {
@@ -212,9 +207,8 @@ public class ImageController : HydraController
     }
 
     /// <summary>
-    /// <para>Ingest specified file bytes to DLCS.
-    /// "File" property should be base64 encoded image.</para>
-    /// <para>This route is now deprecated. <see cref="PutImage"/> should be used instead.</para>
+    /// Ingest specified file bytes to DLCS. Only "I" family assets are accepted.
+    /// "File" property should be base64 encoded image. 
     /// </summary>
     /// <remarks>
     /// Sample request:
@@ -242,7 +236,8 @@ public class ImageController : HydraController
             "Warning: POST /customers/{CustomerId}/spaces/{SpaceId}/images/{ImageId} was called. This route is deprecated.",
             customerId, spaceId, imageId);
         
-        return await PutImage(customerId, spaceId, imageId, hydraAsset, validator, cancellationToken);
+
+        return await PutOrPatchAsset(customerId, spaceId, imageId, hydraAsset, cancellationToken);
     }
 
     /// <summary>
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -235,6 +235,7 @@ public class ImageController : HydraController
         [FromServices] HydraImageValidator validator,
         CancellationToken cancellationToken)
     {
+
         logger.LogWarning(
             "Warning: POST /customers/{CustomerId}/spaces/{SpaceId}/images/{ImageId} was called. This route is deprecated.",
             customerId, spaceId, imageId);
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -241,7 +241,7 @@ public class ImageController : HydraController
             customerId, spaceId, imageId);
         
 
-        return await PutOrPatchAsset(customerId, spaceId, imageId, hydraAsset, cancellationToken);
+        return await PutImage(customerId, spaceId, imageId, hydraAsset, validator, cancellationToken);
     }
 
     /// <summary>
--- a/src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
+++ b/src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
@@ -35,7 +35,14 @@ public class HydraImageValidator : AbstractValidator<DLCS.HydraModel.Image>
         RuleFor(a => a.DeliveryChannels).Must(d => d.IsNullOrEmpty())
             .When(_ => !apiSettings.Value.DeliveryChannelsEnabled)
             .WithMessage("Delivery channels are disabled");
-        
+
+        if (!apiSettings.Value.DisableStrictAssetIdChecks)
+        {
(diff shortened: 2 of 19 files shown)


src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -93,12 +93,13 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
             hydraAsset = LegacyModeConverter.VerifyAndConvertToModernFormat(hydraAsset);
         }
         
+        hydraAsset.ModelId = imageId;
```
donaldgray, line 96: Should this happen prior to `VerifyAndConvertToModernFormat` as that method could change `hydraAsset.ModelId`?

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -36,7 +36,11 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
             .When(_ => !apiSettings.Value.DeliveryChannelsEnabled)
             .WithMessage("Delivery channels are disabled");
         
-        RuleForEach(a => a.DeliveryChannels)
+        RuleFor(a => a.ModelId).Must(id => id.IndexOfAny(apiSettings.Value.RestrictedAssetIdCharacters.ToCharArray()) == -1)
+            .When(asset => !asset.ModelId.IsNullOrEmpty())
+            .WithMessage("Asset id contains a restricted character");
```
donaldgray, line 41: Restricted characters should be included in the exception message to help the caller resolve the issue

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -36,7 +36,11 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
             .When(_ => !apiSettings.Value.DeliveryChannelsEnabled)
             .WithMessage("Delivery channels are disabled");
         
-        RuleForEach(a => a.DeliveryChannels)
+        RuleFor(a => a.ModelId).Must(id => id.IndexOfAny(apiSettings.Value.RestrictedAssetIdCharacters.ToCharArray()) == -1)
```
donaldgray, line 39: Is it worth storing `RestrictedAssetIdCharacters` as a `char[]` if it's going to be converted to that for each check?

src/protagonist/Engine.Tests/Ingest/Image/Appetiser/AppetiserClientTests.cs
```diff
@@ -78,6 +78,21 @@ public async Task ProcessImage_CreatesAndRemovesRequiredDirectories()
         A.CallTo(() => fileSystem.CreateDirectory(A<string>._)).MustHaveHappenedTwiceExactly();
         A.CallTo(() => fileSystem.DeleteDirectory(A<string>._, true, true)).MustHaveHappenedTwiceExactly();
     }
+    
+    [Fact]
+    public async Task ProcessImage_UpdatesContextBasedOnImageIdWithBrackets()
```
donaldgray, line 83: Is `UpdatesContext` correct? Or is the change where the file is saved?

src/protagonist/Engine/Ingest/Image/Appetiser/AppetiserClient.cs
```diff
@@ -183,11 +186,11 @@ private static void UpdateImageDimensions(Asset asset, AppetiserResponseModel re
         asset.Width = responseModel.Width;
     }
 
-    private async Task ProcessOriginImage(IngestionContext context, ImageProcessorFlags processorFlags)
+    private async Task ProcessOriginImage(IngestionContext context, ImageProcessorFlags processorFlags, AssetId modifiedAssetId)
```
donaldgray, line 189: I don't think we want to change the key in S3 if it contains a bracket character - or if we do it should happen when all S3 keys are generated (ie in `IStorageKeyGenerator` impl), rather than just in `AppetiserClient`.

As-is we could end up with S3 keys that are inconsistent for the same object. If I create `9/9/test(1)` then in S3 we could end up with:
* `thumbs-bucket/9/9/test(1)/s.json` - s.json for thumbns
* `thumbs-bucket/9/9/test(1)/100.jpg` - a 100px thumb
* `storage-bucket/9/9/test(1)/cantaloupe/v3/info.json` - cached info.json
* `storage-bucket/9/9/test_1_` - derivative jp2 for use by image-server
* `storage-bucket/9/9/test(1)/original` - source image for use by image-server (if `use-original`)

The root issue is how linux handles `(` and `)` in filenames - we can either escape/replace these when writing to disk or update where they are stored in S3. cc @tomcrane for thoughts

src/protagonist/Engine/Ingest/Image/ImageIngestionHelpers.cs
```diff
@@ -15,7 +16,9 @@ public static string GetSourceFolder(Asset asset, EngineSettings engineSettings)
         var root = imageIngest.GetRoot();
             
         // source is the main folder for storing downloaded image
-        var assetId = asset.Id;
+        var assetId = new AssetId(asset.Id.Customer, asset.Id.Space,
+            asset.Id.Asset.Replace("(", imageIngest.OpenBracketReplacement)
+                .Replace(")", imageIngest.CloseBracketReplacement));;
```
donaldgray, line 21: `;;`

src/protagonist/Engine/Ingest/Persistence/AssetToDisk.cs
```diff
@@ -90,8 +90,8 @@ public class AssetToDisk : AssetMoverBase, IAssetToDisk
     {
         TrySetContentTypeForBinary(originResponse, asset);
         var extension = GetFileExtension(originResponse);
-        
-        var targetPath = $"{Path.Join(destinationTemplate, asset.Id.Asset)}.{extension}";
+
+        var targetPath = $"{Path.Join(destinationTemplate, asset.Id.Asset.Replace('(', '_').Replace(')', '_'))}.{extension}";
```
donaldgray, line 94: Should this logic be the same as other places (where `OpenBracketReplacement` and `CloseBracketReplacement` are used, rather than hardcoded underscore).

donaldgray (changes requested):
This doesn't appear to meet the criteria on #340.

> Configuration includes a string that contains the permitted characters

This has a configuration that stores restricted characters

> and also a flag to enable:
     no restrictions
     restriction for new assets but allow updates of existing
     fail for updates too

This has a single boolean flag to enable/disable and doesn't appear to differentiate creation/update

src/protagonist/API/Features/Image/ImageController.cs
JackLewis-digirati: hrm, I think the issue is that the model id holds the value needed to do validation checks, but it's not filled out until `PutOrPatchAsset`, which happens after validation.  However, looking at `VerifyAndConvertToModernFormat ` it sets the value based on a couple of fields.  Given this would it be better to have it after `VerifyAndConvertToModernFormat `, but also use a null check?

src/protagonist/Engine/Ingest/Image/Appetiser/AppetiserClient.cs
tomcrane: I agree that the names in S3 should always be consistent.
Asset identifiers, with this restriction, will always contain S3-key-safe characters. It's only when a file gets moved to a Unix filesystem for use by an image server that we have an issue, is that right?

Orchestrator does this - but does Cantaloupe also do it (if we configured it) for its own local disk cache? may be worth seeing what Cantaloupe does in that scenario.

Is it the case that `(` and `)` are the only chars we'd want in our whitelist that the filesystem doesn't like? if so we could have a special escape sequence for them `__open_p__` `__close_p__` or something.

We have a lot of identifiers in Wellcome that contain invalid chars - but they won't be images that need orchestrating - they'll be MP3s and similar.

donaldgray: > It's only when a file gets moved to a Unix filesystem for use by an image server that we have an issue, is that right?

The original issue was when file moved to unix filesystem for use by appetiser, rather than image-server. Orchestrator doesn't change or escape characters when copying to scratch disk so we may run into the issue on Orchestration too - will need to test to verify.

src/protagonist/Engine/Ingest/Image/Appetiser/AppetiserClient.cs
```diff
@@ -150,22 +153,22 @@ private string GetJP2FilePath(AssetId assetId, bool forImageProcessor)
...
             root: imageProcessorRoot);
 
-        unixPath += $"/{context.Asset.Id.Asset}.{extension}";
+        unixPath += $"/{modifiedAssetId.Asset}.{extension}";
         return unixPath;
     }
 
     private async Task ProcessResponse(IngestionContext context, AppetiserResponseModel responseModel, 
-        ImageProcessorFlags processorFlags)
+        ImageProcessorFlags processorFlags, AssetId modifiedAssetId)
```
donaldgray, line 171: `AssetId modifiedAssetId` is unnecessary here as it's not used.

donaldgray (approved):
Approved - just need to delete unused param

griffri (approved):

PR #704, opened 2024-01-05
Improving error reporting from appetiser

### LINKED ISSUE #637: Return an informative error message when ingesting an asset fails
The API currently returns an error containing the message `Engine was not able to process this asset` when ingesting an asset fails. This should be changed to something more helpful to the user - e.g, "Unsupported file type" if the file type of the asset is unknown.

JackLewis-digirati:
Resolves #637 
linked to [appetiser](https://github.com/dlcs/appetiser/pull/20)

This pull request updates the application to provide improved logging of errors that come from appetiser

### DIFF
--- a/src/protagonist/Engine.Tests/Ingest/Image/Appetiser/AppetiserClientTests.cs
+++ b/src/protagonist/Engine.Tests/Ingest/Image/Appetiser/AppetiserClientTests.cs
@@ -83,7 +83,18 @@ public class AppetiserClientTests
     public async Task ProcessImage_False_IfImageProcessorCallFails()
     {
         // Arrange
-        httpHandler.SetResponse(new HttpResponseMessage(HttpStatusCode.InternalServerError));
+        // Arrange
+        var imageProcessorResponse = new AppetiserResponseErrorModel()
+        {
+            Message = "error",
+            Status = "some status"
+        };
+
+        var response = httpHandler.GetResponseMessage(JsonSerializer.Serialize(imageProcessorResponse, Settings),
+            HttpStatusCode.InternalServerError);
+        response.Content.Headers.ContentType = new MediaTypeHeaderValue("application/json");
+        
+        httpHandler.SetResponse(response);
         var context = GetIngestionContext();
 
         // Act
@@ -93,6 +104,7 @@ public class AppetiserClientTests
         httpHandler.CallsMade.Should().ContainSingle(s => s == "http://image-processor/convert");
         result.Should().BeFalse();
         context.Asset.Should().NotBeNull();
+        context.Asset.Error.Should().Be("Appetiser Error: error");
     }
 
     [Theory]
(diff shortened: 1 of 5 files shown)


src/protagonist/Engine.Tests/Ingest/Image/Appetiser/AppetiserClientTests.cs
```diff
@@ -83,7 +83,18 @@ public async Task ProcessImage_CreatesAndRemovesRequiredDirectories()
     public async Task ProcessImage_False_IfImageProcessorCallFails()
     {
         // Arrange
-        httpHandler.SetResponse(new HttpResponseMessage(HttpStatusCode.InternalServerError));
+        // Arrange
```
donaldgray, line 86: Duplicate comment

src/protagonist/Engine/Ingest/Image/Appetiser/Models/AppetiserResponse.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Engine.Ingest.Image.Appetiser;
+
+public class AppetiserResponse
+{
+    public ResponseType ResponseType { get; init; }
```
donaldgray, line 5: This field doesn't appear to be used - can it be removed?

src/protagonist/Engine/Ingest/Image/Appetiser/AppetiserClient.cs
```diff
@@ -102,10 +113,19 @@ public async Task<bool> ProcessImage(IngestionContext context)
...
         // TODO - it's possible to get a 200 when appetiser doesn't do anything, e.g. body not understood
-        var responseModel = await response.Content.ReadFromJsonAsync<AppetiserResponseModel>();
+        AppetiserResponse? responseModel = null;
+        
+        if (response.IsSuccessStatusCode)
```
donaldgray, line 120: Is this guaranteed to be an indicator of the shape of message returned (thinking of the `// TODO - it's possible to get a 200 when appetiser doesn't do anything, e.g. body not understood` comment)? And if it is guaranteed to work does it depend on a specific version of Appetiser being available? 

Would it be better to look at the contents of the response message (and avoid a lock step deployment)

src/protagonist/Engine/Ingest/Image/Appetiser/Models/AppetiserResponse.cs
JackLewis-digirati: I can remove this field, but this was an attempt to avoid a marker interface - essentially I was choosing which was less bad

src/protagonist/Engine/Ingest/Image/Appetiser/AppetiserClient.cs
JackLewis-digirati: I could do a check on the image id, and if it get's null, try deserializing to the error format?

For the specific version of appetiser, I need to do some testing, but I think this will work with currently deployed versions of appetiser.  However, the error would be essentially junk (it would always say `500 Internal Server Error: The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.`)

donaldgray (approved):

griffri (approved):

PR #705, opened 2024-01-08
Allow Base64 encoded images to be ingested via PUT

### LINKED ISSUE #338: Should direct ingest of Image with Bytes (`ImageWithFile`) be a PUT?
Currently if you want to send a JSON blob with the asset bytes, you POST to `/customers/n/spaces/n/images/asset-name`

Whereas if you are sending a regular image with an Origin property, you PUT to `/customers/n/spaces/n/images/asset-name`

This feels wrong, they should both be PUTs.

(We have client code that calls this, though).

griffri:
Resolves #338

This PR changes the `PutImage` route in `ImageController` to support the `File` field for directly ingesting image bytes, it also includes a test for verifying this functionality.

`PostImageWithFileBytes` is now deprecated and re-routes requests to `PutImage`.

### DIFF
--- a/src/protagonist/API/Features/Image/ImageController.cs
+++ b/src/protagonist/API/Features/Image/ImageController.cs
@@ -61,7 +61,7 @@ public class ImageController : HydraController
     ///
     /// Image + File assets are ingested synchronously. Timebased assets are ingested asynchronously.
     ///
-    /// "File" property should be base64 encoded image, if included. Only "I" family assets are accepted.
+    /// "File" property should be base64 encoded image, if included.
     /// </summary>
     /// <param name="hydraAsset">The body of the request contains the Asset in Hydra JSON-LD form (Image class)</param>
     /// <returns>The created or updated Hydra Image object for the Asset</returns>
@@ -212,7 +212,7 @@ public class ImageController : HydraController
     }
 
     /// <summary>
-    /// <para>Ingest specified file bytes to DLCS. Only "I" family assets are accepted.
+    /// <para>Ingest specified file bytes to DLCS.
     /// "File" property should be base64 encoded image.</para>
     /// <para>This route is now deprecated. <see cref="PutImage"/> should be used instead.</para>
     /// </summary>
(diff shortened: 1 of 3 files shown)


src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -287,4 +287,29 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
             assetId.ToString(),
             "Upsert asset failed", cancellationToken);
     }
+    
+    private async Task<IActionResult> PutOrPatchAssetWithFileBytes(int customerId, int spaceId, string imageId,
+        ImageWithFile hydraAsset, CancellationToken cancellationToken)
+    { 
+        const string errorTitle = "POST of Asset bytes failed";
+        var assetId = new AssetId(customerId, spaceId, imageId);
+        if (hydraAsset.File == null || hydraAsset.File.Length == 0)
```
donaldgray, line 296: Is this necessary if the check happens prior to calling this method?

src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -194,8 +214,9 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
     }
 
     /// <summary>
-    /// Ingest specified file bytes to DLCS. Only "I" family assets are accepted.
-    /// "File" property should be base64 encoded image. 
+    /// <para>Ingest specified file bytes to DLCS. Only "I" family assets are accepted.
```
donaldgray, line 217: What restricts this ti `I` only? (I know you've not changed that element of it but I can't see anything obvious from review).

src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -79,26 +92,33 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
...
         [FromRoute] int customerId,
         [FromRoute] int spaceId,
         [FromRoute] string imageId,
-        [FromBody] DLCS.HydraModel.Image hydraAsset,
+        [FromBody] ImageWithFile hydraAsset,
         [FromServices] HydraImageValidator validator,
         CancellationToken cancellationToken)
     {
-        if (apiSettings.LegacyModeEnabledForSpace(customerId, spaceId))
+        var assetContainsFile = hydraAsset.File != null;
```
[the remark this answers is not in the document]
griffri: Good suggestion, thanks!

donaldgray (approved):

PR #707, opened 2024-01-17
Accept `wcDeliveryChannels` as an alias for `DeliveryChannels` in `HydraModel.Image`

### LINKED ISSUE #614: Rename Hydra API deliveryChannels property in 2023 Protagonist to wcDeliveryChannels

griffri:
Resolves #614

This PR:

- Adds `wcDeliveryChannels` as an alias for the `deliveryChannels` field in the Hydra model for `Image` 
- Adds a test for ensuring that it can be used to set the value of `DeliveryChannels`
- Prevents the `deliveryChannels` field from being returned in the JSON, returning `wcDeliveryChannels` instead
- Moves the `Patch_Asset_Fails_When_Delivery_Channels_Are_Disabled` test into a separate class for asset tests that require delivery channels to be disabled

### DIFF
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -571,6 +571,34 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
     }
     
     [Fact]
+    public async Task Put_New_Asset_Ignores_DeliveryChannels()
+    {
+        var assetId = new AssetId(99, 1, nameof(Put_New_Asset_Ignores_DeliveryChannels));
+        var hydraImageBody = $@"{{
+          ""@type"": ""Image"",
+          ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
+          ""family"": ""I"",
+          ""mediaType"": ""image/tiff"",
+          ""deliveryChannels"": [""file""]
+        }}";
+
+        A.CallTo(() =>
+                EngineClient.SynchronousIngest(
+                    A<IngestAssetRequest>.That.Matches(r => r.Asset.Id == assetId), false,
+                    A<CancellationToken>._))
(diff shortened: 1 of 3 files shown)


donaldgray (changes requested):
Doesn't meet acceptance criteria. 

> accepts "old" delivery channel impl on BOTH deliveryChannels and wcDeliveryChannels properties

`"deliveryChannel"` field is ignored by API , only `wcDeliveryChannel` is respected

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -540,7 +541,37 @@ public async Task Put_New_Asset_Requires_Origin()
         // assert
         response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
     }
+    
+    [Theory]
+    [InlineData("deliveryChannels")]
+    [InlineData("wcDeliveryChannels")]
+    public async Task Put_New_Asset_Supports_DeliveryChannels_Aliases(string deliveryChannelAlias)
+    {
+        var assetId = new AssetId(99, 1, nameof(Put_New_Asset_Supports_DeliveryChannels_Aliases));
```
donaldgray, line 550: The assetId here will need to be different across the requests, or it will be a Create then Update. Which _could_ give the illusion that it's fine when it could actually have created with first then ignored the second.

```cs
var assetId = new AssetId(99, 1, $"{nameof(Put_New_Asset_Supports_DeliveryChannels_Aliases)}-{deliveryChannelAlias}");
```

donaldgray (approved):
LGTM, only single test needs updated

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
griffri: Good point, I'll include this

tomcrane (changes requested):

tomcrane (approved):

PR #734, opened 2024-02-12
Add Hydra model for delivery channels

### LINKED ISSUE #717: Replace `string[] DeliveryChannels` with `DeliveryChannel[] DeliveryChannels`
Replace the `string[] DeliveryChannels` property on the Hydra Image API class in the DLCS with `DeliveryChannel[] DeliveryChannels` as per the new documentation. Nothing will be calling this any more. All old calls still get routed to `OldDeliveryChannels` via `.wcDeliveryChannels` on hydra JSON.

Leave that property intact on the Hydra class, but ALL use of it within the DLCS is now for removal, as the new behaviour is implemented. We'll come back to deal with this property in #714 

This is a very small task but is the prep for the full implementation work.

We also, during the full implementation, need to remove all usage of `ImageOptimisationPolicy` and `ThumbnailPolicy`.

We don't necessarily have to have a specific task to rip it all out before introducing the new DeliveryChannels. Better to replace bit by bit, so we can see what the old code is doing in places as we implement the new. But the old code will never more be executed, and by the end of the work, the fields and the code that uses them must be gone.

griffri:
Resolves #717

This PR adds a new `DeliveryChannel` Hydra class and replaces the `string[] DeliveryChannels` property in `HydraModel.Image` with `DeliveryChannel[] DeliveryChannels`. It also removes the `OldDeliveryChannels` property.

### DIFF
--- a/src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
+++ b/src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
@@ -33,6 +33,6 @@ public class DeliveryChannelClass: Class
     {
         SupportedOperations = CommonOperations.GetStandardResourceOperations(
             operationId, "Delivery Channel", Id,
-            "GET", "POST", "PUT");
+            "GET", "POST", "PUT", "PATCH", "DELETE");
     }
 }
--- a/src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
+++ b/src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
@@ -1,0 +1,67 @@
+﻿using Hydra;
+using Hydra.Model;
+using Newtonsoft.Json;
+
+namespace DLCS.HydraModel;
+
+[HydraClass(typeof(DeliveryChannelClass),
+    Description = "A delivery channel represents a way an asset on the DLCS can be served.",
+    UriTemplate = "/customers/{0}/defaultDeliveryChannels/{1}, /customers/{0}/spaces/{1}/defaultDeliveryChannels/{2}")]
+public class DeliveryChannel : DlcsResource
+{
+    public DeliveryChannel()
+    {
+        
+    }
+    
+    public DeliveryChannel(string baseUrl, int customerId, string deliveryChannelId, int? spaceId)
+    {
+        ModelId = deliveryChannelId;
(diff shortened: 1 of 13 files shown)


tomcrane:
From discussion

`DeliveryChannel` shoudl not have customerId, SpaceId or ModelId; it's only ever used inline in PUTS GETS and POSTs

JackLewis-digirati (approved):

src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
```diff
@@ -0,0 +1,38 @@
...
+using Newtonsoft.Json;
+
+namespace DLCS.HydraModel;
+
+[HydraClass(typeof(DeliveryChannelClass),
+    Description = "A delivery channel represents a way an asset on the DLCS can be served.",
+    UriTemplate = "")]
+public class DeliveryChannel : DlcsResource
+{
+    [RdfProperty(Description = "The name of the DLCS delivery channel this is based on.",
```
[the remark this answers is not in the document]
griffri: At this stage it should be fine for the descriptions to be somewhat basic, but I would be in support of including examples when we get around to reworking the documentation.

src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
```diff
@@ -0,0 +1,38 @@
...
+    public DeliveryChannelClass()
+    {
+        BootstrapViaReflection(typeof(DeliveryChannel));
+    }
+    
+    public override void DefineOperations()
+    {
+        SupportedOperations = CommonOperations.GetStandardResourceOperations(
+            operationId, "Delivery Channel", Id,
+            "GET", "POST", "PUT");
```
donaldgray, line 36: `PATCH`, `DELETE`

donaldgray (approved):

src/protagonist/DLCS.HydraModel/DeliveryChannel.cs
griffri: Fixed in https://github.com/dlcs/protagonist/pull/734/commits/d98c70f041bf7c5a7926533cf5c6ebf310ab5f3d

PR #740, opened 2024-02-15
Adding delivery channel RFC's for the database and av policies

### LINKED ISSUE #709: (RFC Spike) How do iiif-av policies become Elastic Transcoder settings?
https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/delivery-channels#iiif-av

In #616 we decided to use named _DLCS_ presets as the policyData.

```
   "policyData": "[ \"video-mp4-720p\" ]",
```

When Engine is given an asset with:

```
  "origin":  "s3://my-bucket/video-masters/huge-video.mpeg",
  "deliveryChannels": [
    {
      "@type": "vocab:DeliveryChannel",
      "channel": "iiif-av",
      "policy": "https://api.dlcs.io/customers/2/deliveryChannelPolicies/iiif-av/standard"
    }
  ]
```

... and `https://api.dlcs.io/customers/2/deliveryChannelPolicies/iiif-av/standard` looks like this:

```
{
     "@id": "https://api.dlcs.io/customers/2/deliveryChannelPolicies/iiif-av/standard",
     "@type": "vocab:DeliveryChannelPolicy",
     "displayName": "Default video transcode",
     "channel": "iiif-av",
     "policyData":  "[ \"video-mp4-720p\" ]",
     "policyModified": "2023-09-19T15:36:58.6023600Z"
 }
```

...then Engine _looks up_ "video-mp4-720p" (where?) and sees, in some form, "Transcode with "System preset: 'Mp4 HLS 720p'; Extension: 'mp4'" so it calls Elastic Transcoder with that information.

> (where?)

This could be a hard-coded dictionary or other data structure. It's not modifiable by users, they can't add new policy data keys, and any keys we add would need to be added to the list in the documentation.

A C# file in GitHub is as good a place to keep this lookup as a DB table.

JackLewis-digirati:
Resolves #709 

This PR contains a pair of RFC's used to help describe how the database works and how av policies are managed internally


docs/rfcs/014-delivery-channels-database.md
```diff
@@ -0,0 +1,200 @@
...
+  ]
+```
+
+There is also a one off DB-migration task, and that is to apply this new row creation in both tables to all EXISTING customers, so they get the data they would have if created with this mechanism already in place.
+
+These will act as a template and be copied (not referenced) for new customers when a new customer is created.
+
+## Creating new delivery channels
+
+customer creates custom av policy
```
donaldgray, line 88: `Customer`, all other sections have capet

docs/rfcs/014-delivery-channels-database.md
```diff
@@ -0,0 +1,200 @@
...
+```
+This produces a row in DeliveryChannelPolicies
+```
+     |--------- (constraint) ------------|
+| id | Customer | Channel | Name         | DisplayName               | System | PolicyCreated | PolicyModified | PolicyData            |
+|----|----------|---------|--------------|---------------------------|--------|---------------|----------------|-----------------------|
+| 1  | 20       | iiif-av | specific-mp4 | a specific policy for mp4 | 0      | (now)         | (now)          | [\"video-mp4-1080p\"] |
+```
+
+And there's now a DeliveryChannelPolicy resource at /customers/20/deliveryChannelPolicies/iiif-av/specific-mp4 that would look like this:
```
donaldgray, line 107: backtick url for easier readability?

docs/rfcs/014-delivery-channels-database.md
```diff
@@ -0,0 +1,200 @@
...
+But we like this policy so we want it to apply automatically on Space 5:
+
+```
+POST /{base}/customers/20/spaces/5/defaultDeliveryChannels
+{
+   "channel" : "iiif-av",
+   "policy": "https://api.dlcs.io/customers/20/deliveryChannelPolicies/iiif-av/specific-mp4",
+   "mediaType": "video/*"
+}
+```#
```
donaldgray, line 144: Use the `View file` function to view rendered markup - something is breaking it. I think it's this trailing `#`

docs/rfcs/015-iiif-av-delivery-channel-settings.md
```diff
@@ -0,0 +1,132 @@
...
+     "@id": "https://api.dlcs.io/customers/2/deliveryChannelPolicies/iiif-av/standard",
+     "@type": "vocab:DeliveryChannelPolicy",
+     "displayName": "Default video transcode",
+     "channel": "iiif-av",
+     "policyData":  "[ \"video-mp4-720p\" ]",
+     "policyModified": "2023-09-19T15:36:58.6023600Z"
+ }
+```
+
+...then Engine _looks up_ "video-mp4-720p" and sees, in some form, "Transcode with "System preset: 'Mp4 HLS 720p'; Extension: 'mp4'" so it calls Elastic Transcoder with that information.
```
donaldgray, line 40: Use backticks for readability? `"Transcode with "System preset: 'Mp4 HLS 720p'; Extension: 'mp4'"`

docs/rfcs/015-iiif-av-delivery-channel-settings.md
```diff
@@ -0,0 +1,132 @@
...
+  },
+```
+
+From this, the preset is used to kick off an ET `job` and then get pushed into buckets for output
+
+## Delivery Channel Proposal
+
+Firstly, as [discussed](https://github.com/dlcs/protagonist/issues/709), The `policyData` is an array of policies *which each link to a single system preset*, as opposed to combined values having multiple outputs.  These values are then looped through to create all the outputs that are required.  From an orchestrator perspective, nothing will have changed as these transcoder outputs will not change.
+
+Due to the above, the likely best place to put these values is a dictionary in appsettings, similar to how `TranscoderMappings` works.  This is because there are a limited number of presets we want to support (dev currently uses only 3), rather than all of them and that it is unlikely for these values to need to be changed regularly.  Additionally, `TranscoderMappings` should be deprecated, with a new setting created called `DeliveryChannelMappings`, which consists of key value pairs like the below:
```
donaldgray, line 105: The only issue I can see with storing everything in appSettings is that if a single Protagonist deployment wants to add a new one it would need to replicate all of them, or use indexes to append (I _think_ this would work).

E.g. if `appSettings.json` in Protagonist has all system provided settings:

```json
  "TimebasedIngest": {
    "DeliveryChannelMappings": {
      "video-mp4-480p": "System preset: Generic 480p 16:9",
      "video-webm-720p": "System preset: Webm 720p(webm)",
      "audio-mp3-128k": "System preset: Audio MP3 - 128k(mp3)"
    }
  },
```

and I want to use those, but also add `"audio-flac-128": "my-custom-flac"`, then, assuming I'm using envvars in deployment, I'd need either to list all system ones, plus my own:

```
"TimebasedIngest__DeliveryChannelMappings__video-mp4-480p": "System preset: Generic 480p 16:9",
"TimebasedIngest__DeliveryChannelMappings__video-webm-720p": "System preset: Webm 720p(webm)",
"TimebasedIngest__DeliveryChannelMappings__audio-mp3-128k": "System preset: Audio MP3 - 128k(mp3)",
"TimebasedIngest__DeliveryChannelMappings__audio-flac-128": "my-custom-flac"
```

or (maybe - assuming this works) specify an index to add this to the end:

```
"TimebasedIngest__DeliveryChannelMappings__3__audio-flac-128": "my-custom-flac"
```

the issue with ^ is that I'd need to be aware of any new presets being added. What I'd really want to set is just my own value

```
"TimebasedIngest__DeliveryChannelMappings__audio-flac-128": "my-custom-flac"
```

donaldgray (commented):
Add these to `docs/rfcs/readme.md`

docs/rfcs/015-iiif-av-delivery-channel-settings.md
JackLewis-digirati: On the appsettings AV stuff,  I did some testing locally, with the following:

appsettings.json:
```
  "TimebasedIngest": {
    "DeliveryChannelMappings": {
      "TEST1": "System preset: Web",
      "TEST2": "some value"
    }
  }
```

appsettings.development.json:
```
  "TimebasedIngest": {
    "DeliveryChannelMappings": {
      "TEST2": "overloaded value",
      "TEST3": "someValue 3"
    }
  },
```
launchsettings.json:

```
"environmentVariables": {
        "TimebasedIngest__DeliveryChannelMappings__TEST3": "overloaded env var",
        "TimebasedIngest__DeliveryChannelMappings__TEST4": "new env var"
      }
```

final value was `{"TEST1":"System preset: Web","TEST2":"overloaded value","TEST3":"overloaded env var","TEST4":"new env var"}` - so I think this is ok

docs/rfcs/014-delivery-channels-database.md
```diff
@@ -0,0 +1,201 @@
+# Delivery channel database design
+
+## Brief
+
+This RFC is related to the [table design for delivery channels](https://github.com/dlcs/protagonist/issues/618) to help show a table design for the additional tables required for the delivery channels work
+
+## Delivery Channel Policy Table
+
+![delivery channel policies](img/DeliveryChannels.png)
```
donaldgray, line 9: This image doesn't reflect the below image - it shows `iiif-img` channel having policyData which doesn't match with below table + description

docs/rfcs/015-iiif-av-delivery-channel-settings.md
```diff
@@ -0,0 +1,132 @@
...
+    "PipelineName": "dlcsspinup-timebased",
+    "TranscoderMappings": { // deprecated
+      "Wellcome Standard MP4": "System preset: Web"
+    }, 
+    "DeliveryChannelMappings": {
+      "video-mp4-480p": "System preset: Generic 480p 16:9",
+      "video-webm-720p": "System preset: Webm 720p(webm)",
+      "audio-mp3-128k": "System preset: Audio MP3 - 128k(mp3)"
+    }
+  },
```
tomcrane, line 118: What do you think about this being not json appsettings but a static class, with a Dictionary?

Yes, you need PRs and deployments to add new ones. But that isn't actually a big deal.
And it can be shared.

docs/rfcs/015-iiif-av-delivery-channel-settings.md
```diff
@@ -0,0 +1,132 @@
+# How are iiif-av settings going to be maintained in protagonist
```
donaldgray, line 1: I think it's worth adding brief details around what we spoke about this morn (how they're handled, engine source of truth etc).

donaldgray (approved):

Issue #726, opened 2024-02-06
Engine can process AV channel and its policies


tomcrane:
To be refined once RFC from #709 produced.

## Acceptance Criteria

* Update AppSetting to store mappings as per RFC
* Create endpoint to return known policyData values (keys of dictionary)
* Update AV processing to use these values to create ElasticTranscoder jobs + create transcoded output(s) in correct locations (behaviour as-is).

PR #748, opened 2024-02-28
Add AV settings to Engine

JackLewis-digirati:
This PR is related to #726 

It provides an additional endpoint used to provide valid `iiif-av` policies for use by the API

This is tracked by a sub-task [here](https://digirati.atlassian.net/browse/WDC-57)

### DIFF
--- a/src/protagonist/Engine/Ingest/IngestController.cs
+++ b/src/protagonist/Engine/Ingest/IngestController.cs
@@ -2,7 +2,9 @@
 using System.Text.Json;
 using DLCS.Model.Messaging;
 using Engine.Ingest.Models;
+using Engine.Settings;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
 
 namespace Engine.Ingest;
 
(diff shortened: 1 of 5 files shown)


src/protagonist/Engine/Ingest/IngestController.cs
```diff
@@ -51,6 +55,16 @@ public async Task<IActionResult> IngestAsset(CancellationToken cancellationToken
 
         return ConvertToStatusCode(message, result.Status);
     }
+    
+    /// <summary>
+    /// Retrieve allowed av options
+    /// </summary>
+    [HttpGet]
+    [Route("allowed-av")]
+    public IActionResult ReturnAllowedAvOptions()
```
griffri, line 64: `ReturnAllowedAvOptions` makes sense as a name, but would `GetAllowedAvOptions` be slightly clearer and in-line with other methods seen throughout the DLCS?

griffri (approved):
This looks fine, there's just one (admittedly nitpicky!) thing that's on my mind.

donaldgray (approved):

Issue #629, opened 2023-09-20
Engine handles thumbs and iiif-img delivery channel


tomcrane:
Handle `thumbs` and `iiif-img` delivery-channels. Thumb handling using new service as per https://github.com/dlcs/protagonist/blob/develop/docs/adr/0006-engine-imageserver.md

Incoming asset may have one or the other or both, but a single "unit of work" needs to assess the asset and process in a certain order. See the ADR for sequence diagrams outlining flow.

Engine is always making SOME thumbs for an image, a union of the secret policy #627 and the supplied policy; the supplied policy may be nothing so it's a union of the secret policy and an empty set.

# Acceptance Criteria

- Playwright tests exist to describe all permutations in the ADR
- Engine produces the correct derivatives for each permutation, the broad categories are:
   - Thumbs only. Sizes are union of those specified in policy and "unofficial" sizes. (if `thumbs` channel only).
   - A JP2 and the unofficial thumbs (if `iiif-img` channel only).
   - A JP2 and union of specified + "unofficial" sizes (if `iiif-img` and `thumbs` channels).
- Thumbnail sizes are determined by looking at the actual size of the JPEG returned from Cantaloupe (currently these are provided in JSON payload returned from Appetiser).
- Update https://github.com/dlcs/protagonist/blob/develop/docs/storage-keys.md with location for transient origin-image once decided.

See below for possible permutations as channel, origin-strategy, image-policy and image format can affect processing. This is taken from the ADR - note that this isn't the purpose of that ADR and was working notes helped to produce output.

| d-c             | ImageOptPolicy | Origin Type  | Origin Format | What happens in Engine                                                                                                                              |
| --------------- | -------------- | ------------ | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| iiif-img,thumbs | default        | http*        | *             | Download origin. Calls Appetiser to convert. Uploads converted to S3. Calls Cantaloupe to generate thumbs. Uploads thumbs to S3.                    |
| iiif-img,thumbs | default        | http*        | image/jp2     | Download origin. Calls Appetiser to convert (would be a no-op). Uploads converted to S3. Calls Cantaloupe to generate thumbs. Uploads thumbs to S3. |
| iiif-img,thumbs | default        | s3-optimised | *             | Download origin. Calls Appetiser to convert. Uploads converted to S3. Calls Cantaloupe to generate thumbs. Uploads thumbs to S3.                    |
[...]

Issue #621, opened 2023-09-20
Update logic for Orchestrator and Engine to determine if an Asset has a particular deliveryChannel


tomcrane:
Engine and Orchestrator need to identify if an Asset is available on a particular delivery-channel, using the new schema. This is currently an extension method that interogates the `string[]` on an Asset.

## Acceptance Criteria
* Engine can identify whether an Asset has a particular delivery channel to build a list of workers to process.
* Orchestrator can identify whether an Asset has a particular delivery channel to service delivery requests.
* Update/replace calls to existing `HasDeliveryChannel()` to use new/updated method.

tomcrane:
Read delivery channel information from the new tables - see https://docs.google.com/document/d/16hoPUHWO3Rh4Ko734GiNhdOHmRoE-w7AV4gEF561aPg/edit

tomcrane:
Further development of `hasDeliveryChannel` extension method

PR #755, opened 2024-03-06
Update engine to use an asset from the database, as opposed to the request

### LINKED ISSUE #622: Update Engine HydrateAssetPolicies() method for delivery channel model
Load full delivery channels information and update Asset model correctly. 

Everything in Engine should read the policies populated here rather than fetch again - load the asset, use EF include to populate the actual values of dc policies (e.g., thumb sizes) and then this is used from here on in Engine.

## Acceptance Criteria

* On receipt of sync or async method the Engine will fetch Asset from database, whether we receive fully hydrated object or not.
* This fetch includes all policyData that Engine will use for processing.

### LINKED ISSUE #726: Engine can process AV channel and its policies

JackLewis-digirati:
Resolves #622  

This PR updates engine ingest to use an asset from the database, compared to an asset from the request

### DIFF
--- a/src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
+++ b/src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
@@ -4,14 +4,18 @@ using DLCS.AWS.S3;
 using DLCS.AWS.SQS;
 using DLCS.Core.Caching;
 using DLCS.Core.FileSystem;
+using DLCS.Model;
+using DLCS.Model.Assets;
 using DLCS.Model.Auth;
 using DLCS.Model.Customers;
 using DLCS.Model.Policies;
 using DLCS.Model.Processing;
 using DLCS.Model.Storage;
 using DLCS.Repository;
+using DLCS.Repository.Assets;
 using DLCS.Repository.Auth;
 using DLCS.Repository.Customers;
+using DLCS.Repository.Entities;
 using DLCS.Repository.Policies;
 using DLCS.Repository.Processing;
 using DLCS.Repository.Storage;
@@ -118,6 +122,8 @@ public static class ServiceCollectionX
     public static IServiceCollection AddDataAccess(this IServiceCollection services, IConfiguration configuration)
         => services
             .AddScoped<IPolicyRepository, PolicyRepository>()
+            .AddScoped<IAssetRepository, AssetRepository>()
+            .AddScoped<IEntityCounterRepository, EntityCounterRepository>()
             .AddScoped<IEngineAssetRepository, EngineAssetRepository>()
             .AddScoped<ICustomerOriginStrategyRepository, CustomerOriginStrategyRepository>()
             .AddSingleton<ICredentialsRepository, DapperCredentialsRepository>()
--- a/src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
+++ b/src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
@@ -4,18 +4,14 @@ using DLCS.AWS.S3;
 using DLCS.AWS.SQS;
 using DLCS.Core.Caching;
 using DLCS.Core.FileSystem;
-using DLCS.Model;
-using DLCS.Model.Assets;
 using DLCS.Model.Auth;
 using DLCS.Model.Customers;
 using DLCS.Model.Policies;
 using DLCS.Model.Processing;
 using DLCS.Model.Storage;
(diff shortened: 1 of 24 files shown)


griffri (approved):

src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
```diff
@@ -118,6 +122,8 @@ public static IServiceCollection AddAssetIngestion(this IServiceCollection servi
     public static IServiceCollection AddDataAccess(this IServiceCollection services, IConfiguration configuration)
         => services
             .AddScoped<IPolicyRepository, PolicyRepository>()
+            .AddScoped<IAssetRepository, AssetRepository>()
+            .AddScoped<IEntityCounterRepository, EntityCounterRepository>()
```
donaldgray, line 126: Engine shouldn't require `IEntityCounterRepository` as it never touches entity-counters.

src/protagonist/Engine/Ingest/AssetIngester.cs
```diff
@@ -70,14 +73,22 @@ public Task<IngestResult> Ingest(LegacyIngestEvent request, CancellationToken ca
...
+            return new IngestResult(asset, IngestResultStatus.Failed);
+        }
+        
         // get any matching CustomerOriginStrategy 
-        var customerOriginStrategy = await customerOriginRepository.GetCustomerOriginStrategy(request.Asset, true);
+        var customerOriginStrategy = await customerOriginRepository.GetCustomerOriginStrategy(asset, true);
             
         // set Thumbnail and ImageOptimisation policies on Asset
-        await HydrateAssetPolicies(request.Asset);
+        await HydrateAssetPolicies(asset);
```
donaldgray, line 88: Does this ticket include updating what this method does? It currently loads full `ThumbnailPolicy` and `ImageOptimisationPolicy` data but those have been replaced

src/protagonist/Engine/Ingest/AssetIngester.cs
```diff
@@ -70,14 +73,22 @@ public Task<IngestResult> Ingest(LegacyIngestEvent request, CancellationToken ca
     /// <returns>Result of ingest operations</returns>
     public async Task<IngestResult> Ingest(IngestAssetRequest request, CancellationToken cancellationToken = default)
     {
+        var asset = await assetRepository.GetAsset(request.Asset.Id);
```
donaldgray, line 76: This could result in a cached read. Do we need control over what is loaded here (with `.Include()`)?

src/protagonist/Engine/Ingest/AssetIngester.cs
JackLewis-digirati: I don't think it was explicitly called out for removal, but might make sense to do it here - only reason I can see not to, is if it would be covered under the full "strip out old delivery channels" work, which I don't believe it is - so will get this updated

src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
JackLewis-digirati: it's used by `AssetRepository` to decrement `SpaceImages` and `CustomerImages` on delete - looking at this a bit deeper, it looks like there's a method in `EngineAssetRepository` which does something similar to what I need, so will look to remove `AssetRepository`

JackLewis-digirati: so I've managed to do this, but probably needs a discussion as there's a small change that has to be made to the `EngineAssetRepository` where it changes a `bool` that I'm not sure on the impact of

tomcrane (approved):

src/protagonist/Engine/Ingest/AssetIngester.cs
JackLewis-digirati: actually, I've just looked at this and might make more sense in either #621 or #629 as removing them breaks the `AppetiserRequestModel` from sending values such as `kdu_max`

JackLewis-digirati: stuck a note on #629

PR #762, opened 2024-03-13
Validate AV policyData in API

### LINKED ISSUE #751: API validates AV policyData using known list from Engine
#726 introduces a new endpoint to return a list of known AV policyData values. This list should be used to validate incoming deliveryChannelPolicies for AV channel.

## Acceptance Criteria

* API validates deliveryChannelPolicies for AV against known list.
* List is fetched and cached for long time.
  * Engine is not called on startup to reduce dependency of Engine and API running at the same time.
* If values provided is unknown, return a 400.
  * Includes Engine returning `[]`
* If error fetching from Engine, return a 500.

griffri:
Resolves #751 and WDC-54

This PR gives `DeliveryChannelPolicyValidator` the ability to validate AV policyData against a cached list of transcoder policies retrieved from the new `/allowed-av` endpoint added to Engine in #726.

It also introduces a new change to the API `appSettings.json`: `EngineDirectIngestUri` has now been replaced with `EngineRoot` - `EngineClient` calls `/allowed-av`, `/image-ingest` and `/asset-ingest` using this as the base URL.


src/protagonist/API.Tests/Integration/DeliveryChannelTests.cs
```diff
@@ -18,13 +26,38 @@ namespace API.Tests.Integration;
...
+        "video-webm-720p",
+        "audio-mp3-128k"
+    };
     
     public DeliveryChannelTests(DlcsDatabaseFixture dbFixture, ProtagonistAppFactory<Startup> factory)
     {
         dbContext = dbFixture.DbContext;
-        httpClient = factory.ConfigureBasicAuthedIntegrationTestHttpClient(dbFixture, "API-Test");
+        httpHandler = new ControllableHttpMessageHandler();
+        httpClient = factory
```
donaldgray, line 44: There's an overload of `ConfigureBasicAuthedIntegrationTestHttpClient` that takes a delegate to customise the webAppFactory, saves repeating the Auth stuff.

```cs
httpClient = factory.ConfigureBasicAuthedIntegrationTestHttpClient(dbFixture, "API-Test",
            f => f.WithLocalStack(storageFixture.LocalStackFixture));
```

src/protagonist/API/Features/DeliveryChannels/AvChannelPolicyOptionsRepository.cs
```diff
@@ -0,0 +1,30 @@
...
+        this.appCache = appCache;
+        cacheSettings = cacheOptions.Value;
+        this.engineClient = engineClient;
+    }
+
+    public async Task<IReadOnlyCollection<string>?> RetrieveAvChannelPolicyOptions()
+    {
+        const string key = "avChannelPolicyOptions";
+        
+        return await appCache.GetOrAdd(key, () => engineClient.GetAllowedAvPolicyOptions(),
```
donaldgray, line 27: If we receive a `null` or `empty` value that could signify an issue, either config or service down. If that's the cause would it be worth caching a null object, or caching for a shorter duration? By caching for `Long` if the Engine was down and came back up it would take a while for that to be reflected - or would need the cache to be flushed / service restarted to pickup.

src/protagonist/DLCS.Repository.Tests/Messaging/EngineClientTests.cs
```diff
@@ -28,7 +28,10 @@ public class EngineClientTests
     public EngineClientTests()
```
donaldgray, line 28: Worth adding a test for new method here?

src/protagonist/DLCS.Repository/Messaging/EngineClient.cs
```diff
@@ -26,7 +28,10 @@ public class EngineClient : IEngineClient
     private readonly HttpClient httpClient;
     private readonly ILogger<EngineClient> logger;
     private readonly DlcsSettings dlcsSettings;
-
+    
+    private const string EngineIngestEndpoint = "asset-ingest";
+    private const string LegacyEngineIngestEndpoint = "image-ingest";
```
donaldgray, line 33: We do not need to handle this. The whole Legacy handling will be removed in the future but don't introduce new code to handle it as it's a path that's never called.

src/protagonist/API/Features/DeliveryChannels/AvChannelPolicyOptionsRepository.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using DLCS.Core.Caching;
+using DLCS.Model.DeliveryChannels;
+using DLCS.Repository.Messaging;
+using LazyCache;
+using Microsoft.Extensions.Options;
+
+namespace API.Features.DeliveryChannels;
+
+public class AvChannelPolicyOptionsRepository : IAvChannelPolicyOptionsRepository
```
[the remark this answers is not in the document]
donaldgray: `*Client` classes directly interact with external services so don't think that suits (`EngineClient`, `AppetiserClient` etc).

src/protagonist/API/Features/DeliveryChannels/AvChannelPolicyOptionsRepository.cs
donaldgray: Following on from

>would it be worth caching a null object

I don't think `null` values are cached so it would always be a cache miss in that instance.

src/protagonist/DLCS.Repository/Messaging/EngineClient.cs
```diff
@@ -134,6 +142,22 @@ public class EngineClient : IEngineClient
...
+        {
+            var response = await httpClient.GetAsync("allowed-av", cancellationToken);
+            return await response.Content.ReadFromJsonAsync<IReadOnlyCollection<string>>(
+                cancellationToken: cancellationToken);
+            
+        }
+        catch(Exception ex)
+        {
+            logger.LogError("Failed to retrieve allowed iiif-av policy options from Engine", ex);
+            return null;
```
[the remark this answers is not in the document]
donaldgray: This is a legit error but if Engine is down we don't want to repeatedly try and hit it. Caching null value for a short duration would prevent that, another option would be to use a circuit-breaker on the httpClient.

donaldgray (approved):

JackLewis-digirati (approved):

PR #772, opened 2024-03-22
Remove deprecated functionality (LegacyIngestEvent & InitialOrigin)

### LINKED ISSUE #759: Remove LegacyIngestEvent and InitialOrigin functionality
The following functionality is no longer required and can be removed:

* `LegacyIngestEvent` and all usages. This was used for testing Deliverator-API -> Protagonist-Engine interactions to validate old/new behaviour. At one point there was a possibility of rolling out Protagoinst-Engine while still using Deliverator-API but this won't happen now. `UseLegacyEngineMessage` appSetting controls whether API uses this format so can also be removed.
* `"InitialOrigin"` usage in Portal. This field was removed in #622/#750 but is still available in CSV upload.

griffri:
Resolves #759

`InitialOrigin` is no longer expected by the Portal batch uploader when reading a CSV file. The sample file available for download has also been updated to reflect this change.

Legacy payload support and its associated tests and configuration have been removed from Engine.


donaldgray (approved):
LGTM - there's still a mention of legacy handling in engine readme - that should be removed too.

donaldgray (approved):

PR #773, opened 2024-03-22
Update CleanupHandler to use `ImageDeliveryChannel` delivery channels

### LINKED ISSUE #691: Cleanup Handler use of Delivery Channels
This may be a non-issue in the future but the CleanupHandler uses DeliveryChannel to work out which `/iiif-img/`, `/thumbs/`, `/iiif-av/`, `/file/` paths need invalidating but also uses AssetFamily to work out if iiif-manifest needs invalidation. These should be aligned to use deliveryChannel, falling back to AssetFamily if DC is empty.

## Acceptance Criteria

* CleanupHandler uses values from `ImageDeliveryChannels` table rather than `Asset.DeliveryChannels` to work out what needs deleted.
* CleanupHandler doesn't use AssetFamily to drive any decisions on what is to be deleted.

griffri:
Resolves #691

This PR updates CleanupHandler to use the new `ImageDeliveryChannel` based `deliveryChannel` values to determine what needs to be invalidated.

### DIFF
--- a/src/protagonist/CleanupHandler/AssetDeletedHandler.cs
+++ b/src/protagonist/CleanupHandler/AssetDeletedHandler.cs
@@ -102,7 +102,7 @@ public class AssetDeletedHandler : IMessageHandler
             $"{customerName}/{asset.Id.Space}/{asset.Id.Asset}"
         };
         
-        if (asset.DeliveryChannels.IsNullOrEmpty())
+        if (asset.ImageDeliveryChannels.IsNullOrEmpty())
         {
             logger.LogDebug("Received message body with no 'deliveryChannels' property. {@Request}", 
                 asset);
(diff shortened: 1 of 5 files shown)


src/protagonist/CleanupHandler/AssetDeletedHandler.cs
```diff
@@ -102,26 +102,18 @@ private async Task<bool> InvalidateContentDeliveryNetwork(Asset asset, string cu
             $"{customerName}/{asset.Id.Space}/{asset.Id.Asset}"
         };
         
-        if (asset.DeliveryChannels.IsNullOrEmpty())
-        {
-            logger.LogDebug("Received message body with no 'deliveryChannels' property. {@Request}", 
-                asset);
-        }
-        else
+        if (!asset.ImageDeliveryChannels.IsNullOrEmpty())
```
donaldgray, line 105: Did you run the code end to end as it's called from API? I don't think this property will be populated.

donaldgray (approved):

PR #776, opened 2024-03-26
Allow API to emulate old delivery channel properties

### LINKED ISSUE #714: Emulation layer to convert `.wcDeliveryChannels` at the API level
At the DLCS API level, convert incoming `OldDeliveryChannels`, `ImageOptimisationPolicy` and `ThumbnailPolicy` setter calls (from Wellcome) into equivalent new Delivery Channel resources, and convert the getters. 

API, Engine, Orchestrator all use the full **new DeliveryChannels as per docs**, but the API will translate the incoming old property settings from Wellcome into their new equivalents, and emulate the old properties. 

_This can't be just simple getter and setter interception as there are three fields that need to be evaluated together; has to happen on persistence._

However, apart from `use-original` all normal IOP and ThumbnailPolicy on assets is going to be `null` anyway, and we could reject anything that isn't for a short time.

## Example 1

Incoming Asset from Wellcome (the two nulls would actually be omitted, just making it clearer here):

```
{
    "origin": "...my.jp2",
    "wcDeliveryChannels": ["iiif-img"],
    "imageOptimisationPolicy": "use-original",
    "thumbnailPolicy": null, 
}
```

In old behaviour this means just allocate the default policies, and it also means make thumbnails (we don't have a separate thumbs channel in current Wellcome). So "iiif-img" on its own in the old model implies two distinct channels in the new model.

In order to emulate this, we require _ahead of time_ that we have two DeliveryChannelPolicies (DCPs) to reference:

https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/delivery-channels#policy

 A "system" default DCP called `use-original` (that is returned in compacted form in the API, not a URI), and a thumbnail policy belonging to this customer called "standard" whose policyData is the 1024...400...200...100 sizes.

Then, for this specific asset creation, we would:

- Create 1 row in ImageDeliveryChannels, linking to the `use-original` DCP
- Create 1 row in ImageDeliveryChannels, linking to the thumbnail "standard" DCP

If we then ask the API for the asset just created, we'd get back:

```
{
    "wcDeliveryChannels": ["iiif-img"],
    "imageOptimisationPolicy": null,
    "thumbnailPolicy": null, 
    "deliveryChannels": [
       {
         "@type": "vocab:DeliveryChannel",
         "channel": "iiif-img",
         "policy": "use-original"
       },
       {
         "@type": "vocab:DeliveryChannel",
         "channel": "thumbs",
         "policy": "https://api.dlcs.io/customers/2/deliveryChannelPolicies/thumbs/standard"
       }
   ]
}
```

## Example 2

Incoming **video** Asset from Wellcome (the two nulls would actually be omitted, just making it clearer here):

```
{
    "origin": "...my.mp4",
    "wcDeliveryChannels": ["iiif-av", "file"],
    "imageOptimisationPolicy": null,
    "thumbnailPolicy": null, 
}
```

In the old model, the `null` IOP would become the `video-max` policy. We're going to use preset-like behaviour in the new model too.

In order to emulate this, we require _ahead of time_ that we have two DeliveryChannelPolicies (DCPs) to reference:

 A "system" default DCP called `none` for the `file` channel (that is returned in compacted form in the API, not a URI), and an `iiif-av` policy belonging to this customer called "default-video" whose policyData is the preset key video-mp4-720p (i.e., has a `policyData` field with the escaped JSON string value `"[ \"video-mp4-720p\" ]"` )

Then, for this specific asset creation, we would:

- Create 1 row in ImageDeliveryChannels, linking to the iiif-av "default-video" DCP
- Create 1 row in ImageDeliveryChannels, linking to the `none` DCP

If we then ask the API for the asset just created, we'd get back:

```
{
    "wcDeliveryChannels": ["iiif-av", "file"],
    "imageOptimisationPolicy": null,
    "thumbnailPolicy": null, 
    "deliveryChannels": [
       {
         "@type": "vocab:DeliveryChannel",
         "channel": "iiif-av",
         "policy": "https://api.dlcs.io/customers/2/deliveryChannelPolicies/iiif-av/default-video"
       },
       {
         "@type": "vocab:DeliveryChannel",
         "channel": "file",
         "policy": "none"
       }
   ]
}
```

## Acceptance Criteria

New boolean feature flag which will default to off/disabled and will:
* If enabled, attempt transformation from `wcDeliveryChannel`
* If disabled, non-null `ImageOptimisationPolicy` and `ThumbnailPolicy` values result in a 400 BadRequest.

If attempting transformation:
* `wcDeliveryChannel` of `"iiif-img"` results in:
  * `"iiif-img"` ImageDeliveryChannel. `policy = "imageOptimisationPolicy" == "use-original" ? "use-original" : "default"`
  * `"thumbs"` ImageDeliveryChannel. Always with default policy.
* `wcDeliveryChannel` of `"file"` results in `"file"` delivery-channel
* `wcDeliveryChannel` of `"iiif-av"` results in `"iiif-av"` delivery-channel using default policy depending on whether A or V (check value from other tickets).

Some requests will specify multiple `wcDeliveryChannel` values

griffri:
Resolves #714
Links WDC-47

This PR adds a new `OldHydraDeliveryChannelsConverter` class capable of converting `wcDeliveryChannels` into the new `ImageDeliveryChannel` format.

This converter is called when creating a new asset via PUT, but only if the `EmulateOldDeliveryChannelProperties` flag is enabled in ApiSettings. Otherwise, providing `ImageOptimisationPolicy` or `ThumbsPolicy` returns 400.


src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -687,19 +685,19 @@ public async Task Put_NewAudioAsset_Creates_Asset()
...
         response.Headers.Location.PathAndQuery.Should().Be(assetId.ToApiResourcePath());
-        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
+        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels)
+            .ThenInclude(dc => dc.DeliveryChannelPolicy).Single(x => x.Id == assetId);
         asset.Id.Should().Be(assetId);
         asset.MaxUnauthorised.Should().Be(-1);
         asset.ImageDeliveryChannels.Count.Should().Be(1);
         asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == "iiif-av" &&
-                                                                x.DeliveryChannelPolicyId == 5);
+                                                                x.DeliveryChannelPolicy.Name == "default-audio");
```
donaldgray, line 694: I think I also changed this line in PR that was just merged to `develop` so you may have a few conflicts.

src/protagonist/API.Tests/Integration/ModifyAssetWithoutOldDeliveryChannelEmulationTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Test.Helpers.Integration;
+using Test.Helpers.Integration.Infrastructure;
+
+namespace API.Tests.Integration;
+
+[Trait("Category", "Integration")]
+public class ModifyAssetWithoutOldDeliveryChannelEmulationTests : IClassFixture<ProtagonistAppFactory<Startup>>
```
donaldgray, line 15: Since this flag will be shortlived (it's just for cutover or wellcome deployment) it might make sense to make this class `ModifyAssetWithOldDeliveryChannelEmulationTests` so that we can just drop the entire class when the flag is removed, rather than deleting tests from main `ModifyAssetTests` and moving these over.

donaldgray (changes requested):
LGTM but need to handle `PATCH` image requets and POST to `/customers/{customer}/queue` too.

src/protagonist/API.Tests/Features/Queues/Validation/QueuePostValidatorTests.cs
```diff
@@ -11,11 +11,15 @@ namespace API.Tests.Features.Queues.Validation;
...
+    
     public QueuePostValidatorTests()
     {
-        var apiSettings = new ApiSettings { MaxBatchSize = 4 };
-        sut = new QueuePostValidator(Options.Create(apiSettings));
+        var apiSettingsA = new ApiSettings { MaxBatchSize = 4 };
+        sut = new QueuePostValidator(Options.Create(apiSettingsA));
+        
+        var apiSettingsB = new ApiSettings { EmulateOldDeliveryChannelProperties = true};
+        sutWithOldDcEmulation = new QueuePostValidator(Options.Create(apiSettingsB));
```
donaldgray, line 22: Where there are changes in SUT other tests use a `GetSut()` method that takes parameters to control how it is configured.

src/protagonist/API/Features/Queues/CustomerQueueController.cs
```diff
@@ -142,6 +151,21 @@ private  void UpdateMembers(int customerId, IList<DLCS.HydraModel.Image>? member
         }
     }
 
+    /// <summary>
+    /// Converts WcDeliveryChannels (if set) to DeliveryChannels for a list of assets
+    /// </summary>
+    /// <param name="members">The assets to update</param>
+    private void ConvertOldDeliveryChannelsForMembers(IList<DLCS.HydraModel.Image>? members)
+    {
+        if (members == null) return;
```
donaldgray, line 160: `members.IsNullOrEmpty()` ?

src/protagonist/API/Features/Image/ImageController.cs
```diff
@@ -162,6 +178,19 @@ public async Task<IActionResult> GetImage(int customerId, int spaceId, string im
             return this.HydraProblem("Delivery channels are disabled", assetId.ToString(), 400, "Bad Request");
         }
         
+        if (apiSettings.EmulateOldDeliveryChannelProperties && 
+            hydraAsset.WcDeliveryChannels != null)
```
donaldgray, line 182: Line 175 uses `!hydraAsset.WcDeliveryChannels.IsNullOrEmpty()` but this uses `hydraAsset.WcDeliveryChannels != null` - they should use the same logic? Since they both use this same check can they be combined?

donaldgray (approved):

PR #778, opened 2024-03-27
Allow DeliveryChannels to be modified

### LINKED ISSUE #764: Add support for changing an existing asset's `deliveryChannels` + implement `"channel":"none"` functionality
The `none` channel described in the documentation is currently validated by `HydraImageValidator` but isn't functioning.

> This resource is only valid on asset.deliveryChannels, and only when it is the one and only member. It is a marker to let you be explicit that the asset has no delivery channels, rather than that the platform should attempt to populate delivery channels from defaults.
> For updates, empty "deliveryChannels": [] is invalid. To explicitly remove all channels, use the channel: none marker channel as above.

The ability to update the delivery channels of an existing asset will need to be implemented as part of this task.

## Acceptance Criteria:

- [ ] **An asset's delivery channels can be updated**
  - [x] `deliveryChannels` is **required** when updating an asset via `PUT`, otherwise return 400 Bad Request if `[]` or null.
  - [x] `deliveryChannels` is optional when updating via PATCH - specifying `[]` should return 400 Bad Request
  - [ ] `deliveryChannels` changes are reflected in DB
  - [ ] Deleting an asset should clear the `deliveryChannels` that were assigned to it from the DB
  - [ ] Any changes to an asset's `deliveryChannels` must be appropriate for its mediaType (The `IsChannelValidForMediaType` helper should be used here)
  - [ ] Tests will need to be written for the above
- [ ] **The `none` delivery channel is functional**
  - [ ] Creating a new asset with a delivery channel of type `none` should prevent it from being automatically populated with the defaults for that space/customer
  - [ ] Updating an asset's `deliveryChannels` with a channel of type `none` through PUT or PATCH should wipe any previously assigned delivery channels
  - [ ] Tests will need to be written for the above

donaldgray:
Fixes #764 / WDC-60

## Main Change

As detailed in related ticket, the main change here is to support updating DeliveryChannels (via PUT, PATCH or batch operation).

Refactored `AssetProcessor`, previously this handled creation of DeliveryChannels - either setting from default or finding full policies from DB. This logic was moved to a new class, `DeliveryChannelProcessor`, which will:
* Add default DeliveryChannels as specified in DB - only if creation and no `"deliveryChannels"` specified.
* Add new DeliveryChannels specified in payload if not present in DB.
* Remove existing DeliveryChannels if ommitted from payload and present in DB.
* Update `deliveryChannelPolicyId` for existing ImageDeliveryChannel if policy changed for same channel (e.g. `"iiif-img"` from `"default"` -> `"iiif-img"`)

## Related Refactoring

Refactored `ApiAssetRepository`. Previously this was a decorator around basic `AssetRepository` which shared the `AssetRepositoryCachingBase` abstract base class with `DapperAssetRepository`. This setup no longer made sense and by refactoring it allowed the save and delete functionality in `ApiAssetRepository` to be simplified. Get operations are now specific for API usage and can be controlled by 2 bool params (`bool forUpdate = false, bool noCache = false)`

Updated `DeliveryChannelPolicyRepository` and `DefaultDeliveryChannelRepository` to use `async` operations when interacting with database. Also moved these and `AvChannelPolicyOptionsRepository` to a `.DataAccess` namespace to ease groking in solution explorer.

Extracted hardcoded known DeliveryChannelPolicyIds to `KnownDeliveryChannelPolicies` to make referencing in tests etc easier.


src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
```diff
@@ -0,0 +1,209 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace API.Features.Image.Ingest;
+
+public class DeliveryChannelProcessor
+{
+    private readonly IDefaultDeliveryChannelRepository defaultDeliveryChannelRepository;
+    private readonly IDeliveryChannelPolicyRepository deliveryChannelPolicyRepository;
+    private readonly ILogger<DeliveryChannelProcessor> logger;
+    private const string None = "none";
```
JackLewis-digirati, line 16: would this be better as `NoneChannel`? could possibly get confused as meaning the policy with the same name

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private void AddExplicitNoneChannel(Asset asset)
+    {
+        logger.LogTrace("assigning 'none' channel for asset {AssetId}", asset.Id);
+        var deliveryChannelPolicy = deliveryChannelPolicyRepository.RetrieveDeliveryChannelPolicy(asset.Customer,
+            AssetDeliveryChannels.None, None);
+
+        // "none" channel can only exist on it's own so remove any others that may be there already prior to adding
+        asset.ImageDeliveryChannels.Clear();
```
JackLewis-digirati, line 184: does this mean that if you had `["file","none"]`, the `file` channel would be dropped silently?

JackLewis-digirati (approved):

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
donaldgray: Updated name, this is policy rather than channel (which highlights the point you made perfectly!) so named `NonePolicy`

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
donaldgray: No, this method can only be entered if there is a single delivery-channel and it's `"none"` (validation will catch this higher up too) so in this case `asset.ImageDeliveryChannels.Clear()` is just removing any existing existing IDC records from database (e.g. if this was `["file", "iiif-img", "thumbs"]` it'll remove them and them add the `"none"` record).

griffri (approved):

PR #779, opened 2024-03-28
RFC for asset metadata

### LINKED ISSUE #767: Write RFC for managing Asset/DeliveryChannel metadata
Ticket #631 requires accessing thumbnail sizes in bulk. This would involve making bulk requests to S3 to read the data which would be inefficient at scale.

To counter this we need to investigate a way of storing metadata for Asset/DeliveryChannel in the database. Initially this will be storing available thumbnail sizes but could be extended in the future to store:

* Thumbnails: generated sizes - both system and requested thumbs. Open and Auth thumbs are currently in json in S3. Would we want to store in S3 _and_ database, or just database? Store open and auth?
  * Should `OrchestrationAsset` read from DB rather than S3 (saves an S3 GET request).
* Timebased: Generated transcode types and output locations. 
* File: do we store a copy of the original file? If so, where.
* Image: do we store a copy of the file. Is it original (`use-original`) or a transcode?
* Adjuncts: what is stored where (future).

All of the above can be calculated from ImageDeliveryChannels + Engine processing but this can act as an easy lookup to those values. The above metadata would also help with: 
* Cleaning up no-longer used manifests (#430)
* Generating full single-item manifest that contain all available representations of an asset (not just image services - see #488)

Questions:
* Assuming this is in a separate table, what is that called? `AssetMetadata` or `DeliveryChannelMetadata` (if we store adjunct data is this classifiable as DC related?
* How is the data modelled - column per data type (`File`, `Image`, `Thumbs`) or a JSON blob?

donaldgray:
Closes #767 / WDC-61

RFC exploring option for storing Asset metadata

### DIFF
--- a/docs/rfcs/016-asset-metadata.md
+++ b/docs/rfcs/016-asset-metadata.md
@@ -1,0 +1,87 @@
+# Storing Asset Metadata
+
+## Executive Summary
+
+Create a new `AssetApplicationMetadata` table to store metadata about an Asset for internal use only. This will have a flexible schema to be able to store whatever shape of data is required for an asset.
+
+## Motivation
+
+Ticket [#631](https://github.com/dlcs/protagonist/issues/631) introduces the need to read available image thumbnail sizes at scale. 
+
+Currently 'at scale' operations (e.g. projecting NamedQuery to IIIF Manifest) are done by calculating the available sizes per image on the fly. This is done by using the width + height of the image and relevant thumbnail policy. 
+
+However, now that thumbnail sizes are defined as [IIIF ImageApi Size parameters](https://iiif.io/api/image/3.0/#42-size) this is no longer a viable option as it would involve fairly complex size calculations and could fall foul of off-by-one rounding issues, as seen in the past. To negate this we now use an ImageServer to generate the thumbnails and store the actual sizes of those thumbnails (see [ADR 0006 - Engine ImageServer](../adr/0006-engine-imageserver.md)).
+
+The generated thumbnail sizes are stored in `s.json`, a json file in S3 (see [RFC 001 - Thumbs](001-thumbnails.md)). This is quick to read and parse when handling single asset requests but it would be very inefficient to do so at scale. 
+
+This RFC suggests at an alternative approach to storing the thumbnail sizes for an image.
+
+## Proposed Implementation
+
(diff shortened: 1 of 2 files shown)


docs/rfcs/016-asset-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+```cs
+var assetWithThumbs = dbContext.Images
+                .Include(i => i.AssetApplicationMetadata.Where(m => m.MetadataType == "ThumbSizes"))
+                .Single(i => i.Id == assetId);
+```
+
+## Outstanding Questions
+
+* For `Asset` object - would we use polymorphic deserialisation? How complex would that be? Would storing on object as a `string` be okay, deserialisation as required (rather than on way out of DB)
+* Should we store sizes in s3 as well as db?
```
tomcrane, line 92: I think so, also for making thumb serving independent of DB schema, self-contained.

docs/rfcs/016-asset-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+* `MetadataValue` is a `jsonb` value storing relevant data as JSON. In most cases I think we would always want to read the entire object but it could be useful to have efficient querying afforded by `jsonb` (e.g. to get `"o"`pen thumbs only). This querying has support in [npgsql](https://www.npgsql.org/efcore/mapping/json.html).
+* `AssetId` and `MetadataType` would be composite key.
+
+#### Sample Data
+
+Example data for an asset could be:
+
+| AssetId | MetadataType | MetadataValue                                                     |
+| ------- | ------------ | ----------------------------------------------------------------- |
+| 1/2/foo | ThumbSizes   | `{"o": [[200,127],[100,64]],"a": [[1024,651], [400,254]]}`        |
```
tomcrane, line 73: Not sure of the details but might there be a reason to store the thumb sizes differently from s3 here? e.g., some sort of query by size. This layout suits the needs of thumb serving but may not be the right shape for other uses (although it might be).

docs/rfcs/016-asset-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+
+The suggested schema for the table should be flexible.
+
+| Column        | Type        | Description                         |
+| ------------- | ----------- | ----------------------------------- |
+| AssetId       | text        | AssetId this is for                 |
+| MetadataType  | text        | Identifier for the type of metadata |
+| MetadataValue | jsonb       | JSON object of values for type      |
+| {audit-cols}  | timestamptz | Created/updated dates               |
+
```
tomcrane, line 61: (for later) This table might store information about the use of assets in IIIF Manifests.

tomcrane (approved):

PR #800, opened 2024-04-11
Treat region parameters that cover the entire image as 'full' in Orchestrator

### LINKED ISSUE #781: Detect IIIF region parameter that represents full image
The Orchestrator has special handling for `/full/` [IIIF region parameter](https://iiif.io/api/image/3.0/#41-region) to allow for more efficient serving of images using pre-generated thumbnails or via special-server to save load on the tile handling image-server.

We should add logic to Orchestrator to detect IIIF region parameters that don't specify `/full/` but the region represent the full image. The cached `OrchestrationAsset` contains enough information to calculate this.

Do we also need to update thumbs handler to recognise these as full, or is that already handled in calculations?

Region parameters that could represent `/full/` are:
* `/x,y,w,h/` format requests where x + y are 0 and w + h match those of the source image. E.g. for a 2234x5466 image this would be `0,0,2234,5466`.
* `/square/` format requests where the origin image is already square (`Asset.Width` == `Asset.Height`).

griffri:
Resolves #781

This PR updates `ImageRequestHandler` to treat IIIF image requests that request a region covering the entire image (e.g `0,0,512,512` for a 512x512 image) as `full`. It also includes tests that verify this behaviour.

### DIFF
--- a/src/protagonist/Orchestrator.Tests/Features/Images/ImageRequestHandlerTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Features/Images/ImageRequestHandlerTests.cs
@@ -344,6 +344,32 @@ public class ImageRequestHandlerTests
         result.Path.Should().Be("thumbs/2/2/test-image/0,0,512,512/256,256/0/default.jpg");
     }
     
+    [Fact]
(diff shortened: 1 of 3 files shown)


donaldgray (approved):
LGTM

JackLewis-digirati (approved):

PR #817, opened 2024-04-22
Ensure s.json ordering

donaldgray:
Main change in PR is an update to `ThumbCreator` to reintroduce behaviour to ensure thumb sizes in `s.json` are ordered from largest -> smallest (determined by longest edge as we are not supporting thumb sizes that mutate aspect ratio). This is the single point where `s.json` and `AssetApplicationMetadata` are written so removes need for any callers to reorder. e.g. 

`ThumbnailCalculator` expects sizes to be ordered and without this it doesn't properly identify larger/smaller sizes for Orchestrator optimisations.

Second change was to log a warning in `CantaloupeThumbsClient` if the returned MaxDimension does not match the expected MaxDimension.

### DIFF
--- a/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
+++ b/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
@@ -96,4 +96,30 @@ public class CantaloupeThumbsClientTests
         // Assert
         thumbs.Should().HaveCount(0);
     }
+    
+    [Fact]
+    public async Task GenerateThumbnails_Ignores400_AndProcessesRest()
+    {
+        // Arrange
+        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsNothing_WhenCantaloupeReturns400));
+        var context = IngestionContextFactory.GetIngestionContext(assetId: assetId.ToString());
+
+        var thumbSizes = new List<string> { "!1024,1024", "!400,400" };
+        
+        // first size is 400, then 200 after
+        httpHandler.SetResponse(new HttpResponseMessage(HttpStatusCode.BadRequest));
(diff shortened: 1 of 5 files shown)


src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
```diff
@@ -96,4 +96,30 @@ public async Task GenerateThumbnails_ReturnsNothing_WhenCantaloupeReturns400()
         // Assert
         thumbs.Should().HaveCount(0);
     }
+    
+    [Fact]
+    public async Task GenerateThumbnails_Ignores400_AndProcessesRest()
+    {
+        // Arrange
+        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsNothing_WhenCantaloupeReturns400));
```
JackLewis-digirati, line 104: think `nameof` needs updating

griffri (approved):

src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
```diff
@@ -96,4 +96,30 @@ public async Task GenerateThumbnails_ReturnsNothing_WhenCantaloupeReturns400()
...
+        var context = IngestionContextFactory.GetIngestionContext(assetId: assetId.ToString());
+
+        var thumbSizes = new List<string> { "!1024,1024", "!400,400" };
+        
+        // first size is 400, then 200 after
```
JackLewis-digirati, line 109: is this correct? the `thumbSizes` looks like 1024, then 400

JackLewis-digirati (approved):

src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
donaldgray: Will clarify, it's statusCodes.

PR #832, opened 2024-04-26
Adding initial asset modified cleanup rfc

JackLewis-digirati:


donaldgray (commented):
It's not explicitly mentioned but cleanup logic will need to take into account CustomerOriginStrategy, policy for image (`use-original`/`default`) and other available channels (i.e. it can't just look at the delivery-channels individually).

A thought when reading: How would this work during period of mass reingestion? ie if 10K..100K..1M..10M items are queued - items added last in the queue could take hours/days to be picked up by Engine. The max value for `delay_seconds` is 15mins so everything will be first picked up pretty quickly - assuming Engine is still in flight they wouldn't be acted on. As mentioned we can use `visibility_timeout` (max 12 hours) to ensure it's not picked up again for a while _but_ there are limits on the [number of in-flight messages](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#inflight-messages). The limit is large and we're using long-polling which means we wouldn't receive an error but would receive empty responses until there are less messages in-flight. Do we need to bear that in mind for this doc? Even if it's a backup solution to manually disable asset-cleanup if we know a mass ingest will happen.

General comment, there's inconsistent capitalisation + use of full stops

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,127 @@
...
+  - There's already logic to handle batch sending of notifications, just need to get list of messages to send.
+- It's possible to send `AssetModified` messages that aren't required to be ingested (such as metadata changes). In order to reduce churn, an attribute should be added to the request that indicates the asset will be ingested by engine
+  - This attribute should be called something like `EngineNotified`
+- asset requires the `DeliveryChannelPolicyId` to work out differences in policies (this should be there already)
+- asset requires `roles` as changes to roles can mean the `info.json` needs to be removed
+- API should not be responsible for deciding how cleanup is conducted
+
+## AWS
+
+As part of the changes, there will need to be some changes to the AWS estate to support the changes to asset modified.  Primarily, this will be updating the current SNS topic and adding an SQS listener to listen for asset modified notifications.  Additionally, the queue needs to be set to only listen for assets that have a `EngineNotified` attribute.
```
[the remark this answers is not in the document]
JackLewis-digirati: based on having to check before/after, this needs to be changed to listen to everything on `AssetModified` and let listener decide what to do

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,127 @@
...
+- `delay_seconds` will be used to delay messages being added to the queue, to avoid issues with multiple ingests
+- a dead letter queue should be added, as well as `visiblityTimeout` and `maxRetries`, this will help to avoid issues where assets are continually reprocessed
+
+## Handler
+
+After being added to the SQS queue, it needs to be handled by the cleanup handler being extended.  The reason to use a .Net application, is that there will be a lot of shared logic with Engine (for example, working out storage keys) and that access methods for the database and S3 have been written for the DLCS. Additionally, a new implementation of `ImessageHandler` will be created that's designed to handle `AssetModified`
+
+### Specifics
+
+- Assets should be checked for ingestion complete from the `Finished` property before being processed as well as the `Error` property
```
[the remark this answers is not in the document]
JackLewis-digirati: might be best with the `Before` value - mainly thinking this as it's more relevant to the system than a timestgamp on a message.

For `Ingesting` I don't think it would hurt - so will add in

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,127 @@
...
+
+- delivery channel changed
+- policy id changed
+- roles changed
+- policy data updated
+- origin changes
+
+#### Delivery channels changed
+
+This becomes an issue when a delivery channel is changed away from a specific policy, with the following implications:
```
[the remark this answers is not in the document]
JackLewis-digirati: will make this more clear, but I mean removing a delivery channel from an image

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,140 @@
...
+  - There's already logic to handle batch sending of notifications, just need to get list of messages to send.
+- It's possible to send `AssetModified` messages that aren't required to be ingested (such as metadata changes) and this is primarily controlled by `processAssetResult.RequiresEngineNotification`. In order to reduce churn, an attribute should be added to the request that indicates the asset will be ingested by engine
+  - This attribute should be called something like `EngineNotified`
+- Asset requires the `DeliveryChannelPolicyId` to work out differences in policies (this should be there already)
+- Asset requires `roles` as changes to roles can mean the `info.json` needs to be removed
+- API should not be responsible for deciding when cleanup is conducted
+
+## AWS
+
+As part of the changes, there will need to be some changes to the AWS estate to support the changes to asset modified.  Primarily, this will be updating the current SNS topic and adding an SQS listener to listen for asset modified notifications.  The queue should forward all requests to the listener as there's a possibility
```
donaldgray, line 26: Last sentance isn't finished `"The queue should forward all requests to the listener as there's a possibility..."`

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Logic
+
+#### Thumbs recalculation
+
+Within this, the most complex part of this is recalculating thumbnails.  The general approach to this will be to use the policy of the current asset, along with the thumbnails that currently exist within S3.  This will require a `ListBucket` operation from S3 which has a cost implication.  This could mean that checking 53 million images, would incur a cost of approximately $265
+
+Due to the cost, calls to `ListBucket` should be limited
+
+Thumbs should use the iiif-net library to check expected sizes of thumbnails for an image, and there needs to be some logic to not remove thumbs that are within 2-3 pixels of the expected to avoid off-by-one errors in thumbnail generation.  These off-by-one errors occur due to rounding errors in cantaloupe thumbs generation, more detgails can be found [here](https://github.com/dlcs/protagonist/pull/819)
```
donaldgray, line 56: Typo `"detgails"`

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Delivery channels changed
+
+This becomes an issue when a delivery channel is removed from an asset, with the following implications:
+
+- iiif-img removed
+  - Stored iiif-img derivative needs to be removed
+  - `info.json` removed
+  - Asset can exist at both filename and `/original` path - though the `original` needs to be left if the file channel exists on the asset
+  - if `thumbs` not there, then `s.json` and thumbnails need removed as well
```
donaldgray, line 84: and remove `AssetApplicationMetadata`. I think this is implied by removal of `s.json` but it's explicit elsewhere

docs/rfcs/017-asset-modified-cleanup.md
```diff
@@ -0,0 +1,140 @@
...
+
+This should only have an implication on the `info.json`, which would need to be removed
+
+#### Policy data updated
+
+The policy data being updated can be found from the date that the delivery channel policy was updated, after the `finished` date of the current asset itself. 
+
+- iiif-img changed
+  - `info.json` needs removed
+  - If it moves to a `use-original` policy, is there a need to remove the asset as well?
```
donaldgray, line 128: Yes - if `use-original` then the generated JP2 derivative won't be used. Similarly if it was `use-original` and went to `default` we can remove the `/original` item (if not used by "file" channel)

donaldgray (approved):

PR #833, opened 2024-04-29
Fix API returning 500 when an image is supplied with an unknown channel

### LINKED ISSUE #822: Invalid DeliveryChannel returns 500
If an invalid deliveryChannel value is specified when making API request it results in a 500, rather than a 400. This happens whether using the 'full' format or one of the reduced formats. 

E.g. the following would result in a 500.

```
"deliveryChannels": ["foo"]

"deliveryChannels": [{ "channel": "foo" } ]

"deliveryChannels": [{ "channel": "foo", "policy": "default" } ]
```

griffri:
Resolves #822 and WDC-85

This PR fixes a bug which caused the API to return 500 on attempting to PUT an image containing an unknown channel. It also adds a test confirming that 400 is sent in this scenario.

### DIFF
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -26,6 +26,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Test.Helpers.Data;
 using Test.Helpers.Integration;
 using Test.Helpers.Integration.Infrastructure;
 using AssetFamily = DLCS.Model.Assets.AssetFamily;
@@ -284,7 +285,7 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
     public async Task Put_NewImageAsset_BadRequest_WhenDeliveryChannelInvalid()
     {
         // arrange
-        var assetId = new AssetId(99, 1, nameof(Put_NewImageAsset_BadRequest_WhenDeliveryChannelInvalid));
+        var assetId = AssetIdGenerator.GetAssetId();
         var hydraImageBody = $@"{{
             ""@type"": ""Image"",
             ""origin"": ""https://example.org/my-image.png"",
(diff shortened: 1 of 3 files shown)


JackLewis-digirati (approved):

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -111,4 +111,16 @@ private void DeliveryChannelDependantValidation()
...
                 $"ImageOptimisationPolicy '{KnownImageOptimisationPolicy.UseOriginalId}' only valid for image delivery-channel");
     }
+
+    private bool DeliveryChannelIsValidForMediaType(string channel, string mediaType)
+    {
+        try
+        {
+            return AssetDeliveryChannels.IsChannelValidForMediaType(channel, mediaType);
+        }
+        catch(ArgumentOutOfRangeException)
```
donaldgray, line 121: Would it be cleaner to optionally throw an exception in `IsChannelValidForMediaType` if we know that it could be invalid in some instances?

```cs
static bool IsChannelValidForMediaType(string deliveryChannel, string mediaType);

static bool IsChannelValidForMediaType(string deliveryChannel, string mediaType, bool throwIfUnknownChannel = true);
```

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -279,7 +279,32 @@ public async Task Put_NewImageAsset_Creates_Asset_WhileIgnoringCustomDefaultDeli
                                                                 x.DeliveryChannelPolicyId == KnownDeliveryChannelPolicies.ImageDefault);
         asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == "thumbs");
     }
-    
+
+    [Fact]
+    public async Task Put_NewImageAsset_BadRequest_WhenDeliveryChannelInvalid()
+    {
+        // arrange
+        var assetId = new AssetId(99, 1, nameof(Put_NewImageAsset_BadRequest_WhenDeliveryChannelInvalid));
```
donaldgray, line 287: Use `AssetIdGenerator.GetAssetId()`

donaldgray (approved):

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
griffri: I think it would be. It saves on similar handlers needing to be written elsewhere if needed, too.

PR #835, opened 2024-04-29
Fix incorrect `WcDeliveryChannels` emulation

griffri:
This PR fixes an issue introduced in #776 where `wcDeliveryChannels` weren't being emulated properly.

### DIFF
--- a/src/protagonist/API/Features/DeliveryChannels/Converters/OldHydraDeliveryChannelsConverter.cs
+++ b/src/protagonist/API/Features/DeliveryChannels/Converters/OldHydraDeliveryChannelsConverter.cs
@@ -25,31 +25,42 @@ public class OldHydraDeliveryChannelsConverter
         
         foreach (var wcDeliveryChannel in hydraImage.WcDeliveryChannels)
         {
-            var matchedDeliveryChannel = wcDeliveryChannel switch
+            var matchedDeliveryChannels = wcDeliveryChannel switch
             {
-                AssetDeliveryChannels.Image => new DeliveryChannel()
-                {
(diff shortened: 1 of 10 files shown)


src/protagonist/API/Features/Image/ImagesController.cs
```diff
@@ -26,6 +26,7 @@ namespace API.Features.Image;
 [ApiController]
 public class ImagesController : HydraController
 {
+    private readonly ApiSettings apiSettings;
```
donaldgray, line 29: Unnecessary, exposed by `HydraController`

src/protagonist/API/Features/DeliveryChannels/Converters/OldHydraDeliveryChannelsConverter.cs
```diff
@@ -25,31 +25,42 @@ public class OldHydraDeliveryChannelsConverter
...
+                    new() 
+                    {
+                        Channel = AssetDeliveryChannels.Image,
+                        Policy = hydraImage.ImageOptimisationPolicy == ImageUseOriginalPolicy
+                            ? ImageUseOriginalPolicy
+                            : ImageDefaultPolicy 
+                    },
+                    new() 
+                    {
+                        Channel = AssetDeliveryChannels.Thumbnails
```
donaldgray, line 40: Should this have default policy, rather than just channel? 

Original ACs in #714 have `"thumbs" ImageDeliveryChannel. Always with default policy.` but I'm not sure if there's any practical difference?

donaldgray (approved):

PR #836, opened 2024-04-30
Allow escaped forward slashes to be used in named query requests

JackLewis-digirati:
This PR enables named queries to use `%2F` in the place of `/` to escape values used as parameters.

It works on zip, pdf and standard named queries


src/protagonist/Orchestrator.Tests/Integration/PdfTestsNoLocalstack.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Orchestrator.Infrastructure.NamedQueries.Persistence;
+using Orchestrator.Tests.Integration.Infrastructure;
+using Test.Helpers.Integration;
+
+namespace Orchestrator.Tests.Integration;
+
+/// <summary>
+/// Tests pdf requests that cannot be tested using localstack
```
donaldgray, line 21: Worth adding an explanation of why these can't use localstack + link to the issue? Hopefully once it's fixed we can roll these back into rest of tests.

donaldgray (approved):

PR #837, opened 2024-05-01
Remove image services from asset manifest if not available

### LINKED ISSUE #824: Manifests incorrectly showing image service
Image services are included in manifests when there is no image-service available. This should be fixed on both single-item manifests and named query projections.

Image service should only be included if the asset has "iiif-img" delivery channel (e.g. if combination of "thumbs" and/or "file", or "none") then it shouldn't be included.

* Update single-item manifest filtering. 404 for any images that don't have either "thumbs" or "iiif-img" channels. (currently it's filtering on AssetFamily but this is out of date.
* If no "thumbs" then "thumbnail" property won't be handled (already done)
* If no "iiif-img" property then painting annotation "body" will be empty

_This is a stop gap until we fully implement single-item manifests in #488_

griffri:
Resolves #824 and WDC-87

This PR removes image services from V2/V3 asset manifests for assets that do not have the `iiif-img` delivery channel. It also updates `GetManifestForAsset` to return 404 if the asset is not being served on `thumbs` or `iiif-img`.

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
+++ b/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
@@ -44,12 +44,21 @@ public static class AssetDeliveryChannels
         => asset.ImageDeliveryChannels != null &&
            asset.ImageDeliveryChannels.Count == 1 && 
            asset.HasDeliveryChannel(deliveryChannel);
-    
+
     /// <summary>
     /// Checks if an asset has any delivery channel specified in a list
     /// </summary>
     public static bool HasAnyDeliveryChannel(this Asset asset, params string[] deliveryChannels)
-        => deliveryChannels.Any(asset.HasDeliveryChannel);
+    {
+        if (asset.ImageDeliveryChannels.IsNullOrEmpty() || deliveryChannels.IsNullOrEmpty()) return false;
+        if (deliveryChannels.Any(dc => !All.Contains(dc)))
+        {
+            throw new ArgumentOutOfRangeException(nameof(deliveryChannels), deliveryChannels,
+                $"Acceptable delivery-channels are: {AllString}");
+        }
+        
+        return asset.ImageDeliveryChannels.Any(dc => deliveryChannels.Contains(dc.Channel));
+    }
 
     /// <summary>
     /// Checks if string is a valid delivery channel
--- a/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
+++ b/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
@@ -21,30 +21,7 @@ public static class AssetDeliveryChannels
     /// All possible delivery channels as a comma-delimited string
     /// </summary>
     public static readonly string AllString = string.Join(',', All);
-
-    /// <summary>
-    /// Check if asset has specified deliveryChannel
-    /// </summary>
-    public static bool HasDeliveryChannel(this Asset asset, string deliveryChannel)
-    {
-        if (asset.ImageDeliveryChannels.IsNullOrEmpty()) return false;
-        if (!All.Contains(deliveryChannel))
-        {
-            throw new ArgumentOutOfRangeException(nameof(deliveryChannel), deliveryChannel,
-                $"Acceptable delivery-channels are: {AllString}");
-        }
-
-        return asset.ImageDeliveryChannels.Any(i => i.Channel == deliveryChannel);
-    }
-
-    /// <summary>
-    /// Checks if asset has specified deliveryChannel only (e.g. 1 channel and it matches specified value
-    /// </summary>
-    public static bool HasSingleDeliveryChannel(this Asset asset, string deliveryChannel)
-        => asset.ImageDeliveryChannels != null &&
-           asset.ImageDeliveryChannels.Count == 1 && 
-           asset.HasDeliveryChannel(deliveryChannel);
-
+    
     /// <summary>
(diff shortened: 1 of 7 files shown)


src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
```diff
@@ -45,6 +45,12 @@ public static bool HasSingleDeliveryChannel(this Asset asset, string deliveryCha
            asset.ImageDeliveryChannels.Count == 1 && 
            asset.HasDeliveryChannel(deliveryChannel);
     
+    /// <summary>
+    /// Checks if an asset has any delivery channel specified in a list
+    /// </summary>
+    public static bool HasAnyDeliveryChannel(this Asset asset, params string[] deliveryChannels)
+        => deliveryChannels.Any(asset.HasDeliveryChannel);
```
donaldgray, line 52: This method impl feels like the most convenient way to do this, rather than the most efficient. It won't have a huge effect because `ImageDeliveryChannels` is fairly small but for every `deliveryChannels` passed:

* The `if (asset.ImageDeliveryChannels.IsNullOrEmpty()) return false;` check will happen
* `asset.ImageDeliveryChannels.Any` call will happen meaning multiple iterations.

src/protagonist/Test.Helpers/Integration/DatabaseTestDataPopulation.cs
```diff
@@ -207,4 +210,34 @@ public static class DatabaseTestDataPopulation
         asset.Result.Entity.WithTestThumbnailMetadata(metadataValue);
         return asset;
     }
+
+    public static ValueTask<EntityEntry<Asset>> WithTestDeliveryChannel(
+        this ValueTask<EntityEntry<Asset>> asset,
+        string deliveryChannel,
+        int? policyId = null)
+    {
+        if (!AssetDeliveryChannels.IsValidChannel(deliveryChannel))
```
donaldgray, line 219: Good call adding helper methods for test setups, can save a lot of boiler plate. One question on this check - is this required? As these methods are purely used to generate test data, could a valid test involve having invalid delivery-channels?

src/protagonist/Test.Helpers/Integration/DatabaseTestDataPopulation.cs
```diff
@@ -2,18 +2,21 @@
...
 using DLCS.Model.Assets.NamedQueries;
 using DLCS.Model.Customers;
 using DLCS.Model.DeliveryChannels;
+using DLCS.Model.Policies;
 using DLCS.Model.Spaces;
 using DLCS.Model.Storage;
 using DLCS.Repository.Auth;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
+using NuGet.Packaging;
```
donaldgray, line 19: ?

src/protagonist/Test.Helpers/Integration/DatabaseTestDataPopulation.cs
```diff
@@ -207,4 +210,34 @@ public static class DatabaseTestDataPopulation
...
+            Channel = deliveryChannel,
+            DeliveryChannelPolicyId = policyId ?? deliveryChannel switch
+            {
+                AssetDeliveryChannels.Image => KnownDeliveryChannelPolicies.ImageDefault,
+                AssetDeliveryChannels.Thumbnails => KnownDeliveryChannelPolicies.ThumbsDefault,
+                AssetDeliveryChannels.Timebased => MIMEHelper.IsVideo(asset.Result.Entity.Origin)
+                    ? KnownDeliveryChannelPolicies.AvDefaultVideo
+                    : KnownDeliveryChannelPolicies.AvDefaultAudio,
+                AssetDeliveryChannels.File => KnownDeliveryChannelPolicies.FileNone,
+                _ => throw new ArgumentOutOfRangeException(nameof(deliveryChannel), deliveryChannel,
```
donaldgray, line 236: As above - is this required? Or should it be controllable via a param to this method?

src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
```diff
@@ -404,6 +459,51 @@ public async Task Get_V3ManifestForImage_ReturnsManifestNoThumbnails_WhenNoThumb
...
+            
+        var path = $"iiif-manifest/{id}";
+
+        // Act
+        var response = await httpClient.GetAsync(path);
+            
+        // Assert
+        var jsonResponse = JObject.Parse(await response.Content.ReadAsStringAsync());
+        jsonResponse["id"].ToString().Should().Be($"http://localhost/iiif-manifest/{id}");
+        jsonResponse.SelectToken("items[0].items[0].items[0].body").Should().NotBeNull();
```
donaldgray, line 479: FYI (not worth changing but just to be aware) some of these test pre-date the iiif-net nuget package handling deserialisation correctly which is why `SelectToken` was used (it can still be the best choice sometimes though!). Now that it is supported you can deserialize to a strongly typed model if required.

```cs
// this
var jsonResponse = JObject.Parse(await response.Content.ReadAsStringAsync());
jsonResponse["id"].ToString().Should().Be($"http://localhost/iiif-manifest/{id}");

// can be
var manifest = (await response.Content.ReadAsStreamAsync()).FromJsonStream<IIIF3.Manifest>();
manifest.Id.Should().Be($"http://localhost/iiif-manifest/{id}");

donaldgray (approved):

Issue #430, opened 2022-12-01
Delete no-longer required delivery artifacts


donaldgray:
Ticket #393 introduced the concept of delivery-channels. 

Different combinations of deliveryChannel and mediaType can result in different resources being created and uploaded to S3. We need to catch if a change in deliveryChannel means that some of these resources are now no-longer required and delete them.

This would be difficult to do in the Engine as it gets posted a payload containing the current state of the Asset, at which time the database record has been written to show it is being processed.

A notification raised by `IAssetNotificationSender.SendAssetModifiedNotification(ChangeType changeType, Asset? before, Asset? after)` would be a sensible place to implement this logic. It's already hooked up for single PUT operations but bulk operations would need extended as they only send details of the AssetId, rather than fully hydrated Asset. This was done to avoid additional queries to database (e.g. 100 updates would require 100 db calls) but it sounds like this is the best place to handle this requirement.

donaldgray:
This should also account for items that change Thumbnail policy or auth service.

donaldgray:
Note that this will handle changes to current assets, would require a one off bulk cleanup to remove legacy 'old' derivatives.

JackLewis-digirati:
This ticket now covers cleaning up thumbnail data from the `AssetApplicationMetadata` table

donaldgray:
Note: this requires the addition of a new SQS queue subscribed to existing SNS queue.

PR #841, opened 2024-05-02
Api changes to support asset modified cleanup

JackLewis-digirati:
This PR is supporting #430, but does not complete the ticket by itself.

It works by updating batch operations to send notifications to the `AssetModified` queue, and adds a message attribute to `Update` requests when the engine is notified

### DIFF
--- a/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
+++ b/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
@@ -88,8 +88,8 @@ public class CreateBatchOfImagesHandler : IRequestHandler<CreateBatchOfImages, M
         
         var batch = await batchRepository.CreateBatch(request.CustomerId, request.AssetsBeforeProcessing.Select(a => a.Asset).ToList(), cancellationToken);
 
-        var assetNotificationList = new List<Asset>(request.AssetsBeforeProcessing.Count);
-        var snsNotificationList = new List<AssetModificationRecord>();
+        var engineNotificationList = new List<Asset>(request.AssetsBeforeProcessing.Count);
+        var assetModifiedNotificationList = new List<AssetModificationRecord>();
 
         try
         {
@@ -116,11 +116,11 @@ public class CreateBatchOfImagesHandler : IRequestHandler<CreateBatchOfImages, M
                 var assetModificationRecord = existingAsset == null
                     ? AssetModificationRecord.Create(savedAsset)
                     : AssetModificationRecord.Update(existingAsset, savedAsset, processAssetResult.RequiresEngineNotification);
-                snsNotificationList.Add(assetModificationRecord);
+                assetModifiedNotificationList.Add(assetModificationRecord);
                 
                 if (processAssetResult.RequiresEngineNotification)
                 {
-                    assetNotificationList.Add(savedAsset);
+                    engineNotificationList.Add(savedAsset);
                 }
                 else
                 {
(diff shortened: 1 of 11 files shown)


src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
```diff
@@ -85,6 +89,8 @@ public class CreateBatchOfImagesHandler : IRequestHandler<CreateBatchOfImages, M
         var batch = await batchRepository.CreateBatch(request.CustomerId, request.AssetsBeforeProcessing.Select(a => a.Asset).ToList(), cancellationToken);
 
         var assetNotificationList = new List<Asset>(request.AssetsBeforeProcessing.Count);
+        var snsNotificationList = new List<AssetModificationRecord>();
```
donaldgray, line 92: is prefixing this with `sns` leaking the underlying implmementation? It'll currently be published to an sns topic but that could change without calling code needing to know.

src/protagonist/API/Infrastructure/Messaging/AssetModificationRecord.cs
```diff
@@ -13,22 +13,25 @@ public class AssetModificationRecord
     public Asset? Before { get; }
     public Asset? After { get; }
     
+    public bool AssetModifiedEngineNotified { get; }
```
donaldgray, line 16: As this is a property of `AssetModificationRecord` is the `AssetModified` prefix redundant? `AssetModificationRecord.AssetModifiedEngineNotified` vs `AssetModificationRecord.EngineNotified`.

This new property is specific to `ChangeType.Update` notifications. `DeleteFrom` is specific to `ChangeType.Delete` notifications - should be be cleverer about how `ChangeType`-specific information is included, rather than adding a new property every time?

src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
```diff
@@ -10,11 +10,11 @@ public interface ITopicPublisher
...
+    public Task<bool> PublishToAssetModifiedTopic(IReadOnlyList<AssetModifiedNotification> messages, 
         CancellationToken cancellationToken);
 }
 
 /// <summary>
 /// Represents the contents + type of change for Asset modified notification
 /// </summary>
-public record AssetModifiedNotification(string MessageContents, ChangeType ChangeType);
\ No newline at end of file
+public record AssetModifiedNotification(string MessageContents, ChangeType ChangeType, bool EngineNotified);
```
donaldgray, line 20: Is it wrong to have an "Update"-specific property for what should be a generic representation of an asset modified notification? If we needed to add other attributes would these be chained here? (e.g. if we wanted to have the `DeleteFrom` as an attribute, would that be added here despite only being relevant for deletes?).

Is there a way to keep `AssetModifiedNotification` as a generic carrier of notification data without having ChangeType specific props? (similar comment to one for `AssetModificationRecord`)

src/protagonist/DLCS.AWS/SNS/TopicPublisher.cs
```diff
@@ -98,17 +99,29 @@ public class TopicPublisher : ITopicPublisher
...
         };
-        return new Dictionary<string, MessageAttributeValue>
+        var messageAttributes = new Dictionary<string, MessageAttributeValue>
         {
             { "messageType", attributeValue }
         };
+        
+        if (engineNotified)
+        {
+            messageAttributes.Add(new KeyValuePair<string, MessageAttributeValue>("EngineNotified",
```
donaldgray, line 117: Should this be `"engineNotified"`? Other used attribute is `"messageType"` so having camelCase for some and PascalCase for others could lead to possible issues reading them as they're case sensitive.

src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
JackLewis-digirati: I wanted something that would make it obvious it was doing something different from engine notification - best way to do this might be to change `assetNotificationList` to `engineNotificationList`, then changing this value to `assetModifiedNotificationList`

donaldgray (approved):

PR #856, opened 2024-05-23
reate asset modified cleanup handler

JackLewis-digirati:


src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,357 @@
...
+    private async Task CleanupChangedTimebasedDeliveryChannel(ImageDeliveryChannel imageDeliveryChannel, AssetId assetId)
+    {
+        var presetList = JsonSerializer.Deserialize<List<string>>(imageDeliveryChannel.DeliveryChannelPolicy.PolicyData);
+        List<ObjectInBucket> assetsToDelete;
+        var keys = new List<string>();
+        var containers = new List<string>();
+
+        foreach (var presetIdentifier in presetList)
+        {
+            var presetDictionary = await engineClient.GetAvPresets();
```
donaldgray, line 194: This can be called once, don't need to call per preset. 

The results of this call should also be cached to avoid repeated HTTP requests to Engine for data that's unlikely to change. What happens if we can't fetch presets from Engine? The changes are it'll be due to a transient failure so maybe worth failing the message processing and picking up later?

src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,357 @@
...
+                    keys.AddRange(keysFromAws);
+                    containers.Add(presetDetails.Container);
+                }
+            }
+        }
+
+        assetsToDelete = keys.Where(k =>
+                !containers.Contains(k.Split('.').Last()) &&
+                k.Contains(handlerSettings.AssetModifiedSettings.TimebasedKeyIndicator))
+            .Select(k => new ObjectInBucket(handlerSettings.AWS.S3.StorageBucket, k)).ToList();
```
donaldgray, line 215: I think this method should make use of `TranscoderTemplates`. `.mp4` is a valid extension for both audio and video. This looks like it could lead to accidental deletion of the wrong one.

src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,357 @@
...
+using DLCS.Model.Messaging;
+using DLCS.Model.Policies;
+using DLCS.Repository.Messaging;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Version = IIIF.ImageApi.Version;
+
+namespace CleanupHandler;
+
+public class AssetUpdatedHandler  : IMessageHandler
```
donaldgray, line 22: General observation on this class - is it worth having a 'transaction'. That could be done with any DB updates (ie only call `.SaveChanges()` once everything has been done). Similar approach to deleting keys/folders from S3 would be to run through and build up a list of all the operations to carry out and then actually do them at the end.

```cs
public async Task<bool> HandleMessage(QueueMessage message, CancellationToken cancellationToken = default)
{
  try
  {
    // as-is but don't actually call S3, just build up lists of keys to delete
    // means that stuff isn't left in a half removed state
    await CleanupRemoved();
    await CleanupChanges();
    
    // now do the actual removal / commit of transactino
    await dbContext.SaveChanges();
    await DeleteFromS3(keysToDelete);
  }
  catch (Exception) { return false ; }
}

File: src/protagonist/CleanupHandler/AssetUpdatedHandler.cs [hunk 4]
```diff
@@ -0,0 +1,357 @@
...
+        
+        var thumbsToDelete = await ThumbsToBeDeleted(assetAfter);
+        
+        bucketObjectsTobeRemoved.AddRange(thumbsToDelete);
+        RemoveObjectsFromBucket(bucketObjectsTobeRemoved);
+    }
+
+    private void CleanupChangedImageDeliveryChannel( ImageDeliveryChannel modifiedDeliveryChannel, AssetId assetId)
+    {
+        List<ObjectInBucket> bucketObjectsTobeRemoved = new()
```
griffri, line 232: Casing typo - should be `bucketObjectsToBeRemoved`

src/protagonist/CleanupHandler/Repository/CleanupHandlerAssetRepository.cs
```diff
@@ -0,0 +1,21 @@
+﻿using DLCS.Core.Types;
+using DLCS.Model.Assets;
+using DLCS.Repository;
+using DLCS.Repository.Assets;
+
+namespace CleanupHandler.Repository;
+
+public class CleanupHandlerAssetRepository : ICleanupHandlerAssetRepository
```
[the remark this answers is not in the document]
JackLewis-digirati: I used this as I'm using other `repo` classes in the `AssetModifiedHandler`, so I didn't want to mix `Repo` classes and the `DlcsContext`

src/protagonist/DLCS.AWS/ElasticTranscoder/ElasticTranscoderWrapper.cs
```diff
@@ -51,31 +51,52 @@ public class ElasticTranscoderWrapper : IElasticTranscoderWrapper
...
+            return presetsDictionary;
+        }, cacheSettings.GetMemoryCacheOptions(CacheDuration.Long, priority: CacheItemPriority.Low));
+    }
 
-            if (presets.Count == 0)
-            {
-                logger.LogWarning("No ElasticTranscoder presets found");
-                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromSeconds(cacheSettings.GetTtl(CacheDuration.Short));
-            }
+    public Task<Preset?> GetPresetDetails(string name, CancellationToken token)
```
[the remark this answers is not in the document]
JackLewis-digirati: think the cache should probably return a list, as that's what we're getting from AWS, and the key is technially a redundant value - think `GetPresetIdLookup` should return a dictionary though, to avoid potential (but highly unlikely) issues of multiple presets with the same name

src/protagonist/DLCS.Model/Assets/Asset.cs
```diff
@@ -116,4 +116,17 @@ public Asset(AssetId assetId)
...
         Space = assetId.Space;
     }
+
+    public Asset Clone()
+    {
+        var asset = (Asset)MemberwiseClone();
+
+        var deliveryChannels = new ImageDeliveryChannel[asset.ImageDeliveryChannels.Count];
+
+        asset.ImageDeliveryChannels.CopyTo(deliveryChannels, 0);
```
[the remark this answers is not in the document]
JackLewis-digirati: this did work when items were added/removed, but wouldn't work  with modified - I've moved this to use a deep copy

src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,357 @@
...
+    {
+        var filteredFilenames = thumbsBucketKeys.Where(t => FileSystemName.MatchesSimpleExpression("*.jpg", t));
+
+        var thumbBucketSizes = filteredFilenames
+            .Select(f => (f.Split("/").Last().Split('.').First(), f)).ToList();
+
+        return thumbBucketSizes;
+    }
+
+    private void CleanupRemovedImageDeliveryChannel(Asset assetAfter)
```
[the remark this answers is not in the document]
JackLewis-digirati: I thought that was `storageKeyGenerator.GetStorageLocation` - is there another one I've missed?

src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,388 @@
...
+            default:
+                logger.LogDebug("policy {PolicyName} does not require any changes for asset {AssetId}",
+                    deliveryChannelModified.DeliveryChannelPolicy.Name, assetAfter.Id);
+                break;
+        }
+    }
+
+    private async Task CleanupChangedTimebasedDeliveryChannel(ImageDeliveryChannel imageDeliveryChannel, Asset assetAfter, HashSet<ObjectInBucket> objectsToRemove)
+    {
+        var presetList = JsonSerializer.Deserialize<List<string>>(imageDeliveryChannel.DeliveryChannelPolicy.PolicyData);
```
donaldgray, line 214: is it worth using overload from `DeliveryChannelPolicyX`? Either directly or introducing a new one?

```cs
var presetList = imageDeliveryChannel.DeliveryChannelPolicy.PolicyDataAs<List<string>>();

// or with new method, like .ThumbsDataAsSizeParameters()
var presetList = imageDeliveryChannel.DeliveryChannelPolicy.FooData();
```

benefit of latter, specific method is it incorporates safety check DCP is of correct type and avoids other consumers needing to explicitly know and specify the type each time.

src/protagonist/CleanupHandler/AssetUpdatedHandler.cs
```diff
@@ -0,0 +1,388 @@
...
+        var keys = new List<string>();
+        var extensions = new List<string>();
+        var mediaPath = RetrieveMediaPath(assetAfter);
+        
+        var presetDictionary = await engineClient.GetAvPresets();
```
donaldgray, line 219: Should this fail if call to Engine failed?

donaldgray (approved):

Issue #784, opened 2024-04-04
Update LegacyModeConverter to handle new DeliveryChannels


donaldgray:
`LegacyMode` introduced in #463 defaults various settings on the way 'in' to allow for backwards compatibility with payloads from before DeliveryChannels were introduced.

This will need extended to handle `DeliveryChannels` being empty. An empty `DeliveryChannel` element on create will use defaults but on update will result in a 400, however for legacy customers this is still a valid payload.

Also, does this #714 have any bearing on this? That change will return 400 if `ImageOptimisationPolicy` or `ThumbnailPolicy` are specified - this will maybe need resolved too. How do we do this without having an overly complicated set of flags that need to be turned on/off. #714 added a shortlived emulation layer to allow deployment to Wellcome environment only - it might be easier to do this task after that layer has been removed.

## 

For customers who have LegacyMode enabled:

For "Image"
  * If `"ImageOptimisationPolicy"` specified - only accept `"fast-higher"`, which we map to image channel `"default"`
  * If `"ThumbnailPolicy"` specified - only accept `"default"`
  * If `"DeliveryChannel"` empty, set to `["iiif-img", "thumbs"]`

For "Audio"
  * If `"ImageOptimisationPolicy"` specified - only accept `"audio-max"`, which we map to channel `"default-audio"`
  * If `"DeliveryChannel"` empty, set to `["iiif-av"]`

For "Video"
  * If `"ImageOptimisationPolicy"` specified - only accept `"video-max"`, which we map to channel `"default-video"`
  * If `"DeliveryChannel"` empty, set to `["iiif-av"]`

For "File"
  * If `"DeliveryChannel"` empty, set to `["file"]`

PR #858, opened 2024-05-28
Remove `wcDeliveryChannels` + emulation layer

### LINKED ISSUE #855: Remove `wcDeliveryChannels` and emulation layer
`wcDeliveryChannels` was a temporary setting that was only required for temporary rollout. This can now be removed, along with the emulation layer introduced in #714

griffri:
Resolves #855 and WDC-90

This PR removes support for `wcDeliveryChannels` and the emulation layer added for #714. It also updates how `ImageOptimisationPolicy` and `ThumbnailPolicy` are handled in preparation for work on on #784


src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -28,29 +28,26 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
...
+
+        RuleFor(a => a.ThumbnailPolicy)
+            .Null()
+            .When(_ => !apiSettings.Value.EmulateOldDeliveryChannelProperties)
+            .WithMessage("'ThumbnailPolicy' is disabled");
+        
         // System edited fields
+        RuleFor(a => a.Width).Empty().WithMessage("Should not include width");
+        RuleFor(a => a.Height).Empty().WithMessage("Should not include height");
+        RuleFor(a => a.Duration).Empty().WithMessage("Should not include duration");
```
donaldgray, line 47: are these oversimplified compared to the rules they are replacing? Taking Duration as an example the deleted rule was

```cs
RuleFor(a => a.Duration)
    .Empty()
    .WithMessage("Should not include duration")
    .Unless(a =>
        a.WcDeliveryChannels.ContainsOnly(AssetDeliveryChannels.File) && !MIMEHelper.IsImage(a.MediaType));
```

(ie you can set Duration if it only has "File" channel and it's not an image) but it's been replaced with 
```cs
RuleFor(a => a.Duration).Empty().WithMessage("Should not include duration");
```

Width, Height and Duration are set by system when ingesting is completed but if I want to serve an Audio file via the "file" channel, no transcoding has taken place so we don't know the Duration; in that case I might want to set it directly via the API.

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -28,29 +28,26 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
...
-                RuleFor(a => a.Height).Empty().WithMessage("Should not include height");
-                RuleFor(a => a.Duration).Empty().WithMessage("Should not include duration");
-            });
-        
         When(a => !a.DeliveryChannels.IsNullOrEmpty(), ImageDeliveryChannelDependantValidation);
         
+        // Legacy policy fields
+        RuleFor(a => a.ImageOptimisationPolicy)
+            .Null()
+            .When(_ => !apiSettings.Value.EmulateOldDeliveryChannelProperties)
```
donaldgray, line 36: Will these be expanded in next ticket? (to capture that there are only some policies that we'll accept)

donaldgray (approved):

PR #866, opened 2024-06-18
Adding abilty to use a stickiness cookie when generating thumbs through cantaloupe

JackLewis-digirati:
Links DLCS-846

This PR implements handling a stickiness cookie coming from AWS which allows all thumbs requests to go through the same cantaloupe when using a load balancer with multiple cantaloupe instances.

It does this by attaching a cookie to requests made to cantaloupe for an instance of the `HttpClient`, after the first thumbs response.  This cookie comes from the AWS "stickiness" attribute and is entirely managed by AWS

### DIFF
--- a/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
+++ b/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
@@ -311,9 +311,9 @@ public class CantaloupeThumbsClientTests
         var response = new HttpResponseMessage(HttpStatusCode.OK);
         response.Headers.Add("Set-Cookie", new List<string?>()
         {
-            "AWSALBAPP-0=_remove_; Expires=Tue, 25 Jun 2024 10:56:45 GMT; Path=/",
-            "AWSALBAPP-1=_remove_; Expires=Tue, 25 Jun 2024 10:56:45 GMT; Path=/",
-            "AWSALBAPP-2=_remove_; Expires=Tue, 25 Jun 2024 10:56:45 GMT; Path=/"
+            "AWSALBAPP-0=_remove_; Path=/",
+            "AWSALBAPP-1=_remove_; Path=/",
+            "AWSALBAPP-2=_remove_; Path=/"
         });
         httpHandler.SetResponse(response);
         context.Asset.Width = 2000;
--- a/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
+++ b/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
@@ -10,6 +10,7 @@ using Engine.Ingest.Image.ImageServer.Measuring;
 using Engine.Settings;
 using FakeItEasy;
 using Microsoft.Extensions.Logging.Abstractions;
+using Test.Helpers.Data;
 using Test.Helpers.Http;
 using Test.Helpers.Settings;
 
@@ -316,7 +317,7 @@ public class CantaloupeThumbsClientTests
     public async Task GenerateThumbnails_UpdatesHandlerWithCookies()
     {
         // Arrange
-        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsThumbForSuccessfulResponse));
+        var assetId = AssetIdGenerator.GetAssetId();
         var context = IngestionContextFactory.GetIngestionContext(assetId: assetId.ToString());
 
         var response = new HttpResponseMessage(HttpStatusCode.OK);
@@ -355,7 +356,7 @@ public class CantaloupeThumbsClientTests
     public async Task GenerateThumbnails_DoesNotUpdateHandlerWithCookiesWhenUnrecognised()
     {
         // Arrange
-        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsThumbForSuccessfulResponse));
+        var assetId = AssetIdGenerator.GetAssetId();
         var context = IngestionContextFactory.GetIngestionContext(assetId: assetId.ToString());
 
         var response = new HttpResponseMessage(HttpStatusCode.OK);
@@ -403,13 +404,17 @@ public class CantaloupeThumbsClientTests
             S3 = "//some/location/with/s3"
         });
         
+        await sut.GenerateThumbnails(context, defaultThumbs, ThumbsRoot);
+        
+        List<CookieHeaderValue> cookieHeaders = new();
+        httpHandler.RegisterCallback(message => cookieHeaders = message.Headers.GetCookies().ToList());
+        httpHandler.GetResponseMessage("{ \"engine\": \"hello\" }", HttpStatusCode.OK);
+    
         // Act
         await sut.GenerateThumbnails(context, defaultThumbs, ThumbsRoot);
-
-        var cookies = httpClient.DefaultRequestHeaders.GetCookies();
-
+        
         // Assert
-        cookies.Count.Should().Be(0);
+        cookieHeaders.Count.Should().Be(0);
     }
 
     public class ImageOnDiskResults
--- a/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
+++ b/src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
@@ -1,5 +1,4 @@
 ﻿using System.Net;
-using System.Net.Http.Headers;
 using DLCS.Core.Exceptions;
 using DLCS.Core.FileSystem;
 using DLCS.Core.Types;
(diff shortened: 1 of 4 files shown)


griffri (approved):

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
```diff
@@ -84,6 +86,18 @@ public class CantaloupeThumbsClient : IThumbsClient
...
     }
 
+    private void AttemptToAddStickinessCookie(HttpResponseMessage response)
+    {
+        if (!cantaloupeClient.DefaultRequestHeaders.Contains("Cookie"))
+        {
+            var hasCookie = response.Headers.TryGetValues("Set-Cookie", out var cookies);
+            if (hasCookie)
+            {
+                cantaloupeClient.DefaultRequestHeaders.Add("Cookie", cookies!);
```
donaldgray, line 96: Is there any danger of this being reused if it's on the `DefaultRequestHeaders`? IIRC some element of the HttpClient can be pooled and reused when using `HttpClientFactory` does `DefaultRequestHeaders` fit the bill? 

Would it to be better to set it per request, would mean we could handle the following from docs:
> If a target fails or becomes unhealthy, the load balancer stops routing requests to that target, and chooses a new healthy target based on the chosen load balancing algorithm. The load balancer treats the session as now being "stuck" to the new healthy target, and continues routing requests to the new healthy target even if the failed target comes back.

It's unlikely that could happen but if we send cookies returned from request A with request B, then cookies from B -> C etc (rather than using cookies returned from request A with B, C, D....K) then we'd always be using the latest.

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
```diff
@@ -84,6 +86,18 @@ public class CantaloupeThumbsClient : IThumbsClient
         throw new HttpException(response.StatusCode, "failed to retrieve data from the thumbs processor");
     }
 
+    private void AttemptToAddStickinessCookie(HttpResponseMessage response)
+    {
+        if (!cantaloupeClient.DefaultRequestHeaders.Contains("Cookie"))
+        {
+            var hasCookie = response.Headers.TryGetValues("Set-Cookie", out var cookies);
```
donaldgray, line 93: Do we want to grab _all_ cookies returned or would it be better to have a whitelist?

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
JackLewis-digirati: spent some time taking a look at this, and like you said unlikely but possible - think I went this way as it's using `GetAsync` - will change it to use `SendAsync`

src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
```diff
@@ -299,6 +311,106 @@ public async Task GenerateThumbnails_ReturnsThumbForSuccessfulResponse_AfterFirs
             "Landscape images - invalid sizes altered",
         },
     };
+    
+    [Fact]
+    public async Task GenerateThumbnails_UpdatesHandlerWithCookies()
+    {
+        // Arrange
+        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsThumbForSuccessfulResponse));
```
donaldgray, line 319: can be `var id = AssetIdGenerator.GetAssetId();` or `var id = AssetIdGenerator.GetAssetId(2, 1);` if cust / space are important.

src/protagonist/Engine.Tests/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClientTests.cs
```diff
@@ -299,6 +311,106 @@ public async Task GenerateThumbnails_ReturnsThumbForSuccessfulResponse_AfterFirs
...
+        cookieHeaders[0].Cookies[0].Name.Should().Be("AWSALB");
+        cookieHeaders[0].Cookies[0].Value.Should().Be("_remove_");
+        cookieHeaders[1].Cookies[0].Name.Should().Be("AWSALBCORS");
+    }
+    
+    [Fact]
+    public async Task GenerateThumbnails_DoesNotUpdateHandlerWithCookiesWhenUnrecognised()
+    {
+        // Arrange
+        var assetId = new AssetId(2, 1, nameof(GenerateThumbnails_ReturnsThumbForSuccessfulResponse));
```
donaldgray, line 358: prefer `AssetIdGenerator`, this is using `nameof()` for an alternative method, which `AssetIdGenerator` will help avoid (same with other methods)

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
```diff
@@ -84,6 +93,37 @@ public class CantaloupeThumbsClient : IThumbsClient
...
+        {
+            request.Headers.Add("Cookie", loadBalancerCookies);
+        }
+
+        return request;
+    }
+
+    private void AttemptToAddStickinessCookie(HttpResponseMessage response)
+    {
+        if (!cantaloupeClient.DefaultRequestHeaders.Contains("Cookie"))
```
donaldgray, line 110: What is this check for? Is it needed?

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
```diff
@@ -84,6 +93,37 @@ public class CantaloupeThumbsClient : IThumbsClient
...
+        {
+            var hasCookie = response.Headers.TryGetValues("Set-Cookie", out var cookies);
+            if (hasCookie)
+            {
+                foreach (var cookie in cookies!)
+                {
+                    if (CookieHeaderValue.TryParse(cookie, out var parsedCookie) && parsedCookie.Cookies.Any(x =>
+                            engineSettings.ImageIngest!.LoadBalancerStickinessCookieNames.Contains(x.Name)))
+                    {
+                        loadBalancerCookies.Add(cookie);
```
donaldgray, line 120: Does this need to be a dict? Can cookies with same name but differing values or expires be returned? How are they handled if sent to ALB?

src/protagonist/Engine/Settings/EngineSettings.cs
```diff
@@ -114,6 +114,15 @@ public class ImageIngestSettings
     /// </summary>
     public List<string> DefaultThumbs { get; set; } = new();
 
+    /// <summary>
+    /// A set of cookie names used by the load balancer to indicate stickiness
+    /// </summary>
+    public List<string> LoadBalancerStickinessCookieNames { get; set; } = new()
```
donaldgray, line 120: Are these prefixes rather than set cookie names? To handle ALB cookies can split like `AWSALBAPP-*`

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
JackLewis-digirati: this check should be using the `loadBalancerCookies` list insdtead I think, and it's supposed to short circuit adding cookies to a request if they've already been added

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/CantaloupeThumbsClient.cs
JackLewis-digirati: the `cookie` being added is actually a string containing all the values of the cookie, so it would have the correct expiry and values for the load balancer - thought this was ok, as this is only set for the lifecycle of a request, and every new request would get these values added - should note the `Name` value taken from the parsed cookie is the key of a kvp that doesn't change (though I left it as configurable via app settings)

src/protagonist/Engine/Settings/EngineSettings.cs
JackLewis-digirati: these are the `key` value of a kvp, and are what I could see for an `lb_cookie` type of stickiness - the `AWSALPAPP-*` are   specific to `app_cookie` stickiness, and the reason they work like this, is (I believe) because the standard size of a cookie is 4k, and the load balancer `app_cookie` requires up to 16k, so it can split the cookie into a maximum of 4 shards.  This doesn't seem to be required by `lb_cookie` and the values below this line are what I could see being returned by the load balancer

JackLewis-digirati: for context, this is what the cookie looks like coming back from the load balancer - `AWSALB=IOCtmCsAoVQyTEz3uphH7JfMhKI5/378EAs5ZyUSKDY1Uzpxt9kAfmR7yi3m5C17EKkSt0XVh08Zich3qUk6p7+eBxBR4ET/Xvzf61AbEPxruWKqg/zflioL8fQJ; Expires=Tue, 09 Jul 2024 15:45:02 GMT; Path=/`

donaldgray (approved):

PR #871, opened 2024-06-26
Allow batch with empty 'members' in legacy mode

donaldgray:
If customer has LegacyMode enabled and we receive empty members then we will create an empty batch. To simplify processing a new handler was created solely for creating the batch, saves any complication of `CreateBatchOfImages` and makes removing simpler when LegacyMode is disabled.

Note that `null` members will still result in a 400 exception.

### DIFF
--- a/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
+++ b/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
@@ -559,6 +559,30 @@ public class CustomerQueueTests : IClassFixture<ProtagonistAppFactory<Startup>>
     }
     
     [Fact]
+    public async Task Post_CreateBatch_201_IfLegacyModeEnabled_MembersEmpty()
+    {
+        // Arrange
+        var hydraImageBody = @"{
+    ""@context"": ""http://www.w3.org/ns/hydra/context.jsonld"",
+    ""@type"": ""Collection"",
+    ""member"": []
+}";
+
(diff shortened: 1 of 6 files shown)


src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
```diff
@@ -558,6 +558,30 @@ public async Task Post_CreateBatch_201_IfLegacyModeEnabled()
         response.StatusCode.Should().Be(HttpStatusCode.Created);
     }
     
+    [Fact]
+    public async Task Post_CreateBatch_201_IfLegacyModeEnabled_MembersEmpty()
```
griffri, line 562: Would `Post_CreateBatch_201_IfLegacyModeEnabled_AndMembersEmpty()` be slightly clearer as a test name?

src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
```diff
@@ -698,6 +722,26 @@ public async Task Post_CreateBatch_400_WithInvalidIdsSet()
         response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
     }
 
+    [Fact]
+    public async Task Post_CreateBatch_400_MembersEmpty()
```
griffri, line 726: Similar to previous comment - `Post_CreateBatch_400_IfMembersEmpty()`?

griffri (approved):

PR #872, opened 2024-06-26
Adding support for optional named query parameters

### LINKED ISSUE #870: NamedQuery parameters should be optional
There has been a change in behaviour between how Deliverator and Protagonist handles NamedQuery parameters.

Deliverator allows you to specify multiple parameters in an NQ definition and only supply some of them, the others would effectively be ignored. However, this was tightened up in Protagonist and parameters became required - e.g. if an NQ is configured with 3 `p*` values but was only given 1 it will return a 400 exception, logging an error of `Not enough query arguments to satisfy template element parameter`.

Protagonist should allow missing parameters, a follow up ticket with introduce a new syntax to specify required parameters.

JackLewis-digirati:
Resolves #870 
Links DLCS-853

This PR adds support for optional query parameters in named queries.  This behavior mimics deliverator functionality, with an added guardrail to prevent no parameters being sent

### DIFF
--- a/src/protagonist/DLCS.Repository/NamedQueries/Parsing/BaseNamedQueryParser.cs
+++ b/src/protagonist/DLCS.Repository/NamedQueries/Parsing/BaseNamedQueryParser.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using Amazon.CloudFront.Model;
 using DLCS.Core.Collections;
 using DLCS.Core.Guard;
 using DLCS.Model.Assets.NamedQueries;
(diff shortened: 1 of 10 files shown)


src/protagonist/DLCS.Repository/NamedQueries/Parsing/BaseNamedQueryParser.cs
```diff
@@ -123,15 +124,15 @@ private T GenerateParsedNamedQuery(int customerId, string[] templatePairing, Lis
                         break;
                     case Number1:
                         assetQuery.Number1 =
-                            long.Parse(GetQueryArgumentFromTemplateElement(queryArgs, elements[1]));
+                            ConvertIntegerQueryArg(GetQueryArgumentFromTemplateElement(queryArgs, elements[1]));
```
donaldgray, line 127: Behaviour change - was `long.Parse` but it's now `int.Parse`

donaldgray (approved):

src/protagonist/DLCS.Repository/NamedQueries/Parsing/BaseNamedQueryParser.cs
JackLewis-digirati: could set the return from `ConvertIntegerQueryArg` to `long`?

PR #875, opened 2024-06-28
Allow named queries to be managed in Portal

### LINKED ISSUE #869: Reinstate Named Query page in portal
The interim portal does not have a page for viewing named queries, editing them, and creating new ones.

We need a simple named query page that uses the API to do these CRUD operations:
https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/named-queries

On create or update, this page will not validate the named query syntax.

Only `name` and `template` are editable.

NB there is no individual page for each NQ. The page in the portal lists the NQ as a table, you can edit a row, and create a new one at the foot of the table.

griffri:
Resolves #869 and DLCS-854

### DIFF
--- a/src/protagonist/API.Client/DlcsClient.cs
+++ b/src/protagonist/API.Client/DlcsClient.cs
@@ -317,7 +317,15 @@ public class DlcsClient : IDlcsClient
         var queue = await response.ReadAsHydraResponseAsync<CustomerQueue>(jsonSerializerSettings);
         return queue;  
     }
-
+    
+    public async Task<HydraCollection<NamedQuery>> GetNamedQueries()
+    {
+        var url = $"customers/{currentUser.GetCustomerId()}/namedQueries";
+        var response = await httpClient.GetAsync(url);
+        var namedQueries = await response.ReadAsHydraResponseAsync<HydraCollection<NamedQuery>>(jsonSerializerSettings);
+        return namedQueries;
+    }
+    
     private HttpContent ApiBody(JsonLdBase apiObject)
     {
         var jsonString = JsonConvert.SerializeObject(apiObject, jsonSerializerSettings);
--- a/src/protagonist/API.Client/IDlcsClient.cs
+++ b/src/protagonist/API.Client/IDlcsClient.cs
@@ -34,4 +34,5 @@ public interface IDlcsClient
     Task<bool> TestBatch(int batchId);
     Task<HydraCollection<Image>> GetBatchImages(int batchId, int page, int pageSize);
     Task<CustomerQueue> GetQueue();
+    Task<HydraCollection<NamedQuery>> GetNamedQueries();
 }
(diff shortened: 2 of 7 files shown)


src/protagonist/API.Client/DlcsClient.cs
```diff
@@ -317,7 +317,38 @@ public async Task<CustomerQueue> GetQueue()
...
+
+        return includeGlobal
+            ? namedQueries.Members
+            : namedQueries.Members.Where(nq => nq.Global == false);
+    }
+
+    public async Task DeleteNamedQuery(string namedQueryId)
+    {
+        var url = $"customers/{currentUser.GetCustomerId()}/namedQueries/{namedQueryId}";
+        await httpClient.DeleteAsync(url);
```
donaldgray, line 335: Feels like we should report success/failure back to use for this? (same goes for update too).

Could be frustrating to hit 'Save' and think it's went through but nothing lets you know it's failed.

src/protagonist/API.Client/DlcsClient.cs
```diff
@@ -317,7 +317,38 @@ public async Task<CustomerQueue> GetQueue()
...
+    {
+        var url = $"customers/{currentUser.GetCustomerId()}/namedQueries/{namedQueryId}";
+        await httpClient.PutAsync(url, ApiBody(new NamedQuery(){ Template = template }));
+    }
+    
+    public async Task<NamedQuery> CreateNamedQuery(NamedQuery newNamedQuery)
+    {
+        var url = $"customers/{currentUser.GetCustomerId()}/namedQueries";
+        var response = await httpClient.PostAsync(url, ApiBody(newNamedQuery));
+        var namedQuery = await response.ReadAsHydraResponseAsync<NamedQuery>(jsonSerializerSettings);
```
donaldgray, line 348: How would this be represented in UI if it wasn't a `NamedQuery` (e.g. it failed and you got an `application/problem+json` back?

src/protagonist/Portal/Features/NamedQueries/NamedQueryController.cs
```diff
@@ -0,0 +1,54 @@
...
+    public async Task<IActionResult> Create(string queryName, string queryTemplate)
+    {
+        try
+        {
+            await mediator.Send(new CreateNamedQuery() { Name = queryName, Template = queryTemplate });
+            return Ok();
+        }
+        catch (DlcsException dlcsEx)
+        {
+            return BadRequest(dlcsEx.Message);
```
donaldgray, line 51: It's not worth changing unless it would be _really_ quick to do but ideally we wouldn't just return a blanket exception code here. HTTP 400 should mean that the client can do something to resolve the issue (e.g. they've specified a duplicate name) but the API call could have failed for another reason that they can't do anything about (e.g. API is down)

donaldgray (approved):

PR #881, opened 2024-07-10
Return 400 for image requests that cannot be parsed

### LINKED ISSUE #879: Return 400 / 404 for invalid image requests
Ticket raised after looking through logs, there are a number of requests seen in logs that are returning 500 that should return a better error message. These are down to `//` in the path, e.g.

`/iiif-img/1/2/foo//full/800,/0/default.jpg`

We should detect incorrect paths and return a friendly error message.

griffri:
Resolves #879

This PR fixes an error that would be thrown by Orchestrator when attempting to parse an invalid requested image size - uses a new optional parameter introduced by https://github.com/digirati-co-uk/iiif-net/pull/45.

### DIFF
--- a/src/protagonist/DLCS.Model/DLCS.Model.csproj
+++ b/src/protagonist/DLCS.Model/DLCS.Model.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


donaldgray (approved):
No tests for `ParseForHttp` - are those exercised through integration tests?

PR #882, opened 2024-07-11
Create asset modified cleanup handler 

### LINKED ISSUE #430: Delete no-longer required delivery artifacts

JackLewis-digirati:
Resolves #430 
links [WDLS-507](https://digirati.atlassian.net/browse/WDLS-507)

This PR reimplements work done on #841 as well as implementing the work required by the asset modified cleanup handler.

It includes a new appsettings section:

```
"AssetModifiedSettings": {
  "DryRun": false,
  "EngineRoot": "https://localhost:7014"
},
```

And

```
"UpdateNotificationQueueName": "dlcsspinup-update-notification"
```

`DryRun` determines whether the application should actually remove files/db entries and is `false` by default
`EngineRoot` is used for setting the engine URI, which is used by the cleanup handler to retrieve details encoding presets from engine

Additional changes in this PR includes allowing the cleanup handler to listen to multiple SQS queues and an additional endpoint added to the engine for retrieving a dictionary of presets

`UpdateNotificationQueueName` details the name of the queue

[WDLS-507]: https://digirati.atlassian.net/browse/WDLS-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

*NOTE*: this PR has previously been approved under #856 when targeting the `develop` branch.  This PR rebases and retargets the release to `main`.

### DIFF
--- a/docs/rfcs/017-asset-modified-cleanup.md
+++ b/docs/rfcs/017-asset-modified-cleanup.md
@@ -106,6 +106,8 @@ This is that the delivery channel stays the same, but the id of the policy has c
 
 - iiif-img changed
   - `info.json` needs removed
+  - If it moves to a `use-original` policy, the derivative asset can be removed
+  - If it moves away from `use-original`, then the `/original` asset can also be removed, provided there isn't a `file` channel
 - Thumbs changed
   - Thumbs need to be removed that are no longer required.
   - `s.json` and asset application metadata should be updated - `s.json` should be updated by the reingest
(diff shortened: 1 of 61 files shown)


donaldgray (approved):
Approving as I'd already approved #856

PR #886, opened 2024-07-25
Add types to ProducesResponseType where possible

p-kaczynski:
This change is IMHO unlikely to cause any issues, as it should only affect the generated OpenAPI doc, not anything functional.

This is to allow autogeneration of an API client that helps ensure compatibility between API and client application(s).

### DIFF
--- a/src/protagonist/API/Features/Customer/Converters/CustomerConverter.cs
+++ b/src/protagonist/API/Features/Customer/Converters/CustomerConverter.cs
@@ -23,6 +23,7 @@ public static class CustomerConverter
     
     /// <summary>
     /// Converts the EF model object to a simplified JObject with just @id and @type
+    /// (PK): Could this return dedicated type, e.g. a record or sth?
     /// </summary>
     public static JObject ToCollectionForm(this DLCS.Model.Customers.Customer dbCustomer, string baseUrl)
     {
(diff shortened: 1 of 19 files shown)


src/protagonist/API/Features/Customer/Converters/CustomerConverter.cs
```diff
@@ -23,6 +23,7 @@ public static DLCS.HydraModel.Customer ToHydra(this DLCS.Model.Customers.Custome
     
     /// <summary>
     /// Converts the EF model object to a simplified JObject with just @id and @type
+    /// (PK): Could this return dedicated type, e.g. a record or sth?
```
JackLewis-digirati, line 26: is this to make the generated classes easier to understand? this was possibly written before `records`  existed, definitely possible to convert it to one, but would this just be pushing the conversion around as at some point it would need to be serialized anyway?

JackLewis-digirati (approved):

src/protagonist/API/Features/Customer/Converters/CustomerConverter.cs
p-kaczynski: Mostly so that the API description is well, describing the output, not "any JObject" :D

Issue #880, opened 2024-07-08
Error processing SFTP origin containing spaces


donaldgray:
If SFTP origin contains spaces the image is unable to be fetched from origin. Resulting error on Image is in format:

> Error! Unable to get asset ‘1/2/foo’ from origin ‘sftp://sftp.example/origin with space/image.jpg’ using SFTP

PR #887, opened 2024-07-29
Allow SFTP origins containing spaces

griffri:
Resolves DLCS-865 and https://github.com/dlcs/protagonist/issues/880

This PR fixes a bug which prevented Engine from being able to retrieve files from SFTP origins containing spaces - `originUri.AbsolutePath` would cause it to try and retrieve the image with an escaped version of the path. `originUri.LocalPath` preserves these characters while turning any escaped ones into their unescaped counterparts.

### DIFF
--- a/src/protagonist/DLCS.Repository/Strategy/SftpOriginStrategy.cs
+++ b/src/protagonist/DLCS.Repository/Strategy/SftpOriginStrategy.cs
@@ -52,8 +52,7 @@ public class SftpOriginStrategy : IOriginStrategy
 
         try
         {
-            var originPath = Uri.UnescapeDataString(originUri.AbsolutePath);
-            var outputStream = await sftpReader.RetrieveFile(connectionInfo, originPath, cancellationToken);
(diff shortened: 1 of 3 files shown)


src/protagonist/DLCS.Repository/Strategy/SftpOriginStrategy.cs
```diff
@@ -51,8 +51,9 @@ public class SftpOriginStrategy : IOriginStrategy
         ConnectionInfo connectionInfo = GetConnectionInfo(originUri, port, basicCredentials);
 
         try
-        { 
-            var outputStream = await sftpReader.RetrieveFile(connectionInfo, originUri.AbsolutePath, cancellationToken);
+        {
+            var originPath = Uri.UnescapeDataString(originUri.AbsolutePath);
```
JackLewis-digirati, line 55: this PR specifically talks about spaces, but `Uri.UnescapeDataString` feels like it should work with other escape character types - is this ok?

donaldgray (approved):
LGTM, maybe stick a note on PR as to why switch from `originUri.AbsolutePath` to `originUri.LocalPath`?

Issue #885, opened 2024-07-24
Tiles cached if source cannot be found?


donaldgray:
Reported by Wellcome (so, at time of reporting, Cantaloupe 5.0.5, kdu 8.4.1).

The lustre volume for Cantaloupe was reaching capacity so some jp2's couldn't be orchestrated. This resulted in 404's being returned from Cantaloupe. After disk space was cleared up the issue went away, however user is still seeing blurry tiles cached in browser. Switching to an anonymous browser resolved issue so it looks like issue was browser related.

Check:
* Are cache headers set for these 404 responses? If so, from where? Are they private or public?
* Are they cached anywhere else?

griffri:
I removed a newly orchestrated image from an Orchestrator volume and attempted to retrieve a tile in an anonymous browser - this resulted in a 404 with `Cache-Control: public...` and `X-Cache: HIT` headers set.

> Are they cached anywhere else?
Currently looking into this.

PR #890, opened 2024-08-01
Update Orchestrator `PathRewriteTransformer` to set correct `max-age` header values with 404 responses

griffri:
This PR updates a bug seen in #885 where Orchestrator would not apply a `cache-control: max-age=60` header value to Orchestrator's 404 responses - previously, this was only done for 500, 502, 503 and 504. `IsDownstreamError()` has been modified to check the status code against `.IsSuccessStatusCode` instead of a list.


donaldgray (approved):
LGTM, tiny change in desc - `.IsSuccessStatusCode` is a property, rather than a method

PR #894, opened 2024-08-06
Only run build + test for draft PRs

donaldgray:
Unsure if this is the best way to approach it; intent is to reduce build churn for draft PRs.

### DIFF
--- a/.github/workflows/run_build.yml
+++ b/.github/workflows/run_build.yml
@@ -11,7 +11,33 @@ on:
       - "scripts/**"
 
 jobs:
+  # if this is for a draft PR, build + run tests only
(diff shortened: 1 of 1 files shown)


donaldgray:
@JackLewis-digirati / @griffri - what are the thoughts on this? Wondering if we should bother running tests on draft PRs?

griffri:
I agree with Jack - I think it would be useful to continue allowing tests run on draft PRs. It's good for spotting overlooked issues before opening up the PR.

JackLewis-digirati (approved):

donaldgray:
Closing as out of date - can revisit if still required

Issue #898, opened 2024-08-12
Swagger docs incorrectly outputting `member` rather than `members`


donaldgray:
`HydraCollection<T>` has the following property

```cs
[JsonProperty(Order = 20, PropertyName = "member")]
public T[]? Members { get; set; }
```

in the output swagger documentation `members` is used so it looks like it's not respecting the `[JsonProperty]` name. Look at how we can get the swagger documentation correctly respecting these properties.

Is this related to #886 and use of `[ProducesResponseType]` ?

donaldgray:
Similar issue with `@type` and `type`, `@context` and `context`

PR #903, opened 2024-09-18
Respect JsonPropertyName attributes in generated swagger

garyttierney:
The swagger is currently generated using Swashbuckle, which only supports System.Text.Json attributes out of the box.

Resolves https://github.com/dlcs/protagonist/issues/898.


donaldgray (approved):
LGTM

PR #914, opened 2024-11-08
Raise notification on customer creation

donaldgray:
When a customer is created, the API will publish to an SNS topic for notifying any downstream subscribers. The message is simply `{"name": "new-name", "id": 1234}`

`CustomerNotificationSender` seems superfluous but I've left it in as it follows the same model as raising asset-modified notifications and gives us a hook to enrich the notification should we need to.

Failure to raise this notification will _not_ prevent customer creation or result in a 500 but instead log a warning which will need to be picked up and dealt with.

> [!IMPORTANT]
> Requires a new SNS notification to be setup for API to publish to

### DIFF
--- a/src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
+++ b/src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
@@ -204,7 +204,7 @@ public class TopicPublisherTests
     }
 
     [Fact]
-    public async Task PublishToCustomerCreatedTopic_ReturnsTrue_IfNoArn()
+    public async Task PublishToCustomerCreatedTopic_ReturnsFalse_IfNoArn()
     {
         // Arrange
         var notification = new CustomerCreatedNotification(new Customer());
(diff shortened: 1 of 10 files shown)


src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
```diff
@@ -1,4 +1,4 @@
-﻿using DLCS.Model.Messaging;
+﻿using DLCS.Model.Customers;
```
JackLewis-digirati, line 1: what was the reason for moving this to `Customer`? Mainly asking as there's methods in here for publishing to the `AssetModified` topic

src/protagonist/DLCS.AWS.Tests/SNS/TopicPublisherTests.cs
```diff
@@ -197,6 +202,55 @@ public async Task PublishToAssetModifiedTopicBatch_SuccessfullyPublishesSingleBa
                                                          b.PublishBatchRequestEntries.Count == 2),
                 A<CancellationToken>._)).MustHaveHappened();
     }
+
+    [Fact]
+    public async Task PublishToCustomerCreatedTopic_ReturnsTrue_IfNoArn()
```
JackLewis-digirati, line 207: I think the name of this test might need updating - the result returns false

JackLewis-digirati (approved):

src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
donaldgray: It hasn't moved - this is just updating a reference. The new `CustomerCreatedNotification` takes a `Customer` object. `using DLCS.Model.Messaging` was redundant.

PR #916, opened 2024-11-14
Add no-store to the Cache-Control header on auth responses

JackLewis-digirati:
Related to #833

This PR adds a `Cache-Control` header of `no-cache,no-store` to the following endpoints:

- `auth/{customer}/clickthrough`
- `auth/{customer}/token`
- `auth/{customer}/{authService}`
- `auth/{customer}/{authService}/logout`

This should help to avoid issues where auth responses are cached incorrectly

**NOTE:** this header has not been added to `auth/v2/probe/{customer}/{space}/{image}` as this resource already has a `Cache-Control` value of `private`


src/protagonist/Orchestrator.Tests/Integration/AuthHandlingTests.cs
```diff
@@ -54,6 +54,8 @@ public async Task Get_Clickthrough_UnknownCustomer_Returns400()
         
         // Assert
         response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+        response.Headers.CacheControl!.NoCache.Should().BeTrue();
+        response.Headers.CacheControl.NoStore.Should().BeTrue();
```
donaldgray, line 58: Is it okay to cache a 400 for bad customer, even if it's briefly? Alleviate any load on server if we keep getting repeated requests for a customer that's deleted?

src/protagonist/Orchestrator.Tests/Integration/AuthHandlingTests.cs
```diff
@@ -67,6 +69,8 @@ public async Task Get_UnknownRole_Returns404()
         
         // Assert
         response.StatusCode.Should().Be(HttpStatusCode.NotFound);
+        response.Headers.CacheControl!.NoCache.Should().BeTrue();
+        response.Headers.CacheControl.NoStore.Should().BeTrue();
```
donaldgray, line 73: As above - if a role is deleted should we briefly be able to cache a 404?

donaldgray (commented):
As the issue that we encountered is related to the [`AccessTokenResponse`](https://iiif.io/api/auth/1.0/#access-token-service) - is it worth adding tests to confirm the appropriate headings are set there? Would pick up if the attribute was somehow removed

src/protagonist/Orchestrator.Tests/Integration/AuthHandlingTests.cs
JackLewis-digirati: I think that might require a custom attribute as the `ResponseCacheAttribute` doesn't appear to support doing that - makes sense to have it though

garyttierney:
Btw, `no-cache,no-store` is equivalent to `no-store` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#use_cases

JackLewis-digirati:
should be changed to use just `no-store` - thanks Gary

donaldgray (approved):

PR #926, opened 2024-12-02
Adding ability for engine to notify on batch completion

JackLewis-digirati:


src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -163,10 +172,17 @@ private void UpdateAsset(Asset asset, bool ingestFinished)
...
-            .Where(b => b.Id == batchId && b.Count == b.Completed + b.Errors)
-            .UpdateFromQueryAsync(b => new Batch { Finished = DateTime.UtcNow }, cancellationToken);
+    private async Task<(int rowsAffected, IQueryable<Batch> batches)> TryFinishBatch(int batchId,
+        CancellationToken cancellationToken)
+    {
+        var batches = dlcsContext.Batches
+            .Where(b => b.Id == batchId && b.Count == b.Completed + b.Errors);
+        
+        var rowsAffected =
+            await batches.UpdateFromQueryAsync(b => new Batch { Finished = DateTime.UtcNow }, cancellationToken);
```
donaldgray, line 182: The value in using `UpdateFromQueryAsync` is that you don't need to load any entities into context first, it just issues an update statement to DB directly. ie in 'before' diff it would be `update batches set finished = now() where id = @id and count = completed + errors`

However, this is now _maybe_ loading the batch but _always_ issuing the `update`. Might be cleaner written to always load the Batch, and let EF take care of working out whether to issue the update?

```cs
var batch = dlcsContext.Batches.Find(batchId);
// if batch is complete, set Finished
```

src/protagonist/Engine/Infrastructure/ServiceCollectionX.cs
```diff
@@ -164,4 +168,13 @@ public static IServiceCollection ConfigureHealthChecks(this IServiceCollection s
 
         return services;
     }
+
+    public static IServiceCollection AddTopicNotifiers(this IServiceCollection services)
+    {
+        services
+            .AddScoped<IBatchCompletedNotificationSender, BatchCompletedNotificationSender>()
+            .AddScoped<IBatchCompletedNotificationSender, BatchCompletedNotificationSender>();
```
donaldgray, line 176: Duplicate dependency. Does this method need to be called from API too, for raising notifications after `/test`?

src/protagonist/DLCS.AWS/SNS/Messaging/IBatchCompletedNotificationSender.cs
```diff
@@ -0,0 +1,16 @@
+﻿using DLCS.Model.Assets;
+
+namespace DLCS.AWS.SNS.Messaging;
+
+public interface IBatchCompletedNotificationSender
+{
+    /// <summary>
+    /// Broadcast multiple batch completed notification
+    /// </summary>
+    Task SendBatchCompletedMessages(IQueryable<Batch> completedBatches, CancellationToken cancellationToken = default);
```
donaldgray, line 10: This shouldn't be possible - any action that can 'finish' a batch will be done on a single asset.

src/protagonist/Engine/Data/EngineAssetRepository.cs
donaldgray: Just thinking about this - the reason for issueing the UPDATE direct to the DB was to avoid multiple engine instances attempting to update the Batch at the same time. By the time instance 1 loads it instance 2 may have completed it. Is it better to issue update and only load the batch if rowsAffected as a value?

JackLewis-digirati: That potentially feels safer, so I'll change it to only load if `rowsAffected > 0`

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -103,7 +108,7 @@ private async Task<bool> NonBatchedSave(CancellationToken cancellationToken)
         return updatedRows > 0;
     }
 
-    private async Task<bool> BatchSave(int batchId, CancellationToken cancellationToken)
+    private async Task<bool> BatchSave(Asset asset, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
JackLewis-digirati: think this got changed based on an earlier iteration and not moved back - changing to use `batchId` now

PR #932, opened 2025-01-09
Introduce "BatchAssets"

### LINKED ISSUE #491: Revisit image batches
The "Batches" table is a loose and transient grouping of images - is this enough?

On creation of a batch:
* each image stores the associated batch number.
* each batch maintains a count of total/complete/failed images.

If an image is reingested, it is associated with a new batch so that link is lost. The batch counts do not keep track of which items have completed or failed - they only increment (which has lead to some bugs where complete + error > total). 

Due to the transient nature of the link between image and batch we need the `/test` endpoint for batches. This will mark a batch as superseded or complete, or update it's counts to reflect what is actually stored in the image table.

We should reconsider the use case for Batches in general. They are useful to ingest X number of images and track progress - is that enough? Do we need to maintain a richer/longer running record of images associated with a batch? Should we change how `/test` works? Or automate it so that integrations don't need to concern themselves with it?

One option to consider is introducing a BatchImages table. This could get large but could be periodically cleaned out over time. This would also allow us to retry images that failed due to a transient error (#355).

donaldgray:
Resolves #491 

Adds a new table to database, `BatchAssets`, which tracks which assets have been in each batch over time.

The main points of change were:
* API - when processing a batch, save a `BatchAsset` record for every item in batch. 
* API now includes the batchId in notifications sent to Engine, both for sync and async ingests. Previously we read the value from DB but this isn't necessarily the batch this ingest is related to. Similarly, ET metadata now includes batchId. Batch is `0` rather than `null` for historical reasons but rewrite `0` -> `null` at API boundary to simplify Engine.
* Refactored when asset modified message is sent as discovered a bug where notifications could be raised when data is will not be persisted to DB - could have happened if 1 asset in a batch caused a failure in API.
* API - new endpoint `/customers/{customer}/queue/batches/{batchId}/assets` which returns details of all assets in specified batch.
* Engine - largest changes were in `EngineAssetRepository`. We now calculate the `Batches` totals based in `BatchAsset` records, this simplified the implementation and removed the need to Transactions. We can now use the EF transaction as part of `.SaveChanges()` once records are saved we can run the update script to set the `Batches` totals independently on a new connection - ensures we're always using current state of table.

> [!IMPORTANT]  
> This PR contains a migration
> When deployed, any in-flight ingest operations (particularly AV) may not be handled correctly, leading to batch totals being inaccurate.

### DIFF
--- a/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
+++ b/src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
@@ -56,29 +56,15 @@ public class CreateBatchOfImagesHandler : IRequestHandler<CreateBatchOfImages, M
         this.assetNotificationSender = assetNotificationSender;
         this.logger = logger;
     }
-
+    
     public async Task<ModifyEntityResult<Batch>> Handle(CreateBatchOfImages request,
         CancellationToken cancellationToken)
     {
-        // TODO - we may need to support non-Image assets here 
-        if (request.IsPriority)
-        {
-            if (request.AssetsBeforeProcessing.Any(a =>
-                    a.Asset.Family != AssetFamily.Image && !a.Asset.HasDeliveryChannel(AssetDeliveryChannels.Image) &&
-                    !MIMEHelper.IsImage(a.Asset.MediaType)))
-            {
-                return ModifyEntityResult<Batch>.Failure("Priority queue only supports image assets",
-                    WriteResult.FailedValidation);
-            }
-        }
+        var priorityValidation = ValidatePriorityQueueRequest(request);
+        if (priorityValidation != null) return priorityValidation;
         
-        var (exists, missing) = await DoAllSpacesExist(request.CustomerId, request.AssetsBeforeProcessing.Select(a => a.Asset), cancellationToken);
-        if (!exists)
-        {
-            var spaceList = string.Join(", ", missing);
-            return ModifyEntityResult<Batch>.Failure($"The following space(s) could not be found: {spaceList}",
-                WriteResult.FailedValidation);
-        }
+        var spaceValidation = await ValidateAllSpaces(request, cancellationToken);
+        if (spaceValidation != null) return spaceValidation; 
 
         bool updateFailed = false;
         var failureMessage = string.Empty;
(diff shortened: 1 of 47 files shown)


src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -88,13 +81,21 @@ public async Task<bool> UpdateIngestedAsset(Asset asset, ImageLocation? imageLoc
         }
     }
 
-    public ValueTask<Asset?> GetAsset(AssetId assetId, CancellationToken cancellationToken = default)
-        => new(dlcsContext.Images.IncludeDeliveryChannelsWithPolicy()
-            .SingleOrDefaultAsync(i => i.Id == assetId, cancellationToken));
+    public ValueTask<Asset?> GetAsset(AssetId assetId, int? batchId, CancellationToken cancellationToken = default)
```
JackLewis-digirati, line 84: would it be worth doing `int? batchId = null`?

JackLewis-digirati (approved):

src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
```diff
@@ -56,29 +56,15 @@ public CreateBatchOfImagesHandler(
         this.assetNotificationSender = assetNotificationSender;
         this.logger = logger;
     }
-
+    
     public async Task<ModifyEntityResult<Batch>> Handle(CreateBatchOfImages request,
         CancellationToken cancellationToken)
     {
-        // TODO - we may need to support non-Image assets here
```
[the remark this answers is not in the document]
donaldgray: We might need to support non-image assets in the priority queue but if we do I think it best be specified via an issue rather than a `TODO` comment that's a few years old (this is from https://github.com/dlcs/protagonist/commit/57f91a57192267f5176e7153f0b74c7f89e8deb2), I think I left it after a convo with Tom but there's no context as to _why_ we might need that hence issue being better

src/protagonist/Engine/Data/EngineAssetRepository.cs
donaldgray: I don't think so, no. This `GetAsset()` call is specific for the Engine and every entrypoint into the Engine (sync http, async from queue or transcode complete) will have both the `assetId` and `batchId` vars so it'll always be called with both.

PR #933, opened 2025-01-10
Handle NQ 'batch' parameter

### LINKED ISSUE #930: Support Batch as NamedQuery parameter
Extend the list of supported NQ parameters to include `?batch`. This is a comma delimited list of values, if multiple values are selected there may be duplicate images returned so we may need to dedup.

`?batch` won't use the Image.Batch property as this is only the most recent batch, instead it will use the new `BatchAsset` table to derive assets.

donaldgray:
Resolves #930

This introduces a new NQ parameter, `batch`. This uses the `BatchAssets` table only (so won't pull images from batches created before the introduction of `BatchAssets`).

`batch` can be a single integer, or multiple comma delimited values. e.g. the NQ "batch-test" with template `batch=p1&assetOrder=n1;n2&s1=p2&space=p3`

can be called with the following:
* `/iiif-resource/{cust}/batch-test/1234/string1-val/99`
* `/iiif-resource/{cust}/batch-test/1234,9999,2222/string1-val/99`

It's likely this is just for internal use but I've included it in the base parser so the parameter is available to all NQ projections (IIIF, zip and PDF).

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/NamedQueries/ParsedNamedQuery.cs
+++ b/src/protagonist/DLCS.Model/Assets/NamedQueries/ParsedNamedQuery.cs
@@ -55,6 +55,11 @@ public class ParsedNamedQuery
     public long? Number3 { get; set; }
     
     /// <summary>
+    /// Value of "batches" after parsing
+    /// </summary>
+    public int[]? Batches { get; set; } 
+    
+    /// <summary>
     /// The name of the namedQuery this object was parsed from.
     /// </summary>
     public string NamedQueryName { get; set; }
@@ -99,7 +104,8 @@ public class ParsedNamedQuery
         String3,
         Number1,
         Number2,
-        Number3
+        Number3,
+        Batch,
     }
 
     public enum OrderDirection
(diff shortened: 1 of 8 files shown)


src/protagonist/DLCS.Model/Assets/NamedQueries/ParsedNamedQuery.cs
```diff
@@ -99,7 +104,7 @@ public enum QueryMapping
         String3,
         Number1,
         Number2,
-        Number3
+        Number3,
```
JackLewis-digirati, line 107: is this needed?

JackLewis-digirati (approved):

src/protagonist/DLCS.Model/Assets/NamedQueries/ParsedNamedQuery.cs
donaldgray: Technically not, no. I appended `Batch` to this list in error and removed it after realising we don't need to support ordering by `Batch`. However, I don't think it's worth a new commit to remove .

PR #938, opened 2025-01-21
Create batch asset when using PUT for a timebased asset

### LINKED ISSUE #937: Creating timebased assets via PUT does not create a `BatchAsset` row in the DB
related to #491

As part of running the automated tests it was noted that timebased assets were timing out and not completing correctly.

After some investigation it was discovered that assets created via `PUT` do not run through the standard process for batch creation and instead go through the process found here - https://github.com/dlcs/protagonist/blob/main/src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs#L107

This method does not have the handling introduced to add assets to the `BatchAsset` table so it ultimatgely causes this issue

The `BatchAsset` table is used to tell the engine that the asset has a batch that needs to be completed, so without this record the batch does not get finished in the database causing checks for batch completion to no longer work

JackLewis-digirati:
Resolves #937

This PR adds a call to create a batch asset record when ingesting timebased assets via PUT.

### DIFF
--- a/src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
+++ b/src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Data;
 using System.Net;
 using API.Exceptions;
@@ -8,7 +7,6 @@ using API.Infrastructure.Messaging;
 using API.Infrastructure.Requests;
 using DLCS.Core;
 using DLCS.Core.Collections;
-using DLCS.Core.Types;
 using DLCS.Model.Assets;
 using DLCS.Model.Messaging;
 using DLCS.Model.Spaces;
--- a/src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
+++ b/src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
@@ -109,8 +109,10 @@ public class CreateOrUpdateImageHandler : IRequestHandler<CreateOrUpdateImage, M
             if (modifyEntityResult.Entity!.Family == AssetFamily.Timebased)
             {
                 await batchRepository.CreateBatch(modifyEntityResult.Entity.Customer,
-                    modifyEntityResult.Entity.AsList(), cancellationToken);
-                await batchRepository.CreateBatchAsset(modifyEntityResult.Entity!, cancellationToken);
+                    modifyEntityResult.Entity.AsList(), postCreate: b =>
+                    {
+                        b.AddBatchAsset(modifyEntityResult.Entity!.Id);
+                    }, cancellationToken: cancellationToken);
             }
             
             await transaction.CommitAsync(cancellationToken);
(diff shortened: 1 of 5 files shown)


src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
```diff
@@ -114,6 +114,11 @@ public async Task<ModifyEntityResult<Asset>> Handle(CreateOrUpdateImage request,
 
         if (modifyEntityResult.IsSuccess)
         {
+            if (modifyEntityResult.Entity!.HasSingleDeliveryChannel(AssetDeliveryChannels.Timebased))
```
donaldgray, line 117: I don't think this is correct, if an asset has both `["iiif-av", "file"]` it would fail this test but we'd still want to create a batch asset. 

This logic also deviates from when the batch is created - is it simpler to keep them together on the same code path? That way we would ensure that if a batch is created, we'd always create a `BatchAsset` (and conversely never create a `BatchAsset` if we don't create a `Batch`)?

src/protagonist/DLCS.Repository/Assets/BatchRepository.cs
```diff
@@ -44,4 +45,23 @@ public async Task<Batch> CreateBatch(int customerId, IReadOnlyList<Asset> assets
 
         return batch;
     }
+    
+    /// <inheritdoc />
+    public async Task<BatchAsset> CreateBatchAsset(Asset asset, CancellationToken cancellationToken)
+    {
+        asset.Batch.ThrowIfNull("Saving a batch asset requires a batch");
```
donaldgray, line 52: Argument for `ThrowIfNull` isn't an error message, it's the parameter name for [`ArgumentNullException`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-8.0#system-argumentnullexception-ctor(system-string))

donaldgray (changes requested):
As mentioned on comment for `CreateOrUpdateImage`, the "create batch" and "create batch asset" steps are happening in 2 separate places, with slightly different logic - can they be brought together? Should mean we can use same `.AddBatchAsset()` extension as is used when dealing with batch creation in `CreateBatchOfImages`

src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
```diff
@@ -99,21 +99,20 @@ public async Task<ModifyEntityResult<Asset>> Handle(CreateOrUpdateImage request,
...
-            },
-            cancellationToken
+            cancellationToken: cancellationToken
         );
         
         var modifyEntityResult = processAssetResult.Result;
 
         if (modifyEntityResult.IsSuccess)
         {
+            if (modifyEntityResult.Entity!.HasDeliveryChannel(AssetDeliveryChannels.Timebased))
```
donaldgray, line 109: Should this be `if (updatedAsset.Family == AssetFamily.Timebased)` to keep consistent with previous logic?

donaldgray (approved):

PR #939, opened 2025-01-21
Save AddBatchAsset after call, instead of as a postIngest task

### LINKED ISSUE #937: Creating timebased assets via PUT does not create a `BatchAsset` row in the DB

JackLewis-digirati:
Resolves #937

Additional fix to solve the batch id not being set correctly on the asset.

The process is now this:

create batch -> set up asset with batch id + create (but not save) batchAsset -> save asset + batchAsset

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/IBatchRepository.cs
+++ b/src/protagonist/DLCS.Model/Assets/IBatchRepository.cs
@@ -17,8 +17,8 @@ public interface IBatchRepository
     /// <param name="customerId">Id of customer to create batch for.</param>
     /// <param name="assets"></param>
     /// <param name="cancellationToken">Current cancellation token</param>
-    /// <param name="preCreate">Optional action, will be called after Batch is created but before saving in db</param>
+    /// <param name="postCreate">Optional action, will be called after Batch is created but before saving in db</param>
     /// <returns>Created Batch object</returns>
     Task<Batch> CreateBatch(int customerId, IReadOnlyList<Asset> assets, CancellationToken cancellationToken,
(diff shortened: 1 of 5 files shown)


src/protagonist/DLCS.Model/Assets/IBatchRepository.cs
```diff
@@ -17,8 +17,8 @@ public interface IBatchRepository
     /// <param name="customerId">Id of customer to create batch for.</param>
     /// <param name="assets"></param>
     /// <param name="cancellationToken">Current cancellation token</param>
-    /// <param name="postCreate">Optional action, will be called after Batch is created but before saving in db</param>
+    /// <param name="preCreate">Optional action, will be called after Batch is created but before saving in db</param>
```
donaldgray, line 20: Should this maintain `postCreate` name? `"will be called after Batch is created but before saving in db"`

donaldgray (approved):

griffri:
Tested - working. Moving into 'done'.

PR #942, opened 2025-01-30
RFC for presentation-dlcs query

donaldgray:


docs/rfcs/019-presentation-dlcs.md
```diff
@@ -0,0 +1,169 @@
...
+| Check if assets new            | Query `canvasPainting` to check if asset exists in iiif-p (2 do), fallback to `POST customers/{id}/allImages` request for rest (1 exists, 1 doesn't) | As Without                                                                                                                                                                     |
+| 1 asset new                    | `POST customers/{id}/queue` with 1 asset (batch 66666 created)                                                                                       | 1 assets is new, `POST customers/{id}/queue` with 1 asset (batch 66666 created). 3 assets exist, `PATCH customers/{id}/space/{id}/images` to add "m123" to "manifest" property |
+| Get `"assets"` for return body | `POST customers/{id}/allImages` with 4 Ids                                                                                                           | `GET customers/{id}/allImages?q={"manifest": ["m123"]}`                                                                                                                        |
```
JackLewis-digirati, line 137: is it worth removing the spaces from this syntax, or using `%20`?

JackLewis-digirati (approved):

docs/rfcs/019-presentation-dlcs.md
donaldgray: This syntax is from docs at https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/asset-queries#query-object - we'd likely replace spaces when implementing but I'll leave them here as I think it makes it easier to read.

tomcrane:
Another way of looking at the suggestion is:

If everything was one schema, the canvas_paintings *_is_* the joining table between the manifests table and the images table (assets). It has more cols than a simple joining table, because it provides context about how they are joined.

If we can't query across, we need a little bit of the other db in both dbs.

In the presentation db, manifest + canvas_paintings.asset_id gives us the `id` of any *asset*; we have that half of the join.
In the asset/dlcs db, asset.manifest[] gives us the `id` od any *manifest*; we now have the other half of the join.

They are actually _**symmetrical**_ because the pgsql array type gives us the effect of an assets table and an assets_manifests joining table, we just don't see the latter as a real table.

Having the symmetry allows us to do "joins" over HTTP but in one operation and without vast list payloads to generate `where in (...)` queries.

PR #944, opened 2025-01-31
Getting image delivery channels to work with the default policy

### LINKED ISSUE #931: `"deliveryChannels"` should be optional for updates
Currently the `"deliveryChannels"` property is required when updating an asset, whether that be via a batch or PUT.

This keeps the logic in API simple but pushes the complexity to the API consumer. It's then more difficult to submit updates as you either need to know what delivery-channels have been configured or issue a GET first before constructing an update.

`"deliveryChannels"` should be optional for both create and update operations. To accommodate this we need a mechanism to allow delivery channels for an existing asset to be "re-defaulted". This mechanism is a new special channel, `"default"` (similar to the special `"none"` channel). When `"none"` channel is specified we place a marker in the database, however `"default"` _doesn't_ place a marker but instead sets whatever the actual defaults are.

The possible scenarios are now:

* `"deliveryChannels": []`:  400 bad request whether new or existing asset (consumer needs to be more explicit)
* `"deliveryChannels": [{ "channel: "none" }]`: store 'none' marker in DB whether new or existing asset
* `"deliveryChannels": [ ..explicit, valid channels.. ]`: use specified values
* `"deliveryChannels": null`
  * If new asset use defaults 
  * If existing asset, use delivery channels already configured for asset
* `"deliveryChannels": [{ "channel: "default" }]`: 
  * If new asset use defaults (equivalent to `null`)
  * If existing asset, reset to defaults (different from `null`)

_Note that it's only the behaviour of the last 2 bullets that change - everything else is as-is but listing for completeness_

JackLewis-digirati:
Resolves #931 

This PR implements the new `default` delivery channel which does the following:

- if `default` specified
  - if new asset, act like `null` delivery channel
  - if updated asset, ingest with customer specified default delivery channels
- if `null` specified
  - if new asset, act the same as previous behavior
  - if updated asset, reuse the previous delivery channels

### DIFF
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -39,7 +39,7 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
 {
     private readonly DlcsContext dbContext;
     private readonly HttpClient httpClient;
-    private static readonly IAssetNotificationSender NotificationSender = A.Fake<IAssetNotificationSender>();
+    private static readonly IAssetNotificationSender AssetNotificationSender = A.Fake<IAssetNotificationSender>();
     private readonly IAmazonS3 amazonS3;
     private static readonly IEngineClient EngineClient = A.Fake<IEngineClient>();
 
@@ -57,7 +57,7 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
             .WithLocalStack(storageFixture.LocalStackFixture)
             .WithTestServices(services =>
             {
-                services.AddSingleton<IAssetNotificationSender>(_ => NotificationSender);
+                services.AddSingleton(_ => AssetNotificationSender);
                 services.AddScoped<IEngineClient>(_ => EngineClient);
                 services.AddAuthentication("API-Test")
                     .AddScheme<AuthenticationSchemeOptions, TestAuthHandler>(
@@ -147,6 +147,46 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
             .Should().Satisfy(
                 i => i.Channel == AssetDeliveryChannels.None &&
                      i.DeliveryChannelPolicyId == KnownDeliveryChannelPolicies.None);
+    }
+    
+    [Fact]
+    public async Task Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToDefault()
+    {
+        var customerAndSpace = await CreateCustomerAndSpace();
+
+        var assetId = new AssetId(customerAndSpace.customer, customerAndSpace.space, nameof(Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToDefault));
+        var hydraImageBody = $@"{{
+            ""@type"": ""Image"",
+            ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
+            ""family"": ""I"",
+            ""mediaType"": ""image/tiff"",
+            ""deliveryChannels"": [
+            {{
+                ""channel"": ""default""
+            }}]
+        }}";
+        A.CallTo(() =>
+                EngineClient.SynchronousIngest(
+                    A<Asset>.That.Matches(r => r.Id == assetId),
+                    A<CancellationToken>._))
+            .Returns(HttpStatusCode.OK);
+        
+        // act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customerAndSpace.customer).PutAsync(assetId.ToApiResourcePath(), content);
+
+        // assert
+        response.StatusCode.Should().Be(HttpStatusCode.Created);
+        response.Headers.Location.PathAndQuery.Should().Be(assetId.ToApiResourcePath());
+        
+        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
+        asset.Id.Should().Be(assetId);
+        asset.MaxUnauthorised.Should().Be(-1);
+        asset.ImageDeliveryChannels
+            .Should().HaveCount(2).And.Contain(i =>
+                i.Channel == AssetDeliveryChannels.Image &&
+                i.DeliveryChannelPolicyId == KnownDeliveryChannelPolicies.ImageDefault).And
+            .Contain(i => i.Channel == AssetDeliveryChannels.Thumbnails);
     }
     
     [Fact]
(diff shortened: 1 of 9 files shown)


src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -91,7 +92,15 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
...
-            else if (assetBeforeProcessing.DeliveryChannelsBeforeProcessing.IsNullOrEmpty() && alwaysReingest)
+            else if (assetBeforeProcessing.DeliveryChannelsBeforeProcessing == null && alwaysReingest)
+            {
+                assetBeforeProcessing.DeliveryChannelsBeforeProcessing = assetFromDatabase.ImageDeliveryChannels
+                    .Select(d => new DeliveryChannelsBeforeProcessing(d.Channel, d.DeliveryChannelPolicy.Name))
+                    .ToArray();
+
+                updatedDeliveryChannels = true;
+            }
+            else if (assetBeforeProcessing.DeliveryChannelsBeforeProcessing?.Length == 0 && alwaysReingest)
```
donaldgray, line 103: To catch `"deliveryChannels": []` we could use fluent-validator before doing any work.

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -91,7 +92,15 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
 
                 counts.CustomerStorage.NumberOfStoredImages++;
             }
-            else if (assetBeforeProcessing.DeliveryChannelsBeforeProcessing.IsNullOrEmpty() && alwaysReingest)
+            else if (assetBeforeProcessing.DeliveryChannelsBeforeProcessing == null && alwaysReingest)
+            {
+                assetBeforeProcessing.DeliveryChannelsBeforeProcessing = assetFromDatabase.ImageDeliveryChannels
```
donaldgray, line 97: Is this going beyond what `DeliveryChannelsBeforeProcessing` were intended for? They were originally a carrier of "brief" data about delivery-channels to be set (channel + policyName) but this is taking the fully fleshed out data from DB and then reducing it down to the "brief" data and then passing it to the `DeliveryChannelProcessor`, which would then reprocess it (when we already know it?)

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
```diff
@@ -73,20 +72,20 @@ private bool DeliveryChannelsRequireReprocessing(Asset originalAsset, DeliveryCh
         return false;
     }
 
-    private async Task<bool> SetImageDeliveryChannels(Asset asset, DeliveryChannelsBeforeProcessing[] deliveryChannelsBeforeProcessing, bool isUpdate)
+    private async Task<bool> SetImageDeliveryChannels(Asset asset, DeliveryChannelsBeforeProcessing[]? deliveryChannelsBeforeProcessing, bool isUpdate)
```
donaldgray, line 75: Line 95 in this class is giving a nullable warning - by this stage do we need to have `null` passed down? Would it simplify to `?? Array.Empty` and treat null + empty as the same without worrying?

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
```diff
@@ -73,20 +72,20 @@ private bool DeliveryChannelsRequireReprocessing(Asset originalAsset, DeliveryCh
...
-            logger.LogTrace("Asset {AssetId} is new, resetting ImageDeliveryChannels", assetId);
+            logger.LogTrace("Setting delivery channels for asset {AssetId}", assetId);
             asset.ImageDeliveryChannels = new List<ImageDeliveryChannel>();
-            
-            // Only valid for creation - set image delivery channels to default values for media type
-            if (deliveryChannelsBeforeProcessing.IsNullOrEmpty())
+
+            if (deliveryChannelsBeforeProcessing.IsNullOrEmpty() || isDefaultChannel)
             {
-                logger.LogDebug("Asset {AssetId} is new, no deliveryChannels specified. Assigning defaults for mediaType",
```
donaldgray, line 88: Worth keeping log saying we're resetting delivery-channels to aid in debugging scenarios?

src/protagonist/API/Features/Image/AssetBeforeProcessing.cs
```diff
@@ -4,15 +4,15 @@ namespace API.Features.Image;
...
+    public AssetBeforeProcessing(Asset asset, DeliveryChannelsBeforeProcessing[]? deliveryChannelsBeforeProcessing)
     {
         Asset = asset;
         DeliveryChannelsBeforeProcessing = deliveryChannelsBeforeProcessing;
     }
 
     public Asset Asset { get; }
 
-    public DeliveryChannelsBeforeProcessing[] DeliveryChannelsBeforeProcessing { get; }
+    public DeliveryChannelsBeforeProcessing[]? DeliveryChannelsBeforeProcessing { get; set; }
```
donaldgray, line 15: If we don't set this to data loaded from DB can it go back to being immutable? This makes it easier to reason about as it's set once then passed down

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -1073,14 +1135,80 @@ public async Task Put_Existing_Asset_Returns400_IfDeliveryChannelsNull()
...
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(99).PutAsync(assetId.ToApiResourcePath(), content);
+        
+        // assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        
+        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
+        asset.Id.Should().Be(assetId);
+        asset.MaxUnauthorised.Should().Be(-1);
+        asset.ImageDeliveryChannels.Should().HaveCount(3).And.Contain(i =>
```
donaldgray, line 1154: This could be simplified to `asset.ImageDeliveryChannels.Should().BeEquivalentTo(deliveryChannels);`

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -149,6 +149,46 @@ public async Task Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToNone(
                      i.DeliveryChannelPolicyId == KnownDeliveryChannelPolicies.None);
     }
     
+    [Fact]
```
donaldgray, line 152: Test for `"deliveryChannel": []`?

JackLewis-digirati: there's an existing test that covers this here - https://github.com/dlcs/protagonist/blob/main/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs#L1087

this should be changed slightly with the latest changes

donaldgray (approved):

PR #948, opened 2025-02-03
Check database before deleting derivatives

### LINKED ISSUE #940: Cleanup handler can delete derivatives if asset is ingested again
If an asset is deleted via the API and then ingested shortly afterwards it can lead to the CleanupHandler removing derivatives (e.g. thumbs/jpeg2000 etc) that were created as part of the follow up ingest request. The result is an `"Images"` db record that looks fine, has no recorded error etc but any orchestrator request returns a 404.

Example timeline of issue:
* `00:00` API request to delete existing Asset (`POST /customers/{id}/deleteImages` OR `DELETE /customers/{id}/spaces/{id}/images/{id}`)
  * Delete DB rows (`image`, `imageLocation`, `imageStorage`)
  * Raise an asset deleted notification.
* `00:02` API request to ingest Asset (via batch or direct PUT)
  * Create DB rows (`image`, `imageLocation`, `imageStorage`)
  * Derivatives created and uploaded to S3.
* `00:04` CleanupHandler picks up asset deleted notification
  * Derivatives deleted from S3
* `00:10` Orchestrator request for assets result in 404 as derivative not found in S3.

This fix should be as simple as the `CleanupHandler` checks for the existence of a DB record for that Asset, if it exists then it's subsequently been recreated and shouldn't be deleted (log a message).

JackLewis-digirati:
resolves #940

This PR updates the `CleanupHandler` to check that a database record exists before deleting, which helps avoid issues with a deleted asset being ingested again before the `CleanupHandler` runs, causing an asset which should be there to have it's derivatives removed

### DIFF
--- a/src/protagonist/CleanupHandler/AssetDeletedHandler.cs
+++ b/src/protagonist/CleanupHandler/AssetDeletedHandler.cs
@@ -64,7 +64,7 @@ public class AssetDeletedHandler : IMessageHandler
 
         logger.LogDebug("Processing delete notification for {AssetId}", request.Asset.Id);
         
-        // if the itme exists in the db, assume the asset has been reingested after delete
+        // if the item exists in the db, assume the asset has been reingested after delete
         if (await assetRepository.CheckExists(request.Asset.Id))
         {
             logger.LogInformation("asset {Asset} can be found in the database, so will not be deleted", request.Asset.Id);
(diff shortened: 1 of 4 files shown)


src/protagonist/CleanupHandler/AssetDeletedHandler.cs
```diff
@@ -59,6 +63,13 @@ public async Task<bool> HandleMessage(QueueMessage message, CancellationToken ca
         if (request?.Asset?.Id == null) return false;
 
         logger.LogDebug("Processing delete notification for {AssetId}", request.Asset.Id);
+        
+        // if the itme exists in the db, assume the asset has been reingested after delete
```
donaldgray, line 67: tiny typo

PR #964, opened 2025-02-26
Non-image content resource RFC

donaldgray:
RFC on how we can express non-image delivery channels on IIIF manifests.


tomcrane (approved):
(reviewed via Slack)

PR #972, opened 2025-02-28
Store Transcode metadata in AssetApplicationMetadata table

### LINKED ISSUE #966: Save transcodes to `AssetApplicationMetadata`
`AssetApplicationMetadata` table needs to store the available transcodes. 

Suggestion of what we should store:

```json
[
   { "location": "s3://path/to/key.webm", "mediaType": "audio/webm", "width": 123, "height": 123, "duration": 123.42 },
   { "location": "s3://path/to/key.mp3", "mediaType": "audio/mp3", "width": 122, "height": 122, "duration": 123.44 }
]
```

Where:
* The 'location' is where the transcode for that extension is stored in S3
* The 'mediaType' is the mediaType of that transcode (this will be derived from the extension but saves multiple repeated calculations of it on read).
* Dimensions from elastic-transcoder result. Note that dimensions may differ between transcodes due to codec value.

donaldgray:
Resolves #966

Add `"AVTranscodes"` metadatatype, which stores a `AVTranscode[]`. Populating this involved a refactor of `TimebasedIngesterCompletion`.

Example metadata for single audio output:
```json
[
    {
        "d": 15000,
        "h": null,
        "l": "s3://bucket/7/8/foo/full/max/default.mp3",
        "n": "System preset: Audio MP3 - 128k",
        "w": null,
        "ex": "mp3",
        "mt": "audio/mp3"
    }
]
```

Example metadata for multiple video output:
```json
[
    {
        "d": 16000,
        "h": 720,
        "l": "s3://bucket/7/8/foo/full/full/max/max/0/default.webm",
        "n": "System preset: Webm 720p",
        "w": 1280,
        "ex": "webm",
        "mt": "video/webm"
    },
    {
        "d": 16000,
        "h": 720,
        "l": "s3://bucket/7/8/foo/full/full/max/max/0/default.mp4",
        "n": "System preset: Generic 720p",
        "w": 1280,
        "ex": "mp4",
        "mt": "video/mp4"
    }
]
```

Moved `UpsertApplicationMetadata()` out of `IAssetApplicationMetadataRepository` and into an extension method specific for `Engine` as no other services need to write `AssetApplicationMetadata`. The repo was calling `.SaveChanges()` internally which felt wrong as it would persist any changes made to `Asset`. Approach was instead to read include `AssetApplicationMetadata` when reading `Asset` at the start of Engine request - all modifications to `AssetApplicationMetadata` can happen in the `Asset` alone, including identifying if create/update etc. EFCore change tracking can do the heavy lifting and all data written in transaction as part of single `SaveChanges()` at end of engine process. Has the added benefit of being less chatty to DB.

Refactored `ElasticTranscoderWrapper` - moved a few methods to `ElasticTranscoderPresetLookup` as a few places need lookups but most places don't need the rest of the ET methods.

Extended `MIMEHelper` to have _waay_ more recognised formats.

`AssetUpdateHandler` now removes `"AVTranscodes"` metadata if iiif-av channel removed.


src/protagonist/Engine/Data/AssetApplicationMetadataX.cs
```diff
@@ -0,0 +1,56 @@
...
+            return existingMetadata;
+        }
+
+        asset.AssetApplicationMetadata ??= new List<AssetApplicationMetadata>();
+        var assetApplicationMetadata = new AssetApplicationMetadata
+        {
+            MetadataType = metadataType,
+            MetadataValue = metadataValue,
+            Created = DateTime.UtcNow,
+            Modified = DateTime.UtcNow
```
JackLewis-digirati, line 44: is there a potential for getting microsecond differences when setting UtcNow twice?

src/protagonist/Engine/Ingest/Timebased/Completion/TimebasedIngestorCompletion.cs
```diff
@@ -47,47 +54,29 @@ public async Task<bool> CompleteSuccessfulIngest(AssetId assetId, int? batchId,
...
 
-    private List<Task<LargeObjectCopyResult>> CopyTranscodeOutputs(TranscodeResult transcodeResult,
+    private async Task<long> CopyTranscodeOutputs(TranscodeResult transcodeResult,
         List<string> errors, Asset asset, CancellationToken cancellationToken)
     {
         bool dimensionsUpdated = false;
         var transcodeOutputs = transcodeResult.Outputs;
-        var copyTasks = new List<Task<LargeObjectCopyResult>>(transcodeOutputs.Count);
+        var applicationMetadata = new List<AVTranscode>(transcodeOutputs.Count);
+        var transcodeSize = 0L;
```
JackLewis-digirati, line 76: comment explaining `0L`?

JackLewis-digirati (approved):

src/protagonist/Engine/Data/AssetApplicationMetadataX.cs
donaldgray: Updated to use `var`. Note that the above was refactored from elsewhere so there'll be items in DB where Created and Modified are slightly out. I don't think there'll be any issues in practice.

src/protagonist/Engine/Ingest/Timebased/Completion/TimebasedIngestorCompletion.cs
donaldgray: Renamed var to clarify use.

PR #978, opened 2025-03-06
Prevent duplicate iiif-av policy values

### LINKED ISSUE #975: `iiif-av` policies with duplicate transcode presets can be created
It is possible to create a `iiif-av` delivery channel policy that references the same transcode preset multiple times - e.g, `"policyData": "[\"video-webm-720p\", \"video-mp4-720p\",\"video-mp4-720p\"]"`

Attempting to ingest timebased assets using this configuration will result in Engine throwing the following error:
```
Ouput Key cannot contain duplicates: 03d4e2c7-7da6-4566-9d16-4e8c6da7bb1c/52/2341/fish-test-2/full/full/max/max/0/default.mp4
``` 

Protagonist should prevent `iiif-av` policy data containing duplicate keys from being set.

(Additionally, there is a typo in the error message - 'Ouput')

donaldgray:
Fixes #975


griffri (approved):
LGTM

PR #982, opened 2025-03-20
Allow thumbsClient to work with other image-server

donaldgray:
Specifically laya - have configurable basePath.
Rename class to remove mentions of "Cantaloupe"


griffri (approved):
LGTM

PR #984, opened 2025-03-28
Chore/645/dotnet8

p-kaczynski:


src/protagonist/Test.Helpers/Integration/DlcsDatabaseFixture.cs
```diff
@@ -208,7 +203,7 @@ public async Task InitializeAsync()
         }
         catch (Exception ex)
         {
-            var m = ex.Message;
+            _ = ex.Message;
```
JackLewis-digirati, line 206: can this just be removed?

JackLewis-digirati (approved):

src/protagonist/Test.Helpers/Integration/DlcsDatabaseFixture.cs
p-kaczynski: Fine with me, I focused on "not actually making direct changes", but this case definitely warrants it (I'm guessing it's just a leftover from someone running local debug)

src/protagonist/DLCS.Core.Tests/DLCS.Core.Tests.csproj
```diff
@@ -1,29 +1,28 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>net6.0</TargetFramework>
 
         <IsPackable>false</IsPackable>
 
         <LangVersion>10.0</LangVersion>
```
JackLewis-digirati, line 7: just working on another PR within protagonist, and ran into an issue with this where I tried to use `[]` for a List, and it fails due to being a C# 12 feature - should this be updated?

p-kaczynski: The `LangVersion` should prob be removed and left at the "default for current dotnet version", sounds good?

JackLewis-digirati: sounds good to me :thumbsup:

PR #991, opened 2025-04-07
Getting manifests to be persisted in the database + retrieval working

### LINKED ISSUE #952: Extend asset endpoints to expose "manifests"
The new `"manifests"` property needs to be added to Hydra Image class and mapped from DB entity and returned for any read/write endpoints. 

```json
{
  "@context": "https://api.dlcs.digirati.io/contexts/Image.jsonld",
  "@id": "https://api.dlcs/customers/2/spaces/1/images/first",
  "@type": "vocab:Image",
  "id": "first",
  "manifests": ["m123", "aadopsi0"]
}
```

This property should be rendered, if set, everywhere the Hydra image class can be returned:

* `GET|PATCH /customers/{c}/spaces/{s}/images`
* `POST /customers/{c}/allImages`
* `GET|PUT|PATCH /customers/{c}/spaces/{s}/images/{i}`
* `GET /customers/{c}/queue/batches/{b}/images`
* `GET /customers/{c}/queue/batches/{b}/assets`
* `POST /customers/{c}/queue`
* `POST /customers/{c}/queue/priority`
* _verify that the above is the comprehensive list of endpoints, add any that have been missed_

For updates, `"manifests"` can be:
* An array with values (e.g. `["foo", "bar"]`) - this will replace any currently saved values
* `null` - this will save `null` if new asset, or leave any existing `"manifests"` if update
* `[]` - this is explicitly clearing the values, so `null` in database.

JackLewis-digirati:
Resolves #952

This PR updates `PUT`/`POST`/`PATCH` to correctly save the `manifests` values, and `GET` to retrieve properly


src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
```diff
@@ -459,6 +460,7 @@ public async Task Get_BatchImages_200_IfImagesFound()
         var images = await response.ReadAsHydraResponseAsync<HydraCollection<DLCS.HydraModel.Image>>();
         images.TotalItems.Should().Be(3);
         images.Members.Should().HaveCount(3);
+        images.Members.All(x => x.Manifests.ContainsOnly("first")).Should().BeTrue();
```
donaldgray, line 463: I don't think existing tests should be updated to check if `manifests` field is returned, unless it's relevant to the test. `Get_BatchImages_200_IfImagesFound()` is a focussed test that the correct counts are returned. Sticking assertion for `"manifests"` seems a bit unusual, ie there are no other fields like `"tags"` or `"roles"` being checked.

src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
```diff
@@ -819,6 +825,7 @@ public async Task Post_CreateBatch_201_WithMixtureOfIdSet()
         Guid.TryParse(assetInDatabase.ToList()[0].Id.Asset, out _).Should().BeTrue();
         Guid.TryParse(assetInDatabase.ToList()[1].Id.Asset, out _).Should().BeTrue();
         assetInDatabase.ToList()[2].Id.Asset.Should().Be("someId");
+        assetInDatabase.ToList().All(a => a.Manifests.ContainsOnly("first")).Should().BeTrue();
```
donaldgray, line 828: As above `Post_CreateBatch_201_WithMixtureOfIdSet` is specifically testing that `"id"` can be empty, missing or set and it's resolved. Assertion on `manifests` doesn't fit the test.

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -108,6 +108,38 @@ public async Task Put_NewImageAsset_Creates_Asset()
         asset.BatchAssets.Should().BeEmpty();
     }
     
+    [Fact]
+    public async Task Put_NewImageAsset_CreatesAsset_WithManifestsAttached()
+    {
+        var customerAndSpace = await CreateCustomerAndSpace();
```
donaldgray, line 114: I think this is unnecessary. You can use the name of this method to ensure it's unique. `var assetId = AssetIdGenerator.GetAssetId()` should be enough?

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -881,6 +913,7 @@ public async Task Put_NewImageAsset_Creates_Asset_WithImageBytesProvided()
         var asset = await dbContext.Images.FindAsync(assetId);
         asset.Id.Should().Be(assetId);
         asset.MediaType.Should().Be(hydraJson.MediaType);
+        asset.Manifests.Should().BeEquivalentTo("first");
```
donaldgray, line 916: If asserting this should all fields be asserted?

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -1155,6 +1188,95 @@ public async Task Put_Existing_Asset_ClearsError_AndMarksAsIngesting()
         newAsset.Entity.Error.Should().BeEmpty();
     }
     
+    [Fact]
+    public async Task Put_ExistingAsset_UpdatesManifests()
+    {
+        var assetId = new AssetId(99, 1, nameof(Put_ExistingAsset_UpdatesManifests));
```
donaldgray, line 1194: `var assetId = AssetIdGenerator.GetAssetId();` (same with other new tests)

donaldgray (commented):
Overall LGTM but adding `"manifest"` assertions to some of the existing tests seems more for convenience to confirm the new field works. Is it enough to test the `"manifest"` update logic explicitly and use the hydra model conversion tests to confirm the field is mapped?

donaldgray (approved):

PR #995, opened 2025-04-10
Add ability to PATCH manifests using AllImages

### LINKED ISSUE #953: Support PATCH for `/allImages`
See [RFC-019](https://github.com/dlcs/protagonist/blob/main/docs/rfcs/019-presentation-dlcs.md#updating-resources) for more info.

Add PATCH support for `/customers/{c}/allImages`, accepting payloads in format:

```json
{
 "member": [ {"id": "1/2/a"},{"id": "1/2/b"} ],
 "field": "manifests",
 "operation": "add|remove|replace",
 "value": [ "m123" ]
}
```

* `"members"` is the list of image ids to apply PATCH to
* `"field"` specifies which field is being modified - `manifests` is the only valid value for now (400 if other field)
* `"operation"` can be one of (400 if other value):
  * 'add' - add the specified value(s) to _field_
  * 'remove' - remove the specified value(s) from _field_
  * 'replace' - replace the _field_ with specified value(s). `null` is rejected with 400 as need to be explicit, use `[]` to clear values.
* `"value"` is the value(s) to add/remove/replace

Returning HydraCollection of images. If no images found return an empty HydraCollection.

Writes should be atomic - competing requests that come in at the same time should both be recorded in the database. See [npgsql EF support for arrays](https://www.npgsql.org/efcore/mapping/array.html#operation-translation) and [RFC-019#implementation-details](https://github.com/dlcs/protagonist/blob/main/docs/rfcs/019-presentation-dlcs.md#implementation-detail)

Optional:
* If possible the implementation of this should be fairly dynamic to make it easier to extend operations and fields in the future

JackLewis-digirati:
Resolves #953

This PR adds an additional endpoint that allows manifests to be patched through `allImages`, in the below format:

```json
{
 "member": [ {"id": "1/2/a"},{"id": "1/2/b"} ],
 "field": "manifests",
 "operation": "add|remove|replace",
 "value": [ "m123" ]
}
```

This code has also been tests using parallel processing to make sure that multiple values will update the database, rather than replace existing values


src/protagonist/API/Features/Customer/Requests/AssetUpdater.cs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+public class AssetUpdater(DlcsContext dlcsContext) : IAssetUpdater
+{
+    public async Task<List<Asset>> UpdateAssets(UpdateAllImages request, CancellationToken cancellationToken = default)
+    {
+        var assetIds = ImageIdListValidation.ValidateRequest(request.HydraUpdate.Members!.Select(m => m.Id).ToList(),
+            request.CustomerId);
+
+        var assets = await dlcsContext.Images
```
donaldgray, line 30: From ticket:

> Writes should be atomic - competing requests that come in at the same time should both be recorded in the database. See [npgsql EF support for arrays](https://www.npgsql.org/efcore/mapping/array.html#operation-translation) and [RFC-019#implementation-details](https://github.com/dlcs/protagonist/blob/main/docs/rfcs/019-presentation-dlcs.md#implementation-detail)

By loading these upfront, competing requests could end up with differing results in DB as the final stored result is based on the state of data at load time. It'd be better to just issue UPDATE requests to DB,.

src/protagonist/Hydra/Collections/HydraUpdate.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Hydra.Model;
+using Newtonsoft.Json;
+
+namespace Hydra.Collections;
+
+public class HydraUpdate<T> : HydraCollection<T>
```
donaldgray, line 6: This is a for a specific type of `Update`, `BulkPatch` maybe? 

Does this need to inherit from `HydraCollection` - I _think_ all that this class and collection share is `Members[]`?

src/protagonist/Hydra/Model/OperationType.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Runtime.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Hydra.Model;
```
donaldgray, line 5: This is a DLCS specific thing, rather than a Hydra concern (AFAIK anyway, unless they have this concept too!)

src/protagonist/Hydra/Collections/HydraUpdate.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Hydra.Model;
+using Newtonsoft.Json;
+
+namespace Hydra.Collections;
```
donaldgray, line 4: Only generic `Hydra` classes (for https://www.hydra-cg.com/) should go in Hydra assembly

src/protagonist/API/Features/Customer/Requests/UpdateAllImages.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+using API.Infrastructure.Requests;
+using DLCS.Core;
+using DLCS.Model;
+using DLCS.Model.Assets;
+using Hydra.Collections;
```
donaldgray, line 6: Hydra concerns are limited to controller.  Controller receives hydra _stuff_ and converts before passing to mediatr/other classes. Mediatr returns entites/whatever and these are converted before being returned to consumer.

src/protagonist/API/Features/Customer/Requests/AssetUpdater.cs
```diff
@@ -0,0 +1,73 @@
...
+using Hydra.Model;
+using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json.Linq;
+
+namespace API.Features.Customer.Requests;
+
+public interface IAssetUpdater
+{
+    /// <summary>
+    /// Updates assets with new values from all image requests
```
donaldgray, line 18: Small thing, for name and comment: this is a set kind of update. Bulk path, or field patch rather than just `AssetUpdater`. 

re: comment - in future this might not come from "all image request", might be other endpoints that allow the bulk patching of some other asset (e.g. adding tags to a space or something).

src/protagonist/Hydra/Collections/HydraUpdate.cs
```diff
@@ -0,0 +1,16 @@
...
+public class HydraUpdate<T> : HydraCollection<T>
+{
+    [JsonProperty(Order = 30, PropertyName = "field")]
+    public required string Field { get; set; }
+    
+    [JsonProperty(Order = 31, PropertyName = "operation")]
+    public OperationType Operation { get; set; }
+    
+    [JsonProperty(Order = 32, PropertyName = "value")]
+    public required object Value { get; set; }
```
donaldgray, line 15: Can this be something more concrete for now? `string[]` or `T[]`? Can make it handle more complex types as/when need arises

src/protagonist/API.Tests/Integration/CustomerImageTests.cs
```diff
@@ -195,4 +198,199 @@ public async Task Post_DeleteImages_200_WithMatches()
...
+    [InlineData("first", "second", "remove", "first")]
+    public async Task Patch_AllImages_TestManifestPermutations(string initial, string update, string operation, params string[] result)
+    {
+        // Arrange
+        var assetid = AssetIdGenerator.GetAssetId();
+        var asset = await dbContext.Images.AddTestAsset(assetid, manifests: initial?.Split(',').ToList());
+        await dbContext.SaveChangesAsync();
+
+        var patchAllImages = $@"{{
+  ""@type"": ""Collection"",
```
donaldgray, line 217: I don't think this is a collection, it's our own custom patching format

donaldgray (changes requested):
Overall looks functional, main change required is loading the assets upfront and working out what manifests should be in memory, then saving.

src/protagonist/Hydra/Collections/HydraUpdate.cs
JackLewis-digirati: They also share the validators for the `Members[]`

donaldgray: Sharing the validator makes sense, but `HydraUpdate : HydraCollection` means _"HydraUpdate is-a HydraCollection"_, which isn't true? Can the validator be refactored?

JackLewis-digirati: ah, sure - should be refactored now to take in a shared interface that is used for the validator instead

src/protagonist/API/Features/Customer/CustomerImagesController.cs
```diff
@@ -67,6 +70,58 @@ public async Task<IActionResult> GetAllImages(
...
+        CancellationToken cancellationToken = default)
+    {
+        var validationResult = await validator.ValidateAsync(bulkPatch, cancellationToken);
+        if (!validationResult.IsValid)
+        {
+            return this.ValidationFailed(validationResult);
+        }
+
+        return await HandleUpsert(new UpdateAllImages(bulkPatch.Members!.Select(m => m.Id).ToList(), customerId, bulkPatch.Value, bulkPatch.Field, 
+                (OperationType)bulkPatch.Operation), al =>
```
donaldgray, line 112: `(OperationType)` is unnecessary as it's same type as property

donaldgray (approved):

PR #1000, opened 2025-04-17
Fix issue where remove doesn't work for multiple items

JackLewis-digirati:


griffri (approved):
LGTM

src/protagonist/API/Features/Customer/BulkAssetPatcher.cs
```diff
@@ -70,7 +65,28 @@ await dlcsContext.Images
                 throw new InvalidOperationException($"Unsupported operation '{operation}'");
         }
     }
-    
+
+    // allows a query to be generated using a list, while avoiding issues with SQL injection
+    private async Task RemoveManifests(List<AssetId> assetIds, List<string>? value, int customerId, CancellationToken cancellationToken)
```
donaldgray, line 70: Would this method be more efficient if it uses batches and positional placeholders? Not sure what effort would be for doing so?

Positional placeholders are the recommended way as it doesn't involve any rewriting (see https://www.npgsql.org/doc/basic-usage.html#positional-and-named-placeholders).

Batches would mean that 1 connection sends multiple requests, rather than multiple connections sending single requests (https://www.npgsql.org/doc/basic-usage.html#batching)

PR #1005, opened 2025-05-07
Correctly set image service profiles on manifests

donaldgray:
thumbnails are level0, images level2

resolves issue introduced by refactoring `ManifestBuilderUtils` in #1004


griffri (approved):
LGTM

PR #1006, opened 2025-05-12
Validate bulk asset patch member property

### LINKED ISSUE #1002: Receive 500 if `/allImages` PATCH has `null` "member" property
Minimum payload to see error:

```json
{
    "field": "manifests",
    "operation": "remove",
    "value": [
        "m123"
    ]
}
```

initially noticed as I misspelled `"member"` as `"members"`.

This should be caught and a 400 returned. I suspect it's related to refactoring of `ImageIdListValidator` to inherit from `: AbstractValidator<IdentifierOnly[]?>`. The validator is likely not invoked as it can't validate null. 

This will need to happen elsewhere, or validator modified, or payload updated to default `Members` property to be an empty list.

Note that this also happens when POST to `/allImages`, any other usages of this will need to be checked (deleteImages?)

JackLewis-digirati:
Resolves #1002 

This PR fixes an issue where an obtuse error with a stack trace is returned to the user when the `member` property is `null` - this now correctly responds with an error response and a status code of `400`

### DIFF
--- a/src/protagonist/API.Tests/Features/Customer/Validation/BulkAssetPatchValidatorTests.cs
+++ b/src/protagonist/API.Tests/Features/Customer/Validation/BulkAssetPatchValidatorTests.cs
@@ -1,0 +1,47 @@
+﻿using API.Features.Image.Validation;
+using API.Infrastructure.Requests;
+using API.Settings;
+using DLCS.Model;
+using FluentValidation.TestHelper;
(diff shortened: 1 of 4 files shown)


donaldgray (commented):
Does this need applied to other uses of `AbstractValidator<IdentifierOnly[]>`?

donaldgray (approved):

PR #1036, opened 2025-08-19
Small tweaks

donaldgray:
Minor tweaks noticed when investigating code while writing #1035 

* Tweak `AssetUpdatedHandler` 
  * Only call `GetMatchingKeys()` once and fail if unable to get preset details from Engine (alternative deletes all keys, which isn't desirable).
  * Syntax tweaks
* Remove unused method from `S3StorageKeyGenerator`
* Helper for generating `DeliveryChannelsBeforeProcessing`, keep same model of mapping hydra -> internal model and removes duplicate  code


donaldgray:
Rebased and changed base branch. This originally came from same branch as #1035 but I don't want to hold this one up until that is reviewed so will merge direct to `develop`.

PR #777, opened 2024-03-26
Get image and thumbnail channels to work in engine

JackLewis-digirati:
Resolves #629
Links WDC-37

This PR adds the ability for the image and thumbs delivery channel to be correctly processed by engine and viewed using orchestrator.  It also splits the image processor for thumbs out of appetiser to use special server to generate thumbnails 

This PR additionally bumps ImageSharp to 3.1.3, which resolves #754

PR #1046, opened 2025-09-25
Replace EngineThumbs with updated Appetiser

### LINKED ISSUE #1044: Engine should use Appetiser for thumbs
As per https://github.com/dlcs/protagonist/blob/develop/docs/adr/0009-engine-appetiser-thumbs.md, update Engine to use [Appetiser](https://github.com/dlcs/appetiser) for both JP2 and thumbnail generation.

This was the behaviour until https://github.com/dlcs/protagonist/pull/777 refactored to use Cantaloupe instance for thumbnail generation and Appetiser for JP2 generation.

We cannot simply revert to the previous implementation as that only supported single bounding box thumbs (e.g. `!1024,1024`) whereas we now support a variety of valid IIIF ImageApi size parameters for thumbnails. Part of this was due to Appetiser only supporting single "longest edge" sizes but Appetiser will now support Image API size parameters.

Engine can either:
* Provide size-parameters as specified in the delivery channel policy. The logic to confirm that output is of the correct longest-edge dimensions (https://github.com/dlcs/protagonist/blob/develop/src/protagonist/Engine/Ingest/Image/ImageServer/Clients/ImageServerThumbsClient.cs#L177) can be called to ensure what we store in s.json is correct.
* Provide calculated sizes as IIIF size parameters (e.g. rather than `"!1024,1024"` pass `"1024,596"`) as this will ensure the file is of the exact size and we can remove the "check" step afterwards.

The latter should be more efficient overall as it would avoid Engine needing to measure the file - it knows the exact size asked for. The best approach can come with testing.

donaldgray:
Resolves #1044 - remove use of ImageServer for thumbs generation and use Appetiser for thumbs and JP2 generation.

We switched to using an ImageServer when doing delivery-channel work, see #777. This PR reverts back to what we had before but it couldn't be identical due to how other elements of codebase have moved on.

Refactored `AppetiserClient` - this is larger now due to changes in Appetiser code and handling different response bodies. There is logic for dealing with appetiser that ensures paths are unix while local paths may not be - this bled out of this class so refactored to keep all of this logic internal to `AppetiserClient` - basically taking the response model and rewriting some paths on the way out, so that when it's returned to caller all of the paths are in the correct format.

`ImageServerClient` is now `AppetiserImageProcessor`, logic here was mostly the same but now handles generated thumbnails and has logic to work out what kind of call we want to make to Appetiser (ie generate thumbs only, or thumbs and derivative).

Increased logging in `ThumbCreator` and removed blocking if thumbs are larger than incoming image. Allowed as we support `^` when creating thumbnails.

> [!WARNING]
> `ImageServerThumbsClient` contained `ValidateSize()` to verify that that sizes we got back were what we expected but this hasn't been ported over as Appetiser has been reliable in this regard. We can add it back in if required.

Update docker-compose for new image 

Not directly related to ticket:
* Fixed bug with `TopicPublisher.PublishToAssetModifiedTopic` - reduced size of items in batch from 10 -> 5. 10 is max number but this could trip the overall body size limit, decided to half that amount rather than attempt to be clever and calculate the overall size (although we can do in future if required).
* Simplify `IngestExecutor` logic - remove invalid status check and avoid doing potentially unnecessary work.


src/protagonist/Engine.Tests/Ingest/Image/ImageServer/AppetiserImageProcessorTests.cs
```diff
@@ -0,0 +1,497 @@
...
+    {
+        // Arrange
+        var context = IngestionContextFactory.GetIngestionContext(AssetIdGenerator.GetAssetId(asset: "some(id)"));
+        var expectedDirectory = $"scratch/{context.IngestId}/99/1/some_id_/output";
+        
+        // Act
+        await sut.ProcessImage(context);
+
+        // Assert
+        A.CallTo(() => fileSystem.CreateDirectory(
```
JackLewis-digirati, line 88: could maybe do with an explanation of why once, then twice (i.e.: what it's creating/deleting etc.)

src/protagonist/Engine.Tests/Ingest/IngestExecutorTests.cs
```diff
@@ -209,22 +209,27 @@ public async Task IngestAsset_FirstWorkerFail_DoesNotCallFurtherWorkers(IngestRe
...
-            }
+            ]
         };
         
         // Act
         var result = await sut.IngestAsset(asset, customerOriginStrategy);
 
         // Assert
         result.Status.Should().Be(IngestResultStatus.Success);
+        A.CallTo(() => repo.UpdateIngestedAsset(asset, null,
```
JackLewis-digirati, line 229: maybe a comment? - took a second to understand why the update was happening for a `None` channel

src/protagonist/Engine/Ingest/Image/ImageServer/AppetiserImageProcessor.cs
```diff
@@ -11,70 +11,51 @@
...
-                context.Asset.Error = $"Appetiser Error: {failResponse.Message}";
-                return false;
+                case AppetiserResponseModel successResponse:
+                    await ProcessResponse(context, successResponse, flags);
+                    return true;
+                case AppetiserResponseErrorModel failResponse:
+                    context.Asset.Error = $"Appetiser Error: {failResponse.Message}";
+                    return false;
+                default:
+                    return false;
```
JackLewis-digirati, line 57: this can't really happen... but is it worth setting `context.Asset.Error` here?

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/AppetiserClient.cs
```diff
@@ -1,100 +1,275 @@
...
+    /// 
+    /// This logic allows handling when running locally on win/unix and when deployed to unix
+    /// </summary>
+    /// <param name="assetId">Modified assetId</param>
+    /// <param name="ingestId">Ingest identifier, for avoiding overwriting files</param>
+    /// <param name="forAppetiser">True if the generated path will be used by Appetiser</param>
+    /// <param name="templateType">Type of template to use</param>
+    /// <param name="extensionForFile">Leave null if want directory only. If supplied then result will be
+    /// "{destFolder}{assetId.Asset}.{extensionForFile}"</param>
+    /// <returns></returns>
```
JackLewis-digirati, line 152: empty comment

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/IImageProcessorClient.cs
```diff
@@ -0,0 +1,47 @@
...
+    
+    /// <summary>
+    /// Image processor should generate thumbnails
+    /// </summary>
+    Thumbnails = 2,
+    
+    /// <summary>
+    /// Convert incoming file to JP2 derivative
+    /// </summary>
+    Derivative = 4,
```
JackLewis-digirati, line 46: this jumps 3  is there a reason for that?

src/protagonist/Test.Helpers/Http/ControllableHttpMessageHandler.cs
```diff
@@ -12,14 +14,30 @@ namespace Test.Helpers.Http;
 /// </summary>
 public class ControllableHttpMessageHandler : HttpMessageHandler
 {
+    private static readonly JsonSerializerOptions Settings = new(JsonSerializerDefaults.Web);
     private HttpResponseMessage response;
-    public List<string> CallsMade { get; }= new List<string>();
+    public List<string> CallsMade { get; }= new();
```
JackLewis-digirati, line 19: I know this is on the old version and is tiny, but missing space before `=`

src/protagonist/Engine/Ingest/Image/ImageServer/Clients/IImageProcessorClient.cs
donaldgray: Yeah, it's a `[Flags]` enum so required for bitwise operations in `AppetiserClient`

src/protagonist/Engine/Ingest/Image/ImageServer/AppetiserImageProcessor.cs
donaldgray: Added error containing received type - could happen if we add a further impl of `IImageProcessorResponse`

PR #1082, opened 2025-12-15
Add initial external assets adjuncts work

### LINKED ISSUE #1073: Management of externally hosted adjuncts
Basic implementation of API endpoints to support management of externally hosted adjuncts.

Add the following endpoints:

* GET `/customers/{c}/spaces/{s}/images/{i}/adjuncts` - a hydra collection of all adjuncts for Asset, ordered by id. Returns 404 if asset doesn't exist, else 200 (collection may be empty if no adjuncts)
* GET `/customers/{c}/spaces/{s}/images/{i}/adjuncts/{adjunct-id}` - returns a single adjunct. 404 if asset or adjunt doesn't exist, else 200.
* POST `/customers/{c}/spaces/{s}/images/{i}/adjuncts` - returns a 201 w/ Location header to newly created adjunct. 404 if asset doesn't exist.
* PUT `/customers/{c}/spaces/{s}/images/{i}/adjuncts/{adjunct-id}` - returns a 201 w/ Location header to newly created adjunct if create, else returns 200 if update. 404 if asset doesn't exist.
* DELETE `/customers/{c}/spaces/{s}/images/{i}/adjuncts` - return 204 if delete successful. 404 if not found.

Supported fields for PUT and POST are as per documentation, summarised below:
* `@id` - Optional. The API URL of the JSON resource. This is always of the form `<asset-@id>/adjuncts/<adjunct-id>`. It is not usually supplied by the API caller on a PUT or a POST. If it is supplied it must not conflict with any of the other fields that contribute to the URI.
* `id` - The model id of the adjunct. Required on a POST when creating an adjunct, to provide a "filename" or last path element for the adjunct. The platform will not mint a value if one is not supplied (unlike assets). On a POST, it must not conflict with any of the asset's other adjuncts. Optional on a PUT; if present, it must match the last element in the PUT URL path.
* `@type`: Required. It has the same meaning as in the IIIF Presentation Specification (see [type](https://iiif.io/api/presentation/3.0/#type)) and should use type values defined by that specification, like Dataset, Text or AnnotationPage. It will appear in any generated IIIF.
* `mediaType`: Required. The internet content type (or MIME type) of the resource.
* `profile`: Optional. Same as the IIIF Presentation API [profile](https://iiif.io/api/presentation/3.0/#profile) property.
* `label`: Optional. This is the same as the IIIF Presentation API [label](https://iiif.io/api/presentation/3.0/#label) property, and is a JSON LD Language Map
* `language`: Optional. The language of an adjunct's content, if applicable. The value is a JSON array of ISO language codes (sub-codes are allowed). No validation other than field length.
* `externalId` - Required (for mvp). `externalId` is a fully-qualified URL external to the platform. We don't validate whether the resource exists
* `size` - Optional. The size in bytes of the adjunct. For adjuncts created with `externalId` you can supply this yourself. 
* `iiifLink` - Required. This property is used when the adjunct is expressed by the platform in the IIIF Presentation API. Valid values are `seeAlso`, `annotations` and `rendering`
  * `seeAlso`, `annotations` and `rendering` are currently the only accepted values and must match this case (ie `SeeAlso` or `Annotations` is rejected).

In addition to the above fields, the following will also be returned in response payloads
* `publicId` - The URL from which the adjunct's binary content is served. This will match the `externalId` (extended later)
* `created` - The timestamp record was created
* `finished` - The timestamp record was finished processing (will be identical for external-id as no processing required)

Images hydra model gains `adjuncts` property linking to `/customers/{c}/spaces/{s}/images/{i}/adjuncts`.

> [!NOTE]
> PUT requests are always for the full model, there are no partial updates.

JackLewis-digirati:
Resolves #1073

Implements the `/adjuncts` endpoint for external assets.

This is a fairly standard `CRUD` controller that sits under `asset` and currently can only be seen with the `API`.

Additionally, `adjuncts` has been added as a link in an `image`, following details from here - https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/registering-assets

> [!NOTE]
> Also updates appetiser in the local docker compose file

> [!NOTE]
> Contains a migration


src/protagonist/DLCS.Repository/Migrations/DlcsContextModelSnapshot.cs
```diff
@@ -30,6 +30,53 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                    b.Property<DateTime>("Created")
+                        .HasColumnType("timestamp with time zone");
+
+                    b.Property<string>("ExternalId")
+                        .HasColumnType("text");
+
+                    b.Property<string>("IiifLink")
+                        .IsRequired()
+                        .HasMaxLength(500)
+                        .HasColumnType("character varying(500)");
```
stephenwf: Because it would require a DB change later, I think it's best to mention now even if it is not implemented at the moment. I think this should be optional. I think there would be a benefit in being able to add an intermediate step that _doesn't_ appear in the final IIIF - or is supplemental to it.

donaldgray: That functionality will be covered later. Currently IIIFLink can be `seeAlso`, `annotations` or `rendering`.

In a future ticket we'll handle IIIFLink being empty - in that case it will be output on any generated Manifests in an `"otherAdjuncts"` property (see https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/adjuncts#iiiflink). The idea is that we are advertising it but not in a proper IIIF compliant manner. 

When NQ is generated, the consumer can do with it what they will - ignore it, move it etc. They important thing is they'll know it exists and will have paths to request it etc.

src/protagonist/API/Features/Adjuncts/AdjunctsController.cs
```diff
@@ -0,0 +1,131 @@
...
+            return this.HydraProblem($"The adjunct id from the request URI does not match the 'id' from the request body",
+                null, 400);
+        }
+        
+        hydraAdjunct.ModelId = adjunctId;
+        
+        return await CreateOrUpdateAdjunct(customerId, spaceId, imageId, hydraAdjunct, validator, false, cancellationToken);
+    }
+    
+    private async Task<IActionResult> CreateOrUpdateAdjunct(int customerId, int spaceId, string imageId, Adjunct hydraAdjunct,
```
donaldgray, line 113: Small, consistency - other classes in Protagonist have `public` methods first/higher in class, with `private`/`protected` etc lower. 

This can be particularly useful for Controllers as these are the API endpoints

src/protagonist/API/Features/Adjuncts/AdjunctsController.cs
```diff
@@ -0,0 +1,131 @@
...
+            createOrUpdateRequest.Adjunct.Id,
+            "Create or update adjunct failed", cancellationToken);
+    }
+
+    /// <summary>
+    /// Delete an adjunct for an asset.
+    /// </summary>
+    /// <returns>A delete result.</returns>
+    [HttpDelete("{adjunctId}")]
+    [ProducesResponseType(200, Type = typeof(HydraCollection<Adjunct>))]
```
donaldgray: Will return 204|No Content rather than 200/list of Adjuncts

src/protagonist/API/Features/Adjuncts/AdjunctsController.cs
```diff
@@ -0,0 +1,131 @@
...
+[ApiController]
+public class AdjunctsController(
+    IOptions<ApiSettings> options,
+    IMediator mediator)
+    : HydraController(options.Value, mediator)
+{
+    /// <summary>
+    /// Get details of all adjuncts for an asset.
+    /// </summary>
+    /// <returns>A Hydra JSON-LD Adjunct object representing the adjuncts.</returns>
```
donaldgray: Small - returns a Hydra Collection, rather than an Adjunct (same with PUT/POST)

src/protagonist/API/Features/Adjuncts/Requests/GetAdjuncts.cs
```diff
@@ -0,0 +1,29 @@
...
+using DLCS.Repository;
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+
+namespace API.Features.Adjuncts.Requests;
+
+public class GetAdjuncts(AssetId assetId) : IRequest<FetchEntityResult<IReadOnlyCollection<Adjunct>>>, IPagedRequest
+{
+    public int Page { get; set; }
+    public int PageSize { get; set; }
```
donaldgray: This no longer needs to be an `IPagedRequest`

src/protagonist/API/Features/Adjuncts/Requests/GetAdjuncts.cs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+public class GetAdjunctsHandler(DlcsContext dbContext) : IRequestHandler<GetAdjuncts, FetchEntityResult<IReadOnlyCollection<Adjunct>>>
+{
+    public async Task<FetchEntityResult<IReadOnlyCollection<Adjunct>>> Handle(GetAdjuncts request, CancellationToken cancellationToken)
+    {
+        var adjuncts = await dbContext.Adjuncts.AsNoTracking().Where(a => a.AssetId == request.AssetId)
+            .ToListAsync(cancellationToken);
+        
+        return FetchEntityResult<IReadOnlyCollection<Adjunct>>.Success(adjuncts);
```
donaldgray: From ACs

> Returns 404 if asset doesn't exist, else 200 (collection may be empty if no adjuncts)

I don't think this will happen here? It looks like a 200 will be returned whether the asset exists or not?

src/protagonist/API/Features/Adjuncts/Requests/GetAdjuncts.cs
```diff
@@ -0,0 +1,29 @@
...
+using DLCS.Core.Types;
+using DLCS.Model.Assets;
+using DLCS.Model.Page;
+using DLCS.Repository;
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+
+namespace API.Features.Adjuncts.Requests;
+
+public class GetAdjuncts(AssetId assetId) : IRequest<FetchEntityResult<IReadOnlyCollection<Adjunct>>>, IPagedRequest
```
donaldgray: Small/consistency - should this be `GetAllAdjuncts`?

Other similar handlers use `GetAll{Thing}s` and `Get{Thing}` but this uses `Get{Thing}s` and `Get{Thing}`. E.g. `GetAllNamedQueries` and `GetNamedQuery`, `GetAllCustomers` and `GetCustomer`.

src/protagonist/API/Features/Adjuncts/Requests/DeleteAdjunct.cs
```diff
@@ -0,0 +1,36 @@
...
+public class DeleteAdjunctHandler(DlcsContext dbContext)
+    : IRequestHandler<DeleteAdjunct, ResultMessage<DeleteResult>>
+{
+    public async Task<ResultMessage<DeleteResult>> Handle(DeleteAdjunct request, CancellationToken cancellationToken)
+    {
+        var deletion = await dbContext
+            .Adjuncts
+            .Where(a => a.Id == request.Id && a.AssetId == request.AssetId)
+            .ExecuteDeleteAsync(cancellationToken);
+        await dbContext.SaveChangesAsync(cancellationToken);
```
donaldgray: `SaveChangesAsync()` is erroneous when using `ExecuteDeleteAsync()`, one of the benefits of `ExecuteDelete` is that there's no change tracking involved. 

See https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete

> [ExecuteUpdate](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.relationalqueryableextensions.executeupdate) and [ExecuteDelete](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.relationalqueryableextensions.executedelete) are a way to save data to the database without using EF's traditional change tracking and [SaveChanges()](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.savechanges#microsoft-entityframeworkcore-dbcontext-savechanges) method.

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    /// <summary>
+    /// The adjunct to create/update
+    /// </summary>
+    public Adjunct Adjunct { get; set; } = adjunct;
+    
+    /// <summary>
+    /// Whether only creation is allowed (no update)
+    /// </summary>
+    public bool CreateOnly { get; set; } = createOnly;
```
donaldgray: These props can be `{ get; }` only

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -0,0 +1,73 @@
...
+        var adjunct = request.Adjunct;
+        
+        Adjunct? dbAdjunct = null;
+        if (!request.CreateOnly)
+        {
+            dbAdjunct = await dbContext.Adjuncts.SingleOrDefaultAsync(a =>
+                a.Id == adjunct.Id && a.AssetId == adjunct.AssetId, cancellationToken);
+        }
+
+        if (dbAdjunct != null)
```
donaldgray, line 38: `Id` and `AssetId` can't change - is it worth omitting from this list?

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -0,0 +1,73 @@
...
+            dbAdjunct.Profile = adjunct.Profile;
+            dbAdjunct.Label = adjunct.Label;
+            dbAdjunct.Language = adjunct.Language;
+            dbAdjunct.ExternalId = adjunct.ExternalId;
+            dbAdjunct.Finished = DateTime.UtcNow;
+            dbAdjunct.Size = adjunct.Size;
+        }
+        else
+        {
+            request.Adjunct.Created = DateTime.UtcNow;
```
donaldgray: For clarity, should we just use `adjunct` here? (because `var adjunct = request.Adjunct;`). Same with the return object.

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        
+        try
+        {
+            await dbContext.SaveChangesAsync(cancellationToken);
+        }
+        catch (DbUpdateException ex) when (ex.GetDatabaseError() is UniqueConstraintError)
+        {
+            return ModifyEntityResult<Adjunct>.Failure(
+                $"An adjunct called '{request.Adjunct.Id}' already exists",
```
donaldgray: `with id` rather than `called` ?

src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+            .WithMessage("'iiifLink' is required");
+        RuleFor(a => a.MediaType).NotEmpty()
+            .WithMessage("'mediaType' is required");
+        
+        RuleFor(a => a.ExternalId)
+            .NotEmpty()
+            .Must(a => Uri.IsWellFormedUriString(a, UriKind.Absolute))
+            .WithMessage("'externalId' is required and must be a well formed URI");
+        
+        RuleFor(a => a.IIIFLink).Must(a => Enum.IsDefined(typeof(IIIFLinkType), a!))
```
donaldgray: Small - is it worth moving this beside line 11 to keep validators for same fields together? When initially scanning code I thought we were only validating that `iiifLink` had a value.

src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+            .When(a => a.Id != null && a.ModelId != null)
+            .WithMessage("'id' and '@id' must have a matching adjunct identifier");
+        
+        RuleFor(a => a.ModelId)
+            .NotEmpty()
+            .WithMessage("Adjunct identifier could not be found");
+        
+        RuleFor(a => a.Type)
+            .NotEmpty()
+            .WithMessage($"'@type' is required");
```
donaldgray: `$` can go

src/protagonist/DLCS.HydraModel/Adjunct.cs
```diff
@@ -0,0 +1,79 @@
...
+        Range = Names.XmlSchema.String, ReadOnly = false, WriteOnly = false)]
+    [JsonProperty(Order = 16, PropertyName = "language")]
+    public string[]? Language { get; set; }
+    
+    [RdfProperty(Description = "A fully-qualified URL external to the platform.",
+        Range = Names.XmlSchema.String, ReadOnly = false, WriteOnly = false)]
+    [JsonProperty(Order = 17, PropertyName = "externalId")]
+    public required string ExternalId { get; set; }
+    
+    [RdfProperty(Description = "Where this adjunct is hosted publically.",
```
donaldgray: `publically` -> `publicly`

src/protagonist/DLCS.Model/Assets/Adjunct.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public LanguageMap? Label { get; set; }
+    
+    /// <summary>
+    /// The language(s) of the content
+    /// </summary>
+    public string[]? Language { get; set; }
+    
+    /// <summary>
+    /// A fully-qualified URL external to the platform
```
donaldgray: worth adding something like "where the adjunct is hosted" or "where the adjunct can be accessed"?

src/protagonist/DLCS.Model/Assets/Adjunct.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public required string MediaType { get; set; }
+    
+    /// <summary>
+    /// How this adjunct is expressed in IIIF presentation
+    /// </summary>
+    public required IIIFLinkType IIIFLink { get; set; }
+    
+    /// <summary>
+    /// The asset this adjunct is associated with
```
donaldgray: Splitting hairs but this is the Id of the Asset it's associated with, `public Asset Asset` is the actual asset the adjunct is associated with.

src/protagonist/DLCS.Repository/DlcsContext.cs
```diff
@@ -697,6 +704,29 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
             entity.HasKey(ba => new { ba.BatchId, ba.AssetId, });
         });
+        
+        modelBuilder.Entity<Adjunct>(entity =>
```
donaldgray, line 708: Where did these MaxLength values come from? Do we need them, if so should they differ? e.g. the longest `IIIFLink` value is `annotations`, which is 11 chars.

src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+            .WithMessage("'mediaType' is required");
+        
+        RuleFor(a => a.ExternalId)
+            .NotEmpty()
+            .Must(a => Uri.IsWellFormedUriString(a, UriKind.Absolute))
+            .WithMessage("'externalId' is required and must be a well formed URI");
+        
+        RuleFor(a => a.IIIFLink).Must(a => Enum.IsDefined(typeof(IIIFLinkType), a!))
+            .When(a => a.IIIFLink != null)
+            .WithMessage("Valid values for 'iiifLink' are 'SeeAlso', 'Annotations' and 'Rendering'");
```
donaldgray: Valid values are camel-case (`seeAlso`, `annotations` and `rendering`). 

We should enforce that only the correctly case is used be explicit (ie don't accept `SeeAlso`, it must be `seeAlso`).

src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+        var enumParsed = Enum.TryParse(hydraAdjunct.IIIFLink, out IIIFLinkType iiifLink);
+
+        if (!enumParsed)
+        {
+            throw new APIException("Hydra adjunct does iiifLink could not be parsed");
+        }
+        
+        Debug.Assert(adjunctId != null, "adjunctId != null");
+
+        LanguageMap? label = null;
```
donaldgray: If `ToLanguageMap` was tolerant of nulls this block lines 45-49 could be removed

src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+using DLCS.Model.Assets;
+using Adjunct = DLCS.Model.Assets.Adjunct;
+using LanguageMap = IIIF.Presentation.V3.Strings.LanguageMap;
+
+namespace API.Converters;
+
+public static class AdjunctConverter
+{
+    /// <summary>
+    /// This will create a DLCS.Model.Assets.Asset with the correct Id, Customer and Space.
```
donaldgray: XML comment from `Asset`

src/protagonist/API.Tests/Features/Adjuncts/Validation/HydraAdjunctValidatorTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            IIIFLink = "SeeAlso",
+            Type = "Image"
+        };
+        var result = sut.TestValidate(adjunct);
+        result.ShouldHaveValidationErrorFor(r => r.ExternalId)
+            .WithErrorMessage("'externalId' is required and must be a well formed URI");
+    }
+    
+    [Fact]
+    public void ExternalId_Null()
```
donaldgray: Should this test null or empty?

src/protagonist/API.Tests/Features/Adjuncts/Validation/HydraAdjunctValidatorTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            IIIFLink = "SeeAlso",
+            Type = "Image"
+        };
+        var result = sut.TestValidate(adjunct);
+        result.ShouldHaveValidationErrorFor(r => r.ExternalId)
+            .WithErrorMessage("'externalId' is required and must be a well formed URI");
+    }
+    
+    [Fact]
+    public void IiifLink_NotValid()
```
donaldgray: (tiny) `IIIFLink`, rather than `IiifLink` as `IIIF` is an acronym

src/protagonist/API.Tests/Integration/AdjunctTests.cs
```diff
@@ -0,0 +1,424 @@
...
+namespace API.Tests.Integration;
+
+[Trait("Category", "Integration")]
+[Collection(StorageCollection.CollectionName)]
+public class AdjunctTests : IClassFixture<ProtagonistAppFactory<Startup>>
+{
+    private readonly HttpClient httpClient;
+    private readonly DlcsContext dbContext;
+    
+    public AdjunctTests(StorageFixture storageFixture, ProtagonistAppFactory<Startup> factory)
```
donaldgray: This doesn't need `StorageFixture`, can just use `DlcsDatabaseFixture`. The former contains a `DlcsDatabaseFixture` but also a `LocalStackFixture` but that isn't required for these operations (it should marginally speed up tests/reduce overhead as LocalStack container won't need to run).

src/protagonist/API.Tests/Integration/AdjunctTests.cs
```diff
@@ -0,0 +1,424 @@
...
+
+        // Act
+        var response = await httpClient.AsCustomer(assetId.Customer).GetAsync(path);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.NotFound);
+    }
+    
+    [Fact]
+    public async Task GetAdjunctS_RetrievesListOfAdjuncts()
```
donaldgray: Capital `S` here (same with others that test bulk fetch).

src/protagonist/API.Tests/Integration/AdjunctTests.cs
```diff
@@ -0,0 +1,424 @@
...
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        var adjunct = await response.ReadAsHydraResponseAsync<HydraCollection<Adjunct>>();
+
+        adjunct.Members.Length.Should().Be(0);
+    }
+    
+    [Fact]
+    public async Task GetAdjunctS_RetrievesEmptyListOfAdjuncts_WhenNoAsset()
```
donaldgray: This should be a 404 rather than an empty list. From ACs

> Returns 404 if asset doesn't exist, else 200 (collection may be empty if no adjuncts)

src/protagonist/API.Tests/Integration/AdjunctTests.cs
```diff
@@ -0,0 +1,424 @@
...
+        adjunct.ExternalId.Should().Be("https://some-location.com/an-adjunct");
+        adjunct.PublicId.Should().Be("https://some-location.com/an-adjunct");
+        
+        response.Headers.Location.Should()
+            .Be(
+                $"http://localhost/customers/{assetId.Customer}/spaces/{assetId.Space}/images/{assetId.Asset}/adjuncts/someAdjunctId");
+    }
+    
+    [Fact]
+    public async Task DeleteAdjunct_NothingHappens_WhenDoesNotExist()
```
donaldgray: Small - is `DeleteAdjunct_Returns404` better than `DeleteAdjunct_NothingHappens`?

donaldgray (changes requested):
Overall LGTM but a couple of little tweaks required - most comments are small naming etc.

src/protagonist/DLCS.Repository/DlcsContext.cs
JackLewis-digirati: it was based on other similar fields in the `context` class to not allow absolutely gigantic id's etc. - can likely be removed, but I felt they were useful to have as a basic stopgap, though the `iiifLink` property values are not needed - will remove that one

src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <param name="adjunct">The DLCS adjunct to convert</param>
+    /// <param name="urlRoots">The base address used to create FQDN paths</param>
+    /// <returns>A hydra model representation of the adjunct</returns>
+    public static DLCS.HydraModel.Adjunct ToHydra(this Adjunct adjunct, UrlRoots urlRoots)
+    {
+        return new DLCS.HydraModel.Adjunct(urlRoots.BaseUrl, adjunct.AssetId.Customer, adjunct.AssetId.Space, adjunct.AssetId.Asset, adjunct.Id)
+        {
+            Type = adjunct.Type,
+            MediaType = adjunct.MediaType,
+            IIIFLink = adjunct.IIIFLink.ToString().ToCamelCase(true),
```
donaldgray: Rather than `.ToCamelCase()` here, is it worth using `DescriptionAttribute` to explicitly specify the value we want the enum represented as? There are helpers in `EnumX` for reading

src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+        var enumParsed = Enum.TryParse(hydraAdjunct.IIIFLink, true, out IIIFLinkType iiifLink);
+
+        if (!enumParsed)
+        {
+            throw new APIException("Hydra adjunct does iiifLink could not be parsed");
+        }
+        
+        Debug.Assert(adjunctId != null, "adjunctId != null");
+
+        LanguageMap? label = null;
```
donaldgray: Why split declaration from assignment here? Also, don't think this var is needed at all?

src/protagonist/API/Converters/LanguageMapX.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections.Generic;
+using IIIF.Presentation.V3.Strings;
+
+namespace API.Converters;
+
+public static class LanguageMapX
+{
+    public static LanguageMap? ToLanguageMap(this Dictionary<string, List<string>>? dictionaryToConvert)
+    {
```
donaldgray, line 9: Feels a lot cleaner to start this with `if (dictionaryToConvert == null) return null;` and return early. 

Can then just assign `var lm = new LanguageMap();` and remove a level of nesting

src/protagonist/API/Features/Adjuncts/Requests/GetAllAdjuncts.cs
```diff
@@ -0,0 +1,29 @@
...
+public class GetAllAdjuncts(AssetId assetId) : IRequest<FetchEntityResult<IReadOnlyCollection<Adjunct>>>
+{
+    public AssetId AssetId { get; } = assetId;
+}
+
+public class GetAllAdjunctsHandler(DlcsContext dbContext) : IRequestHandler<GetAllAdjuncts, FetchEntityResult<IReadOnlyCollection<Adjunct>>>
+{
+    public async Task<FetchEntityResult<IReadOnlyCollection<Adjunct>>> Handle(GetAllAdjuncts request, CancellationToken cancellationToken)
+    {
+        var assetExists = dbContext.Images.Any(i => i.Id == request.AssetId);
```
donaldgray: Sub-optimal to make 2 requests to DB to fulfil this request. I'd imagine in real use it'll be rare to call this when the image doesn't exist.

src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
```diff
@@ -0,0 +1,42 @@
...
+        RuleFor(a => a.ModelId)
+            .NotEmpty()
+            .WithMessage("Adjunct identifier could not be found");
+        
+        RuleFor(a => a.Type)
+            .NotEmpty()
+            .WithMessage("'@type' is required");
+    }
+
+    private readonly List<string> validIiifLinkTypes =
```
donaldgray: `Iiif` (also, if using `[Description]` for values then this would need to change too)

src/protagonist/API/Converters/AdjunctConverter.cs
donaldgray: Update after submitting review - noticed that the mapping in dbContext is using `EnumX.GetDescription` so that would keep the value stored in DB consistent with value returned from API

donaldgray (approved):

PR #1089, opened 2026-01-21
Output external adjunct on Manifests

### LINKED ISSUE #1074: Generated manifests include externally hosted manifests
Output externally hosted adjunct in generated manifests (single item and NamedQuery).

Any adjuncts associated with an asset are output in the correct location (as governed by the `iiifLink` property, see below) on the generated Manifest, using the supplied properties.

Adjuncts are always output onto the `Canvas` generated for Asset. The `iiifLink` property will dictate how it is output, each of these properties are an array:
* `seeAlso` - adjunct added to [`"seeAlso"`](https://iiif.io/api/presentation/3.0/#seealso)
* `annotations` - adjunct added to [`"annotations"`](https://iiif.io/api/presentation/3.0/#annotations)
* `rendering` - adjunct added to [`"rendering"`](https://iiif.io/api/presentation/3.0/#rendering)

The properties are then populated from the adjunct db row, e.g.:

```json
{
  "{from-iiifLink}": [
    {
        "id": "{{externalId}}",
        "type": "{{@type}}",
        "format": "{{mediaType}}",
        "profile": "{{profile}}",
        "label": "{{label}}",
        "language": "{{language}}",
        "fileSize": "{{size}}"
    }]
}
```

Adjuncts are only output on IIIF P3 resources, they are ignored for P2.

> [!NOTE]
> Implementation notes/thoughts
> The majority of changes will be in `ManifestV3Builder`.  
> Initial filter adjuncts with `where external_id is not null` to ensure we only output externally hosted manifests. This will be removed in future when we implement hosted adjuncts but is protecting us from these being output in manifests too early.

donaldgray:
Resolves #1074

Externally hosted adjuncts are now output on single-item and NQ manifests in accordance with rules on #1074

Summary of changes:
* Renamed `IncludeRelevantMetadata()` to `IncludeRelationsForProjections()` and extended to include adjuncts. The latter name seems clearer as adjuncts aren't metadata.
* Extend `ManifestV3Builcer` to add adjuncts to relevant canvases.

> [!NOTE]
> As mentioned in ticket `IncludeRelationsForProjections()` uses `.Include(a => a.Adjuncts!.Where(adj => adj.ExternalId != null)`. This is non-functional now but means that when hosted adjuncts are introduced to API they don't prematurely appear in generated Manifest.

> [!NOTE]
> In the case of `iiifLink = "annotations"`, some of the provided Adjuncts values are ignored. I will open a further ticket for returning to this as it's low priority.

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/Adjunct.cs
+++ b/src/protagonist/DLCS.Model/Assets/Adjunct.cs
@@ -72,6 +72,9 @@ public class Adjunct
     public Asset Asset { get; set; } = null!;
 }
 
+/// <summary>
+/// The type of linking property to use for adjunct. Determines how this is output on Manifest.
+/// </summary>
 public enum IIIFLinkType
 {
     [Description("seeAlso")]
--- a/src/protagonist/Orchestrator.Tests/Infrastructure/AssetQueryableXTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Infrastructure/AssetQueryableXTests.cs
@@ -24,27 +24,28 @@ public class AssetQueryableXTests
     }
     
     [Fact]
-    public async Task IncludeRelevantMetadata_HandlesNoRelatedData()
+    public async Task IncludeRelationsForProjections_HandlesNoRelatedData()
     {
         var assetId = AssetIdGenerator.GetAssetId();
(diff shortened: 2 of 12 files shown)


p-kaczynski (approved):
Looks good, to my best knowledge at least :)

PR #1090, opened 2026-01-23
ADR for `maxWidth` and `openFullMax`

donaldgray:


docs/adr/0010-replace-maxunauthorised.md
```diff
@@ -0,0 +1,209 @@
...
+> As well as governing the allowed pixel responses from the level 2 `iiif-img` delivery channel, this value is also considered by the `thumbnail` delivery channel. 
+> The platform will not generate thumbnails larger than the maxWidth limit even if the policy defines them.
+>
+> If the image has roles, and openFullMax and openMaxWidth are both 0, no thumbnails will be produced (this behaviour is not dependent on maxWidth but on the requirement that nothing on the thumbnail delivery channel is subject to access control; 
+> if you offer thumbs they must be accessible anonymously).
+
+`maxWidth` applies to _all_ requests regardless of the region. It is not possible to make a request larger than this size, regardless of any other image property.
+
+The available values are:
+* `<= 0`: Orchestrator won't block requests for sizes that are larger than this. Restrictions may be applied by the downstream image server but from Orchestrator perspective it won't attempt to validate/restrict sizes.
```
tomcrane: Is that right though? Orchestrator can have the system maxWidth as an appsetting, for two reasons:

1) it needs to make sure it's correct in info.json
2) it can short-circuit the request immediately

If it's allowed to proceed to an image server then presumably it will trigger orchestration too if not on Lustre, even though the request is going to fail.

docs/adr/0010-replace-maxunauthorised.md
```diff
@@ -0,0 +1,209 @@
...
+* `> 1`: Orchestrator will check _all_ incoming size requests and will reject if they exceed this value. This may require looking up image dimensions and calculating resulting size (e.g. for `max`, `^max`, `pct:n` and `^pct:n` size parameters). Engine will use to determine which thumbnails to generate.
```
tomcrane: Because it can do this there's little extra overhead in checking for maxWidth<=0 assets; that logic always runs either with the asset-specific value or the system default.

(I assume that this is super-efficient)

docs/adr/0010-replace-maxunauthorised.md
donaldgray: Updated to show that all requests will be checked, falling back to new appSetting.

docs/adr/0010-replace-maxunauthorised.md
donaldgray: See response to other comment

tomcrane (approved):

PR #1102, opened 2026-02-04
Hosted Adjunct - API Management

### LINKED ISSUE #1076: API management of hosted adjuncts
Extend adjuncts to support hosted adjuncts. 

Adjuncts hydra model gains `"origin"`, `"ingesting"` and `"error"` properties.
* `"origin"` works in the same way as `asset.origin` and uses the same `origin strategies` where applicable. When received we have further processing to do.
* Ignore `"size"` value as it will be calculated by Engine. 
  * For overwriting - do not overwrite the `size` if it already exists.
* `"ingesting"` is a boolean marker as to whether processing has completed or not.
* `"error"` records any errors encountered during the _latest_ processing of adjunct.

For PUT and POST, extend API validation to ensure `"externalId"` _or_ `"origin"` are provided. There is no change in behaviour if former provided. If the latter provided:
* Mark the DB record as `ingesting=true` and `error=null` (same as https://github.com/dlcs/protagonist/blob/ddb2e624da9a74f6610697846e736e8b14884773/src/protagonist/DLCS.Model/Assets/AssetX.cs#L79
* Publish an SQS notification to a new 'adjuncts' queue with details of the adjunct.
* Return an HTTP 201 if Adjunct created, else 200

> [!NOTE]
> There are commonalities with `Asset` and `Adjunct` models - is worth introducing base class/interface? (e.g. `origin`, `error`, `ingesting` and eventually `roles`)

p-kaczynski:
Impl #1076

---

POST/PUT Adjunct can now - instead of `ExternalId` - provide `Origin` property. If that happens, we set new property `Ingesting` to `true` and `Error` to `(empty string)`, and a request is put to a new SQS queue.

---

> [!WARNING]
> Requires new queue (`-adjuncts`) and `AWS__SQS__AdjunctQueueName` setting for API to work correctly - otherwise `ThrowIfNull` triggers for `Origin`-having adjuncts.

---

- **Update Adjunct models with ingesting, error and origin**
- **Ingest adjuncts with origin**
- **Add migration for hosted adjuncts**
- **Integration test for hosted adjunct**
- **Apply PublicId to Hydra Adjunct**


src/protagonist/DLCS.Model/Messaging/IngestAdjunctRequest.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Diagnostics;
+using DLCS.Model.Assets;
+
+namespace DLCS.Model.Messaging;
+
+/// <summary>
+/// Represents a request to ingest an adjunct.
+/// </summary>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public class IngestAdjunctRequest(Adjunct adjunct, DateTime? created)
```
donaldgray, line 11: This'll need `AssetId` too, for engine to process.

The adjunct id only needs to be unique within an asset, so just knowing the Adjunct Id could return multiple results, potentially for different customers. Once picked up, Engine will need to do `select * from adjuncts where id = <id> and asset_id = <asset_id>;` rather than just `select * from adjuncts where id = <id>;`

griffri:
LGTM - although I'm not as close to the API's internals as the rest of the team, so it would be good to get their eyes on it.

src/protagonist/DLCS.Model/Messaging/IngestAdjunctRequest.cs
p-kaczynski: The request is not being read by anything yet until 1077, so I think for expediency we could merge as-is and I'll modify `IngestAdjunctRequest` in subsequent 1077 PR.

src/protagonist/DLCS.AWS/SQS/IQueueLookup.cs
```diff
@@ -8,4 +8,6 @@ public interface IQueueLookup
     /// Get the name of queue used for processing assets of specified family
     /// </summary>
     string GetQueueNameForFamily(AssetFamily family, bool priority = false);
-}
\ No newline at end of file
+
+    string GetAdjunctsQueueName();
```
JackLewis-digirati, line 12: could do with a summary comment?

src/protagonist/DLCS.Model/Assets/AdjunctX.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace DLCS.Model.Assets;
+
+public static class AdjunctX
+{
+    /// <summary>
+    /// Semantic helper for code readability. If `origin` has been provided in adjunct, it needs to be ingested into hosted adjunct
+    /// </summary>
+    /// <param name="adjunct">Adjunct for which to determine whether it should be ingested</param>
+    /// <returns></returns>
```
JackLewis-digirati, line 9: missing comment

src/protagonist/DLCS.Model/Assets/DeliverableX.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace DLCS.Model.Assets;
+
+public static class DeliverableX
+{
+    /// <summary>
+    /// Reset fields for ingestion, marking as "Ingesting" and clearing errors
+    /// </summary>
+    public static void SetFieldsForIngestion(this IDeliverable asset)
```
JackLewis-digirati, line 8: comment talks about reset, but this `SetFieldsForIngestion`

JackLewis-digirati (approved):

src/protagonist/DLCS.Model/Assets/DeliverableX.cs
p-kaczynski: That's Donald's comment, I just moved the method/made it work with interface :D

src/protagonist/DLCS.Model/Assets/AdjunctX.cs
p-kaczynski: Added comment but will commit it with 1077 👍

src/protagonist/DLCS.AWS/SQS/IQueueLookup.cs
p-kaczynski: added locally, will push with 1077 (to expedite this PR/unblock etc)

Issue #1093, opened 2026-01-26
Allow `maxWidth` + `openFullMax` to be set per asset - API


donaldgray:
See related [ADR](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md) for `maxUnauthorised` replacement.

Extend the API to allow it to accept `maxWidth` and `openFullMax`.
* Any numeric value is accepted. Values of <=0 are considered unset.
  * `null` values are ignored when received (ie underlying value will remain unchanged)
  * A value of <= 0 will mean the value is cleared.
* `maxWidth` value that exceeds the platform-default `maxWidth` will result in a 400|BadRequest
* Changing `maxWidth` and/or `openFullMax` will result in engine being notified.

These can be provided via any endpoints that can update an asset:

* Upsert asset: `PUT customers/{customer}/spaces/{space}/images/{asset}`
* Update asset `PATCH customers/{customer}/spaces/{space}/images/{asset}`
* Batch patch images: `PATCH customers/{customer}/spaces/{space}/images`
* Create batch: `POST /customers/{customer}/queue`
* Create priority batch: `POST /customers/{customer}/queue/priority`

..and are returned in any endpoint that returns a Hydra image. The values are currently only functional for assets with image and/or thumbs delivery-channels but can be set for any assets.

`maxUnauthorised` still be returned in Hydra image type (to avoid breaking change).

`maxUnauthorised` will still be accepted, however the value will be translated to `maxWidth` and/or `openFullMax` in accordance with logic in ADR. This won't be reliant on LegacyMode, as that was just for Deliverator switchover, which is now long gone.

PR #1103, opened 2026-02-05
Migration for OpenFullMax and MaxWidth

### LINKED ISSUE #1092: Allow `maxWidth` + `openFullMax` to be set per asset - Database

This ticket relates to initial DB configuration.

Create an EF migration to add 2 new columns to Images table: `maxWidth` and `openFullMax`. Ultimately these will be replacing `maxUnauthorised` but, to make deployments simpler, we will leave `maxUnauthorised` in place.

This DB migration must also populate `maxWidth` and `openFullMax` based on current `maxUnauthorised` and `roles` table to ensure any existing behaviour remains, see [ADR use-case / behaviours](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md#use-case--behaviours) for details.

> [!NOTE]
> This may involve configuring an unobtainable role, note that name of that role in this ticket.

donaldgray:
These 2 values will supersede MaxUnauthorised so migration contains UPDATE script for setting new fields appropriately.

PR contains migration only - these are not used in application. UPDATE statement based on https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md#use-case--behaviours

Below shows a table of the different permutations of MaxUnauthorised/Role and how they are transformed to OpenFullMax/Role (MaxWidth will be `null` in all instances). Note that the "classification" value is from update statement:

| asset | max\_unauthorised | original\_roles | open\_full\_max | new\_roles | classification |
| :--- | :--- | :--- | :--- | :--- | :--- |
| max\_minus\_no\_role | -1 |  | 0 |  | all |
| max\_minus\_role | -1 | https://whatever | 0 | https://whatever | need-role |
| max\_zero\_no\_role | 0 |  | 0 | https://dlcs.io/roles/unobtainable | none |
| max\_zero\_role | 0 | https://whatever | 0 | https://whatever | need-role |
| max\_val\_no\_role | 2000 |  | 2000 | https://dlcs.io/roles/unobtainable | full-only |
| max\_val\_role | 2000 | https://whatever | 2000 | https://whatever | zoom-with-role |

> [!WARNING]
> Contains a migration. New columns in "Images" and UPDATE statement. 
> This could be slow / expensive in some environments!

Resolves #1092


compose/docker-compose.db.yml
```diff
@@ -0,0 +1,19 @@
+volumes:
+  dlcs_postgres_data: {}
+  dlcs_postgres_data_backups: {}
+
+services:
+  postgres:
+    image: postgres:14
```
JackLewis-digirati, line 7: is it worth using a newer version of Postgres? 13 was just deprecated and 14 came out nearly 5 years ago

src/protagonist/DLCS.Model/Assets/Asset.cs
```diff
@@ -27,11 +27,23 @@ public class Asset : ICloneable
...
+    /// </summary>
+    /// <remarks>Only applies when an image has roles, and the image request region is /full/.</remarks>
+    public int? OpenFullMax { get; set; }
+
     /// <summary>
     /// The maximum size of longest dimension that is viewable by unauthorised users.
     /// -1 = null (all open), 0 = no allowed size without being auth 
     /// </summary>
+    [Obsolete("Replaced by maxWidth and openFullMax")]
     public int? MaxUnauthorised { get; set; }
```
JackLewis-digirati, line 48: what happens if somebody uses `maxUnauthorised` after the point this code gets released?

JackLewis-digirati (approved):

compose/docker-compose.db.yml
donaldgray: No, I don't think it's worth doing so as part of a 'random' PR. 

We use Postgres14 when running in AWS. It's EOL at the end of this year so we will need to update it before then, I think it's safer for that update to be a concerted effort - all envs, docker-compose etc etc. IIRC there's a change in a later PG version that may affect us, around how permissions are handled.

src/protagonist/DLCS.Model/Assets/Asset.cs
donaldgray: It'll be mapped to `openFullMax` and `roles` to mimic the behaviour of `maxUnauthorised`. Will be done as part of #1093.

We want to maintain handling it for a period, to avoid needing consumers to update their integrations immediately.

donaldgray:
Closing PR - the change to make `maxWidth` and `openFullMax` has broken a large number of tests that will be fixed in follow up PR when API can handle new columns. Will update bring these changes into that PR

PR #1105, opened 2026-02-10
Add openFullMax and maxWidth (API only)

### LINKED ISSUE #977: Attempting to PUT an asset with the `space` property defined results in 500
The "space" field should only be supplied with assets that are part of a batch. If you PUT a single asset that contains it, like so:
```
{
    "mediaType": "video/mp4",
    "origin":"...my_video.mp4",
    "space":"12"
}
```
Protagonist will throw the following error:
```
Asserted space does not agree with supplied space.
```
This is correct, but the response received by the caller is currently a `500` error with no body.
~~The Hydra validator should block usage of this property if the asset is not part of a batch.~~
See https://github.com/dlcs/protagonist/issues/977#issuecomment-2704397273:
> The fix is to make this a 'nicer' error message and 400 response.

### LINKED ISSUE #1092: Allow `maxWidth` + `openFullMax` to be set per asset - Database

### LINKED ISSUE #1093: Allow `maxWidth` + `openFullMax` to be set per asset - API

donaldgray:
Resolves #1092
Resolves #1093
Fixes #977

These 2 values will supersede MaxUnauthorised so migration contains UPDATE script for setting new fields appropriately.

PR contains migration and handling in API - these are not used in Engine or Orchestrator. UPDATE statement based on https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md#use-case--behaviours

## Migration
Below shows a table of the different permutations of MaxUnauthorised/Role and how they are transformed to OpenFullMax/Role (MaxWidth will be `0` in all instances). Note that the "classification" value is from update statement:

| asset | max\_unauthorised | original\_roles | open\_full\_max | new\_roles | classification |
| :--- | :--- | :--- | :--- | :--- | :--- |
| max\_minus\_no\_role | -1 |  | 0 |  | all |
| max\_minus\_role | -1 | https://whatever | 0 | https://whatever | need-role |
| max\_zero\_no\_role | 0 |  | 0 | https://dlcs.io/roles/unobtainable | none |
| max\_zero\_role | 0 | https://whatever | 0 | https://whatever | need-role |
| max\_val\_no\_role | 2000 |  | 2000 | https://dlcs.io/roles/unobtainable | full-only |
| max\_val\_role | 2000 | https://whatever | 2000 | https://whatever | zoom-with-role |

## API

Handle new properties, updating underlying DB as required. In keeping with other properties, a value of `null` from Hydra model means "unchanged". A value of `0` or negative-n means unset, so we explicitly set to `0` (ie no value).

Made various tweaks while in classes to reduce warnings.

> [!WARNING]
> Contains a migration. New columns in "Images" and UPDATE statement. 
> This could be slow / expensive in some environments!


src/protagonist/API/Converters/AssetConverter.cs
```diff
@@ -311,14 +229,155 @@ public static Asset ToDlcsModel(this Image hydraImage, int customerId, int? spac
...
+        {
+            targetAsset.MaxWidth = GetUsableValue(hydraImage.MaxWidth.Value);
+        }
+        
+        if (hydraImage.OpenFullMax != null)
+        {
+            targetAsset.OpenFullMax = GetUsableValue(hydraImage.OpenFullMax.Value);
+        }
+
+        return;
```
JackLewis-digirati, line 369: is this required? Maybe moving the `GetUsableValue` up a bit would mean this could be removed

src/protagonist/API/Converters/AssetConverter.cs
```diff
@@ -311,14 +229,155 @@ public static Asset ToDlcsModel(this Image hydraImage, int customerId, int? spac
...
+        // e.g., used for PUT to a URL, the route params are passed to this method.
+
+        return assetId;
+    }
+    
+    /// <summary>
+    /// Set fields that can control access restrictions: maxWidth, openFullMax and roles
+    /// Also handles maxUnauthorised value, which is deprecated but will be supported for a short period 
+    /// </summary>
+    private static void SetSizeRestriction(Image hydraImage, Asset targetAsset)
```
JackLewis-digirati, line 327: I'm not quite sure what, but something about this method feels a bit strange... maybe the void returns? Could use an `if/else` or move the `maxUnauthorised` to a separate method?

src/protagonist/DLCS.Model.Tests/Assets/AssetPreparerTests.cs
```diff
@@ -276,21 +305,11 @@ public void PrepareAssetForUpsert_RequiresReingest_IfDeliveryChannelChanged(stri
...
-        var imageDeliveryChannels = new List<ImageDeliveryChannel>();
-
-        foreach (var deliveryChannel in deliveryChannels)
-        {
-            imageDeliveryChannels.Add(new ImageDeliveryChannel()
-            {
-                Channel = deliveryChannel
-            });
-        }
+        return;
```
JackLewis-digirati, line 308: Feels weird to have a `return` in a test, but I can see it's because of the below method

JackLewis-digirati (approved):

src/protagonist/DLCS.Model.Tests/Assets/AssetPreparerTests.cs
donaldgray: Yeah, it's because of the local function, I think it reads better without but Rider keeps recommending to add https://www.jetbrains.com/help/resharper/SeparateLocalFunctionsWithJumpStatement.html

src/protagonist/API/Converters/AssetConverter.cs
donaldgray: `return` can be removed, it's optional. Added on guidance from Rider (https://www.jetbrains.com/help/resharper/SeparateLocalFunctionsWithJumpStatement.html) as it's before a local function.

src/protagonist/API/Converters/AssetConverter.cs
donaldgray: For future reference: we had discussion but leaving as-is.

All logic around setting properties that can restrict what sizes are available is contained within. `openMaxWidth` will be included in here when introduced, `maxUnauthorised` can be removed from here when finally removed. A separate `maxUnauthorised` would be easier to read but would mean moving `GetUsableValue` local function to a private, which I'm reluctant to do as it's only for the couple of properties set within - not a general use method.

Issue #1077, opened 2025-12-02
Engine handles adjuncts


donaldgray:
Update Engine to handle hosted adjuncts. After processing the adjunct should be stored in a location where it can serve public requests.

The actual processing of an adjunct will be very similar to how assets with `file` delivery-channels are processed. The overall process is:
* Subscribe to queue and on receipt of message..
* Work out what `CustomerOriginStrategy` (`COS`) to use for origin.
* If `COS` is `optimised` then we don't need to copy the file to DLCS s3 storage location. Else..
* Copy binary to DLCS S3 storage location.
* Finally mark asset as completed (`ingesting=true`, `finished=now()`, `error=?`, `size=?`).
* Save details of any `ImageSize` 

Extend the `ImageStorage` table to have a new column `TotalSizeOfAdjuncts`.
* `TotalSizeOfAdjuncts` must account for the action being an update and not double count storage.

A note for testing:
* At this point we are unable to serve the adjunct through Orchestrator. The only way to check success or not is to fetch the adjunct via the API and see the `ingesting=true`, `finished=now()`, `error=?`, `size=?` properties are updated.

*Implementation Notes*

Below are some relevant classes used in Engine for `file` processing. These are likely tied to `Asset` but we should look to reuse where we can.

* `FileChannelWorker` is the worker that handles `file` assets (as part of a larger pipeline that is unnecessary).
* `IStorageKeyGenerator` is where all S3 key generation logic should live (corresponding readme should be updated too). 
  * This should be in the storage bucket, under `/{customer}/{space}/{asset}/adjuncts/{adjunct-id}`
* `AssetToS3` has logic to either do a 'direct' (ie s3->s3) or 'indirect' (ie origin->disk->s3) copy depending on `COS`. Returns an `AssetFromOrigin` that stores `Asset` size.
* `IngestionContext` stores details of the item being processed, a lot of the classes will take this as a parameter so refactoring of this is likely required.

> [!IMPORTANT]
> Do we want to keep things like `AssetFromOrigin` and extend (ie maintain the `AssetId` and add extra properties). Could be useful for storing data.

> [!WARNING]
> We will need to store an `ImageLocation`-like value somewhere. This tells Orchestrator where to stream the contents from. It'll either be the location where the Origin has been copied to, or it will be the origin itself, if optimised s3 ambient `COS`

christinemacdonald:
Ticket has passed testing

PR #1108, opened 2026-02-11
Set missing ingesting/error prop on adjunct hydra output model

### LINKED ISSUE #1106: `"ingesting":true` not returned for new hosted adjuncts
Currently, you can create a hosted adjunct with an origin pointing towards a public resource, like so:
```JSON
{
  "id": "hosted_adjunct_test_4",
  "origin": "https://dlcsstage-public-test-objects.s3.eu-west-1.amazonaws.com/other-video/yellowstone_mugwump_geyser.mp4",
  "@type": "Dataset",
  "profile": "http://www.example.com/profile2.xsd",
  "mediaType": "video/mp4",
  "label": {
    "en": [
      "Hosted adjunct test 3 - async"
    ]
  },
  "iiifLink": "rendering",
  "language": ["en"]
}
```

...but it won't return an `"ingesting"` property set to `true` in the response payload.

p-kaczynski:
Fixes #1106 

Set newly created `error` and `ingesting` properties when converting to a Hydra output model for Adjuncts.

Included the props in the test.

`finished` handling will be done in #1077 - for now it will remain `now()`

### DIFF
--- a/src/protagonist/API/Converters/AdjunctConverter.cs
+++ b/src/protagonist/API/Converters/AdjunctConverter.cs
@@ -84,5 +84,7 @@ public static class AdjunctConverter
             Created = adjunct.Created,
             Finished = adjunct.Finished,
             Size = adjunct.Size,
+            Error = adjunct.Error,
+            Ingesting = adjunct.Ingesting
         };
 }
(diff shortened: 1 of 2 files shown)


src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -84,5 +84,7 @@ public static DLCS.HydraModel.Adjunct ToHydra(this Adjunct adjunct, UrlRoots url
             Created = adjunct.Created,
             Finished = adjunct.Finished,
             Size = adjunct.Size,
+            Error = adjunct.Error,
+            Ingesting =  adjunct.Ingesting
```
griffri: There's an extra space here - should be  `Ingesting = adjunct.Ingesting`

griffri (approved):

PR #1115, opened 2026-02-19
Engine handles hosted adjuncts

### LINKED ISSUE #1077: Engine handles adjuncts

p-kaczynski:
Impl #1077 
---

> [!NOTE]
> Contains migrations necessary for handling hosted Adjuncts

---

This PR brings the ability for Engine to ingest adjuncts that were submitted with `Origin` property.

The implementation is mostly based on the `Asset` ingestion via `File` channel. Where possible, code was reused via either leveraging inheritance or interfaces. Still, some subtle differences warranted numerous new classes/methods etc.

Modifications were made to many pre-existing methods that were written with limited extensibility/reusability. Specifically number of methods were taking `AssetId` only to use it in logging. This was modified to a `string` that can be either `AssetId.ToString()` or a longer description of adjunct within asset.

The SQS handling was expanded to use `MessageAttributes` and inform the request deserialization step as to which ingest request class this is (Asset or Adjunct).

---
Commits:

- **Add missing comments**
- **Include custom attributes in ingest messages to differentiate Assets and Adjuncts**
- **Partial implementation of engine hosted adjunct ingest (WIP)**
- **Make Adjunct.Finished nullable. Set property based on scenario**
- **Move `Finished` property to `IDeliverable` and allow reuse `MarkAsFinished`**
- **Expand Asset ingestion to support Adjuncts (WIP)**
- **Add Adjunct S3 storage key generation**
- **Add TemplatedFolder method for generating path for Adjuncts**
- **Make origin strategies take arbitrary string instead of AssetId and update tests**
- **Subclass AssetFromOrigin to AdjunctFromOrigin that has an AdjunctId prop**
- **Extend StorageX method to get disk-safe file name that can operate on plain string**
- **Extend FileChannelWorker to handle Adjuncts**
- **Fix errors in tests introduced by changes in previous commits**
- **Finish Adjunct ingestion flow including storage size tracking (minus migration)**
- **Add FileChannelWorker positive tests for adjuncts**
- **Add FileChannelWorker tests for Adjuncts (negative/other)**
- **ImageStorage migration to add Adjunct size tracking**
- **Implement adjunct size tracking/calculation and finish IngestExecutor impl for adjuncts**


src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -50,19 +53,53 @@ public async Task<ModifyEntityResult<Adjunct>> Handle(CreateOrUpdateAdjunct requ
...
+            {
+                // For hosted (ingested) adjuncts we let Engine handle this property
+                // as it becomes relevant to storage limits. However, if the pre-existing
+                // adjunct was EXTERNAL, the size doesn't count toward those limits.
+                
+                // To ensure correct calculations in the engine, we will set Size to null.
+                // This will allow Engine to increase the total adjunct size by the size
+                // of new version of the adjunct, regardless what size the external one had.
+
+                dbAdjunct.Size = null;
```
donaldgray, line 67: https://github.com/dlcs/protagonist/issues/1076 has

> * Ignore "size" value as it will be calculated by Engine.
>   * For overwriting - do not overwrite the size if it already exists.

Can we leave this size as is, or maybe only `null` it if moving from not-hosted to hosted adjunct? Until the Engine overwrites the value we are still storing the `size` specified? E.g.

00:00. Ingest adjunct `plain.txt` at  and engine sets size as 100 KB
00:10. Queue up 100K hosted adjuncts to ingest. This is a mix between updates and inserts. They're big, will take 2 hours to process. As these go through api all sizes are `null`.
00:12. Realise `plain.txt` had a typo and PUT it. `size` is now `null` but we're still storing the 100KB. This won't be populated until the above 100KB are finished (say, 02:00)
00:30. The weekly recalculator job runs and it aggregrates what I'm storing but misses some storage values off because some of the above items have a `null` size but we're still storing the value.

Equally if some ingests fail, we'd be left with `size=null` when we will still be holding onto data in S3.

src/protagonist/DLCS.AWS/S3/S3StorageKeyGenerator.cs
```diff
@@ -51,13 +52,19 @@ public RegionalisedObjectInBucket GetStorageLocation(AssetId assetId)
...
+
     public ObjectInBucket GetStorageLocationRoot(AssetId assetId)
     {
         var key = GetStorageKey(assetId);
         return new ObjectInBucket(s3Options.StorageBucket, $"{key}/");
     }
 
+    public RegionalisedObjectInBucket GetStoredAdjunctLocation(AssetId assetId, Adjunct adjunct)
+    {
+        var key = $"{GetStorageKey(assetId)}/adjuncts/{adjunct.Id}";
```
donaldgray, line 64: Small - can this path be added to https://github.com/dlcs/protagonist/blob/develop/docs/storage-keys.md? It's obvs all here but the readme can be handy if wanting to check S3 directly without needing to read through code/tests

src/protagonist/DLCS.Repository/Customers/CustomerOriginStrategyBase.cs
```diff
@@ -69,6 +70,20 @@ public async Task<CustomerOriginStrategy> GetCustomerOriginStrategy(Asset asset,
         return matching;
     }
     
+    public async Task<CustomerOriginStrategy> GetCustomerOriginStrategy(Adjunct adjunct)
```
donaldgray: Is it worth having an internal `GetCOS(IDeliverable)` to share? Realise it's only a couple of lines shared but could handle the fallback to default etc

src/protagonist/DLCS.Repository/Migrations/20260212173058_AdjunctFinishedMakeNullable.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
```
donaldgray, line 1: Can you stick a `> [!NOTE]` in PR description that this has a migration? Will make it easier to spot when generating release notes

src/protagonist/DLCS.Repository/Strategy/Utils/FileSaver.cs
```diff
@@ -13,12 +13,12 @@ public interface IFileSaver
...
     /// Save asset from <see cref="OriginResponse"/> to specified file location.
     /// </summary>
-    /// <param name="assetId">Id of asset being saved</param>
+    /// <param name="itemDesc"></param>
     /// <param name="originResponse"><see cref="OriginResponse"/> object containing data stream</param>
     /// <param name="destination">Location to store binary to, will be deleted if already exists</param>
     /// <param name="cancellationToken">Async cancellationToken</param>
     /// <returns>ContentLength</returns>
-    Task<long> SaveResponseToDisk(AssetId assetId, OriginResponse originResponse, string destination,
+    Task<long> SaveResponseToDisk(string itemDesc, OriginResponse originResponse, string destination,
```
donaldgray: in keeping with shared interface should this be something like `string deliverableId` (could it even take an IDeliverable, could that have a `string Identifier()` or something?

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -50,19 +53,53 @@ public async Task<ModifyEntityResult<Adjunct>> Handle(CreateOrUpdateAdjunct requ
...
+            } else {
```
JackLewis-digirati: formatting

src/protagonist/Engine/Ingest/Persistence/AssetToDisk.cs
```diff
@@ -60,26 +65,74 @@ public AssetToDisk(
...
+    /// <summary>
+    /// Copy adjunct from Origin to local disk.
+    /// </summary>
+    /// <param name="context">Ingestion context containing the <see cref="Adjunct"/> to be copied.</param>
+    /// <param name="destinationTemplate">String representing destinations folder to copy to.</param>
+    /// <param name="verifySize">if True, size is validated that it does not exceed allowed size.</param>
+    /// <param name="customerOriginStrategy"><see cref="CustomerOriginStrategy"/> to use to fetch item.</param>
+    /// <param name="cancellationToken"><see cref="CancellationToken"/>Current cancellation token</param>
+    /// <returns><see cref="AdjunctFromOrigin"/> containing new location, size etc</returns>
+    public async Task<AdjunctFromOrigin> CopyAdjunctToLocalDisk(AdjunctIngestionContext context,
```
donaldgray: Given how close this is to `AssetToDisk` can there be shared code between, I think `AdjunctIngestionContext` inheriting from `IngestionContext` might help?

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -50,19 +53,53 @@ public async Task<ModifyEntityResult<Adjunct>> Handle(CreateOrUpdateAdjunct requ
...
+                dbAdjunct.Size = adjunct.Size;
+            }
         }
         else
         {
             dbAdjunct = adjunct;
             dbAdjunct.Created = DateTime.UtcNow;
-            dbAdjunct.Finished = DateTime.UtcNow;
+
+            if (toBeIngested)
```
JackLewis-digirati, line 86: this is the same as the above - worth refactoring into a single place?

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -50,19 +53,53 @@ public async Task<ModifyEntityResult<Adjunct>> Handle(CreateOrUpdateAdjunct requ
             dbAdjunct.ExternalId = adjunct.ExternalId;
             dbAdjunct.Origin = adjunct.Origin;
             dbAdjunct.Error = adjunct.Error;
-            dbAdjunct.Finished = DateTime.UtcNow;
-            dbAdjunct.Size = adjunct.Size;
             dbAdjunct.Type = adjunct.Type;
+
+            if (toBeIngested)
```
JackLewis-digirati: is there some tests that need to be added for these api changes?

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -10,26 +10,89 @@ namespace Engine.Ingest;
...
-    private readonly IAssetIngestorSizeCheck assetIngestorSizeCheck;
-    private readonly IStorageRepository storageRepository;
     private const int MinimumAssetSize = 100;
 
-    public IngestExecutor(IWorkerBuilder workerBuilder, 
-        IEngineAssetRepository assetRepository,
-        IAssetIngestorSizeCheck assetIngestorSizeCheck,
-        IStorageRepository storageRepository,
-        ILogger<IngestExecutor> logger)
+    public async Task<AdjunctIngestResult> IngestAdjunct(Adjunct adjunct, CustomerOriginStrategy customerOriginStrategy,
```
donaldgray: Reuse/share with asset ingest method?

src/protagonist/DLCS.AWS/S3/IStorageKeyGenerator.cs
```diff
@@ -172,4 +173,13 @@ public interface IStorageKeyGenerator
     /// <param name="assetId">Id of Asset being transcoded</param>
     /// <param name="jobId">Unique job identifier</param>
     ObjectInBucket GetTranscodeDestinationRoot(AssetId assetId, string jobId);
+
+    /// <summary>
+    /// Get <see cref="ObjectInBucket"/> for stored adjunct file.
+    /// This is for adjuncts which all use "file" delivery
+    /// </summary>
+    /// <param name="assetId">id of Asset owning this Adjunct</param>
```
JackLewis-digirati: tiny - lower case at the start of the sentence

src/protagonist/DLCS.AWS/SQS/IQueueSender.cs
```diff
@@ -7,21 +7,24 @@ public interface IQueueSender
...
     /// <returns>Boolean value indicating success</returns>
-    Task<bool> QueueMessage(string queueName, string messageContents, CancellationToken cancellationToken = default);
+    Task<bool> QueueMessage(string queueName, string messageContents, IDictionary<string, string> messageAttributes,
+        CancellationToken cancellationToken = default);
 
     /// <summary>
     /// Queue messages to specified queue.
     /// 
-    /// This is more efficient than calling <see cref="QueueMessage"/> multiple times as it will batch creation of
+    /// This is more efficient than calling <see cref="QueueMessage(string,string,IDictionary{string,string},System.Threading.CancellationToken)"/> multiple times as it will batch creation of
```
JackLewis-digirati, line 19: does this need the parameters added to the comment?

src/protagonist/DLCS.AWS/SQS/SqsQueueSender.cs
```diff
@@ -51,16 +57,20 @@ public async Task<int> QueueMessages(string queueName, IReadOnlyCollection<strin
             {
                 var batchPrefix = $"{batchIdentifier}_{++batchCount}";
                 var entries = batch
-                    .Select(c => new SendMessageBatchRequestEntry($"{batchPrefix}_{++count}", c))
+                    .Select(c => new SendMessageBatchRequestEntry($"{batchPrefix}_{++count}", c)
+                    {
+                        // Note: It seems recommended to have an instance-per-entry, not reuse the same one
+                        MessageAttributes = GetMessageAttributesDictionary(messageAttributes)
```
JackLewis-digirati, line 63: I see the comment says not to reuse, but I believe this method recreates a dictionary (due to `ToDictionary`) every time, which feels inefficient

src/protagonist/DLCS.Repository/Strategy/BasicHttpAuthOriginStrategy.cs
```diff
@@ -21,10 +21,10 @@ public class BasicHttpAuthOriginStrategy(
     ILogger<BasicHttpAuthOriginStrategy> logger)
     : IOriginStrategy
 {
-    public async Task<OriginResponse> LoadAssetFromOrigin(AssetId assetId, string origin,
+    public async Task<OriginResponse> LoadAssetFromOrigin(string itemDesc, string origin,
```
JackLewis-digirati: I know this was done so it could handle adjunct and asset - but would it be cleaner to have a pair of public methods that take in either an adjunct or asset then call this method? Would mean that all of the `ToString()` calls could be avoided, and we're only passing strings around in this class as opposed to the codebase?

src/protagonist/DLCS.Repository/Migrations/20260219101639_ImageStorageAdjunctTracking.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
JackLewis-digirati, line 1: There's a pair of migrations here - could these be condensed into a single one as what's needed to get hosted adjuncts to work in engine?

src/protagonist/DLCS.Repository/Strategy/Utils/FileSaver.cs
```diff
@@ -13,12 +13,12 @@ public interface IFileSaver
     /// <summary>
     /// Save asset from <see cref="OriginResponse"/> to specified file location.
     /// </summary>
-    /// <param name="assetId">Id of asset being saved</param>
+    /// <param name="itemDesc"></param>
```
JackLewis-digirati: empty comment

src/protagonist/DLCS.Repository/Strategy/Utils/OriginFetcher.cs
```diff
@@ -37,23 +27,23 @@ public async Task<OriginResponse> LoadAssetFromLocation(AssetId assetId, string
...
-        return await LoadAssetFromLocation(assetId, location, customerOriginStrategy, cancellationToken);
+        return await LoadAssetFromLocation(assetId.ToString(), location, customerOriginStrategy, cancellationToken);
     }
-    
+
     /// <summary>
     /// Get <see cref="OriginResponse"/> object for provided asset, loading from origin passed origin strategy
     /// </summary>
-    /// <param name="assetId">Asset to load</param>
+    /// <param name="itemDesc">used for logging - should allow identifying the asset/adjunct/other item (future)</param>
```
JackLewis-digirati: tiny - lowercase start

src/protagonist/Engine.Tests/Integration/AdjunctIngestTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    private readonly JsonSerializerOptions settings = new(JsonSerializerDefaults.Web);
+    private readonly DlcsContext dbContext;
+    private static readonly TestBucketWriter BucketWriter = new();
+    private readonly ApiStub apiStub;
+    
+    // These spaces are used in tests 
+    private const int CustomerForLimits = -10;
+    private const int SpaceExceedLimit = 1;
+    private const int SpaceWithinLimit = 2;
+    private const int SpaceExceedLimitReingest = 3;
```
JackLewis-digirati: above variables aren't used - worth writing tests to use them?

src/protagonist/DLCS.Repository/Strategy/Utils/FileSaver.cs
JackLewis-digirati: similar to `saveImageLocation` - would it be worth adding a pair of public methods that accept an assetid/adjunct and internally call this method to save having `ToString` everywhere?

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -147,7 +179,7 @@ private static void UpdateBatchAsset(Asset asset, BatchAsset batchAsset)
         batchAsset.Finished = DateTime.UtcNow;
     }
 
-    private static void UpdateAsset(Asset asset, bool ingestFinished)
+    private static void UpdateItem(IDeliverable asset, bool ingestFinished)
```
JackLewis-digirati: Given this is taking in an `IDeliverable` would `UpdateDeliverable` be better? Additionally `IDeliverable asset` no longer just works for asset

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -81,6 +62,32 @@ public async Task<bool> UpdateIngestedAsset(Asset asset, ImageLocation? imageLoc
         }
     }
 
+    public async Task<bool> UpdateIngestedAdjunct(Adjunct adjunct, ImageStorage? imageStorage, bool ingestFinished,
```
JackLewis-digirati: This method has a fair number of similarities with `UpdateIngestedAsset` - can any of the logic be shared?

src/protagonist/Engine/Ingest/Persistence/AssetToS3.cs
```diff
@@ -28,112 +28,228 @@ Task<AssetFromOrigin> CopyOriginToStorage(ObjectInBucket destination, IngestionC
...
+        return await IndirectAssetCopyBucketToBucket(context, destination, verifySize, customerOriginStrategy,
             cancellationToken);
     }
 
-    private bool ShouldCopyBucketToBucket(CustomerOriginStrategy customerOriginStrategy)
+    private static bool ShouldCopyBucketToBucket(CustomerOriginStrategy customerOriginStrategy)
         => customerOriginStrategy is { Strategy: OriginStrategyType.S3Ambient };
 
-    private async Task<AssetFromOrigin> CopyBucketToBucket(IngestionContext context, ObjectInBucket destination, bool verifySize,
+    private async Task<AdjunctFromOrigin> CopyAdjunctBucketToBucket(AdjunctIngestionContext context,
```
JackLewis-digirati: logic is similar to `CopyAssetBucketToBucket` - can any of this be shared?

src/protagonist/Engine/Ingest/AssetIngester.cs
```diff
@@ -59,14 +47,14 @@ public async Task<IngestResult> Ingest(IngestAssetRequest request, CancellationT
...
+}
 
-    public IngestResult(AssetId? assetId, IngestResultStatus ingestResult)
-    {
-        AssetId = assetId;
-        Status = ingestResult;
-    }
-}
\ No newline at end of file
+public class AdjunctIngestResult(string adjunctId, AssetId? assetId, IngestResultStatus ingestResult)
```
JackLewis-digirati: should this be moved to the `AdjunctIngester`?

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -10,26 +10,89 @@ namespace Engine.Ingest;
...
+            context.WithPreIngestionAssetSize(preIngestionAssetSize);
+        }
+
+        var workers = workerBuilder.GetWorkers(adjunct);
+        var overallStatus = IngestResultStatus.Unknown;
+        var postProcessors = new List<IAdjunctIngesterPostProcess>(workers.Count);
+
+        foreach (var worker in workers)
+        {
+            // ReSharper disable once SuspiciousTypeConversion.Global - hook for future impl
```
JackLewis-digirati: maybe some more info than `hook for future impl`?

JackLewis-digirati (commented):
Generally ok with implementation, but seems like there's a lot of code that could be shared with the asset ingestion flow

src/protagonist/DLCS.Repository/Strategy/Utils/FileSaver.cs
p-kaczynski: I like it!

src/protagonist/DLCS.AWS/SQS/IQueueSender.cs
p-kaczynski: Rider complains otherwise

src/protagonist/DLCS.AWS/SQS/SqsQueueSender.cs
p-kaczynski: because it can be null, it's not very convenient to do it via
```
var dict = Get...Dict();
...
MessageAttributes = new(dict);
```
it would need either some `dict == null ? null : new(dict)` constructs or a separate method that encapsulates it.,

 it's likely microsecond level of effort at most - I'd wager a guess that just instanciating the dict (as recommended) takes more time than populating it with like 1 item.

src/protagonist/DLCS.Repository/Strategy/BasicHttpAuthOriginStrategy.cs
p-kaczynski: Donald had the same suggestion, resolved with some interfaces etc.

src/protagonist/DLCS.Repository/Migrations/20260219101639_ImageStorageAdjunctTracking.cs
p-kaczynski: honestly keeping it separate (and on separate commits) was very useful during development, so while "at the end of development" it is indeed condensable, seems like "not worth the effort".

src/protagonist/Engine.Tests/Integration/AdjunctIngestTests.cs
p-kaczynski: as mentioned on slack - yeah, it's "TODO" - wanted to get the review ASAP due to scheduling. Will be done soon(tm)

src/protagonist/DLCS.Repository/Strategy/Utils/FileSaver.cs
p-kaczynski: `ToString` thing resolved "globally", should be much cleaner now

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
p-kaczynski: Yes, the comments show that was my thinking (null it if it WAS external), but the `if/else` did not reflect that.

I have flipped the `if` and changed the case above to `else if` that will check whether the Adjunct that existed in DB was ingestable(hosted). Only if it was external (not ingestable), and now it is gonna be hosted, we null the size (as it wasn't "our hosted size").

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
p-kaczynski: it changed a bit due to Donald spotting there was insufficient conditions checked, so that's null&void, but in general it would be ever more clunky imho, as the check above has different meaning depending on whether this is create or update.

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
p-kaczynski: yup, "todo"

src/protagonist/DLCS.AWS/SQS/IQueueLookup.cs
```diff
@@ -9,5 +9,9 @@ public interface IQueueLookup
     /// </summary>
     string GetQueueNameForFamily(AssetFamily family, bool priority = false);
 
+    /// <summary>
+    /// Get the name of queue used for processing adjuncts
+    /// </summary>
+    /// <returns></returns>
```
JackLewis-digirati: missing comment

src/protagonist/DLCS.AWS/SQS/SqsQueueSender.cs
```diff
@@ -37,30 +32,45 @@ public async Task<bool> QueueMessage(string queueName, string messageContents,
         }
     }
 
+    /// <summary>
+    /// Converts CLR string-string dictionary to one with Amazon SQS specific <see cref="MessageAttributeValue"/>
+    /// </summary>
+    /// <param name="messageAttributes">attributes as string-string dictionary or null if no custom message attributes are needed</param>
+    /// <returns>new instance of a string-<see cref="MessageAttributeValue"/> dictionary</returns>
```
JackLewis-digirati: tiny - lower case start to senetence

src/protagonist/DLCS.AWS/SQS/SqsQueueUtilities.cs
```diff
@@ -11,6 +11,14 @@ namespace DLCS.AWS.SQS;
 /// </summary>
 public class SqsQueueUtilities
 {
+    public static class Constants
+    {
+        public static class MessageAttributeNames
+        {
+            public const string IngestType = "IngestType";
```
JackLewis-digirati, line 24: public variable - does this need a comment?

src/protagonist/Engine.Tests/Ingest/Persistence/AssetToDiskTests.cs
```diff
@@ -33,14 +33,14 @@ public AssetToDiskTests()
 
         var engineSettings = new EngineSettings { ImageIngest = new ImageIngestSettings() };
         var optionsMonitor = OptionsHelpers.GetOptionsMonitor(engineSettings);
-        
-        
+
+
```
JackLewis-digirati: additional whitespace

src/protagonist/Engine/Ingest/AdjunctIngester.cs
```diff
@@ -0,0 +1,59 @@
+﻿using DLCS.Core.Types;
+using DLCS.Model.Customers;
+using DLCS.Model.Messaging;
+using Engine.Data;
+
+namespace Engine.Ingest;
+
+public interface IAdjunctIngester
+{
+    Task<IngestResult> Ingest(IngestAdjunctRequest request, CancellationToken cancellationToken = default);
```
JackLewis-digirati, line 14: missing comment

JackLewis-digirati (approved):
LGTM, but I do think the tests should be added before this gets merged as I feel they will get lost if they aren't done now

PR #1116, opened 2026-02-23
Replace `maxUnauthorised` in Orchestrator

### LINKED ISSUE #1095: Allow `maxWidth` + `openFullMax` to be set per asset - Orchestrator

All images have an 'effective maxWidth'. This will be the image-specific value, if set, or the system-default value.

Effective `maxWidth` is included on all info.json responses. 

All ImageRequests are rejected if the 'size' value would exceed effective `maxWidth`.

`/max/` size should be proxied to `/full/{maxWidth,}/`

ImageRequest handling logic updated to match [logic in ADR](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md#orchestrator-1).

PDF generation takes into account `maxWidth` value and redacts pages/requires auth depending on values (_update this with what that logic is when investigated_)

* [x] Image requests with sizes that would exceed `maxWidth` are rejected with a 403.
* [x] Sizes calculated for generated manifests that don't have AssetApplicationMetadata take new logic into account.
* [x]  `/max/` or `/full/` size requests should be proxied to specific size.
* [x] All image Info.json advertise `"maxWidth"` property.
* [x] Info.json do not render auth-services for unobtainable role.
* [x] Any `"max*"` properties (`"maxArea"` or `"maxHeight"`) added by the image server are removed.
* [x] Orchestrator will serve/not-serve assets depending on available sizes, as per [logic in ADR](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md#orchestrator-1)
* [x] PDF generation uses same logic as now, without checking maxWidth. Any roles are checked against specified whitelist to determine what images are included on PDF and which are redacted.
  * Serving PDFs enforces auth in same way as now

> [!WARNING]
> This is a fairly fundamental change to how Orchestrator works as we now interogate _every_ incoming request and potentially reject / rewrite requests.

donaldgray:
Resolves #1095

Update Orchestrator to use `maxWidth` and `openFullMax` in image processing, rather than `maxUnauthorised`.

> [!WARNING]
> This change moves more IIIF ImageApi parsing logic into Orchestrator.
> Previously we relied on downstream image servers for this but now more Region + Size parameter processing is in Orchestrator.
> It also changes IIIF ImageApi parameters when making downstream proxy requests, which is a change in behaviour.

Included ADR outlines ImageApi parsing, how it differs per version and the possibility for "strict" and "lax" modes.

Biggest change is introduction of `ImageProxyPathHandler` class. This takes incoming IIIF ImageApi request and parses it to work out the resulting image size, whether this is valid and what the downstream request would be (via https://github.com/digirati-co-uk/iiif-net/pull/86). In the event of a `/max/`, or equivalent, size it will return the new SizeParameter to proxy to. Also - `/max/`, or equivalent sizes will be checked to see if they can be served by thumbs. This wasn't done previously as it would involve rewriting the size parameter but we now have values to do this. The tests for this class are verbose but pretty cumbersome - any suggestions to improve welcome!

> [!CAUTION]
> Orchestrator now has `MaxWidth` property - downstream image-server may need altered to allow requests through OR to support existing images.

Extended `IsFullOrEquivalent` logic to handle `/pct:0,0,100,100/` as full request.

Info.json construction logic is updated, `maxWidth` will be output in all info.json requests.

> [!WARNING]
> To view these changes all cached info.json files will need to be recreated by setting the `OldestAllowedInfoJson` appSetting value.
> Recreating info.json requires orchestration.

Special handling was added for "unobtainable" role introduce in earlier PR. We won't advertise access-services for this role and will shortcut some requests if this is the only role for an image.

Updated cleanup-handler to delete info.json if the `maxWidth` value has changed as this will need to be reflected in info.json (previously this only happened if `roles` changed).

Thumbnails that exceed `maxWidth` are not saved at all.

### DIFF
--- a/src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ImageProxyPathHandler.cs
+++ b/src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ImageProxyPathHandler.cs
@@ -1,10 +1,12 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Net;
 using DLCS.Model.IIIF;
 using IIIF;
 using IIIF.Exceptions;
 using IIIF.ImageApi;
+using Version = IIIF.ImageApi.Version;
 
 namespace Orchestrator.Infrastructure.ReverseProxy;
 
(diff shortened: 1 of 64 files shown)


src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ImageProxyPathHandler.cs
```diff
@@ -0,0 +1,224 @@
...
+        var isV2 = imageVersion == Version.V2;
+        var isExplicitFull = imageRequest.IsExplicitFullSize();
+
+        var failedValidationError = TryValidateRequest(imageRequest, isV2, isExplicitFull, strictMode);
+        if (failedValidationError != null) return failedValidationError;
+        
+        return HandleProxyLogic(imageRequest, imageSize, maxWidth, isV2, isExplicitFull);
+    }
+
+    private static ProxyImageRequest? TryValidateRequest(ImageRequest imageRequest, bool isV2, bool isExplicitFull,
```
JackLewis-digirati: tiny, but is this `TryValidate`? i.e.: we only get a result if there's errors (it's null on success rather than a validation success), so feels more like `CheckErrors`?

JackLewis-digirati (approved):

src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ImageProxyPathHandler.cs
donaldgray: Doesn't follow the idiomatic return `bool` with `out` parameter for TryXXX so renamed to `GetValidationFailure` as that's what trying to do. If it returns `null` then no validation failure.

PR #1118, opened 2026-02-27
Add InlineAnnotation to the API

### LINKED ISSUE #1112: Support `inlineAnnotation` type in API
Accept `inlineAnnotation` as a new link type. 

* [x] Extend the accepted `iiifLink` types to include `inlineAnnotation`.
* [x] Support new `motivation` property. This is optional and only used for `inlineAnnotation` types but can be supplied for other types. Value is a string. The values should be taken from the W3C Web Annotation specification, or from the additional IIIF motivations provided by the Presentation API (e.g., "supplementing") - however the API will not validate against a given list.
* [x] Update validation logic for `iiifLink` type of `annotations`, it is required that the `"@type"` is `"AnnotationPage"`.
  * We could default this but we should require the consumer to explicitly specify this.

JackLewis-digirati:
Resolves #1112 

This PR adds the `InlineAnnotation` iiif link type to the API.  Additionally, `motivation` is now supported, along with a migration to add this to the database

> [!Note]
> This PR contains a migration

### DIFF
--- a/src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
+++ b/src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
@@ -58,7 +58,7 @@ public class HydraAdjunctValidator : AbstractValidator<DLCS.HydraModel.Adjunct>
         RuleFor(a => a.Type)
             .Must(t =>  t == "AnnotationPage")
             .When(a => a.IIIFLink == IIIFLinkType.Annotations.GetDescription())
-            .WithMessage($"When the 'iiifLink' is '{IIIFLinkType.Annotations.GetDescription()}', the '@type' must be 'AnnotationPage'");
+            .WithMessage($"When 'iiifLink' is '{IIIFLinkType.Annotations.GetDescription()}', the '@type' must be 'AnnotationPage'");
     }
 
     private readonly List<string> validIIIFLinkTypes =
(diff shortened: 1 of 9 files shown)


src/protagonist/DLCS.HydraModel/Adjunct.cs
```diff
@@ -81,18 +81,23 @@ public Adjunct(string baseUrl, int customerId, int space, string assetId, string
...
     public DateTime? Finished { get; set; }
     
-    [RdfProperty(Description = "Origin endpoint from where the original adjunct can be acquired (or was acquired)",
+    [RdfProperty(Description = "Origin endpoint from where the original adjunct can be acquired (or was acquired).",
         Range = Names.XmlSchema.String, ReadOnly = false, WriteOnly = false)]
     [JsonProperty(Order = 23, PropertyName = "origin")]
     public string? Origin { get; set; }
     
-    [RdfProperty(Description = "Is the adjunct currently being ingested?",
+    [RdfProperty(Description = "he reason why this adjunct exists.",
```
donaldgray: Missing letter: `"he reason why this adjunct exists."`

src/protagonist/API/Features/Adjuncts/Validation/HydraAdjunctValidator.cs
```diff
@@ -54,6 +54,11 @@ public HydraAdjunctValidator(IOptions<ApiSettings> apiSettings)
         RuleForEach(a => a.Language)
             .MaximumLength(maximumLength)
             .WithMessage($"All 'language' values must be {maximumLength} characters or less. e.g. ISO language codes, sub-codes or 'none'");
+        
+        RuleFor(a => a.Type)
+            .Must(t =>  t == "AnnotationPage")
+            .When(a => a.IIIFLink == IIIFLinkType.Annotations.GetDescription())
+            .WithMessage($"When the 'iiifLink' is '{IIIFLinkType.Annotations.GetDescription()}', the '@type' must be 'AnnotationPage'");
```
donaldgray: Tiny but should we drop `the` in the error messages, make it more concise and feels more inline with what we have elsewhere

```
When the 'iiifLink' is 'annotations', the '@type' must be 'AnnotationPage'
When 'iiifLink' is 'annotations', '@type' must be 'AnnotationPage'
```

donaldgray (approved):

PR #1119, opened 2026-02-27
Refresh OrchestrationImage info.json create

donaldgray:
Avoids issues where stale OrchestrationImage can be used to create info.json. We cache info.json for a long time, so created with stale data that can hang around. If info.json is missing it means the image new, or it has recently been updated and cleanup-handler has removed.

This looks like more changes than it actually is, summary:
* Add new `IOrchestratorAssetRepository : IAssetRepository`, which adds a `noCache` overload of `GetAsset()`
* `DapperAssetRepository` inherits from above, so had to move into Orchestrator (this should have been done a while ago).
* Extend `MemoryAssetTracker.RefreshCachedAsset` to call new overload with `true` (ie get a fresh asset)
* InfoJson* classes now take `AssetId` rather than `OrchestrationAsset` when building info.json.

### DIFF
--- a/src/protagonist/DLCS.Repository/Assets/AssetCachingHelper.cs
+++ b/src/protagonist/DLCS.Repository/Assets/AssetCachingHelper.cs
@@ -12,20 +12,13 @@ namespace DLCS.Repository.Assets;
 /// <summary>
 /// Helper for working with cached assets
 /// </summary>
-public class AssetCachingHelper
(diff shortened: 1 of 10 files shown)


griffri (approved):
LGTM - would be good to have the others' eyes on it too

JackLewis-digirati (approved):

PR #1124, opened 2026-03-04
Add support to orchestrator for inline annotations

### LINKED ISSUE #1113: Output `inlineAnnotation` type on generated Manifests
Output `inlineAnnotation` adjuncts in generated manifests (single item and NamedQuery).

#1074 was the initial implementation of adjunct rendering on Manifests. `inlineAnnotation` items are treated slightly differently as they are output into a "synthetic" AnnotationPage.

The should be built as:

```json
{
    "annotations": [
    {
        "id": "https://dlcs.example/adjunct-annotations/{asset-id}", // not dereferenceable
        "type": "AnnotationPage",
        "label": { "en": [ "Inline annotations" ] }, // hardcoded label
        "items": [ // There will be an "item" included per inlineAnnotation adjunct
          {
            "id": "https://dlcs.example/adjunct-annotations/{asset-id}/{adjunct-id}", // not dereferenceable
            "type": "Annotation",
            "motivation": "{{motivation}}",
            "body": {
              "id": "{{publicId}}",
              "type": "{{@type}}",
              "format": "{{format}}",
              "profile": "{{profile}}",
              "label": "{{label}}",
              "language": "{{language}}"
            },
            "target": "<canvas-id>"
          }
        ]
    }]
}
```

* [x] `inlineAnnotation` types are added to `"annotations"` property
* [x] `annotation` adjuncts may also be added to the same `"annotations"` property.
* [x] `inlineAnnotations` are added to a "synthetic" `AnnotationPage`
  * [x] `"label"` is `{ "en": [ "Inline annotations" ] }`
  * [x] `"id"` uses above format and is not dereferenceable
* [x] All `inlineAnnotations` for a single asset are included in the same "synthetic" `AnnotationPage`.
* [x] See above for construction. All `{{}}` properties come from Adjunct record

JackLewis-digirati:
Resolves #1113 

This PR adds inline annotation adjuncts to the output from orchestrator.  It will collect all `inlineAnnotations` into a single annotation and makes the assumption that if there is an `annotation.Items` property available, that this is a collection of `inlineAnnotations`

This will create output like the below on a manifest:

```json
"annotations": [
                {
                    "id": "https://localhost:5003/adjunct-annotations/26/1/test-image",
                    "type": "AnnotationPage",
                    "label": {
                        "en": [
                            "Inline annotations"
                        ]
                    },
                    "items": [
                        {
                            "id": "https://localhost:5003/adjunct-annotations/26/1/test-image/anotherAdjunct",
                            "type": "Annotation",
                            "motivation": "something",
                            "body": {
                                "id": "https://some-location.com/an-adjunct",
                                "type": "Image",
                                "label": {
                                    "label": [
                                        "value"
                                    ]
                                },
                                "format": "a-mediaType",
                                "language": [
                                    "en"
                                ]
                            }
                        },
                        {
                            "id": "https://localhost:5003/adjunct-annotations/26/1/test-image/anotherAdjunct2",
                            "type": "Annotation",
                            "motivation": "something",
                            "body": {
                                "id": "https://some-location.com/an-adjunct",
                                "type": "Image",
                                "label": {
                                    "label": [
                                        "value"
                                    ]
                                },
                                "format": "a-mediaType",
                                "language": [
                                    "en"
                                ]
                            }
                        }
                    ]
                }
            ]
```

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
+++ b/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
@@ -12,7 +12,6 @@ public static class AssetDeliveryChannels
     public const string File = "file";
     public const string None = "none";
     public const string Default = "default";
-    public const string Adjunct = "adjunct-annotations";
 
     /// <summary>
     /// All possible delivery channels
--- a/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
+++ b/src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
@@ -12,6 +12,7 @@ public static class AssetDeliveryChannels
     public const string File = "file";
     public const string None = "none";
     public const string Default = "default";
+    public const string Adjunct = "adjunct-annotations";
 
     /// <summary>
     /// All possible delivery channels
--- a/src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
@@ -815,7 +815,7 @@ public class ManifestHandlingTests : IClassFixture<ProtagonistAppFactory<Startup
             .WithTestAdjunct("seeAlso2", type: "Text", mediaType: "text/xml", iiifLinkType: IIIFLinkType.SeeAlso,
                 externalId: "https://other.example/2")
             .WithTestAdjunct("inlineAnnotation1", type: "AnnotationPage", mediaType: "text/xml", iiifLinkType: IIIFLinkType.InlineAnnotation,
-                externalId: "https://inline.example/1")
+                externalId: "https://inline.example/1", motivation: "something")
             .WithTestAdjunct("inlineAnnotation2", type: "AnnotationPage", mediaType: "text/xml", iiifLinkType: IIIFLinkType.InlineAnnotation,
                 externalId: "https://inline.example/2");
         
@@ -827,8 +827,8 @@ public class ManifestHandlingTests : IClassFixture<ProtagonistAppFactory<Startup
             new() { Id = $"http://localhost/adjunct-annotations/{id}", Label = new LanguageMap("en", "Inline annotations"), 
                 Items = 
                 [
-                    new GeneralAnnotation(null){Id = $"http://localhost/adjunct-annotations/{id}/inlineAnnotation1", Body = [new AdjunctOutput("AnnotationPage"){ Id = "https://inline.example/1", Format = "text/xml"}]},
-                    new GeneralAnnotation(null){Id = $"http://localhost/adjunct-annotations/{id}/inlineAnnotation2", Body = [new AdjunctOutput("AnnotationPage"){ Id = "https://inline.example/2", Format = "text/xml"}]}
+                    new GeneralAnnotation(null){Id = $"http://localhost/adjunct-annotations/{id}/inlineAnnotation1", Motivation = "something", Body = [new ExternalResource("AnnotationPage"){ Id = "https://inline.example/1", Format = "text/xml"}]},
+                    new GeneralAnnotation(null){Id = $"http://localhost/adjunct-annotations/{id}/inlineAnnotation2", Body = [new ExternalResource("AnnotationPage"){ Id = "https://inline.example/2", Format = "text/xml"}]}
                 ]}
         ];
 
--- a/src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
@@ -14,6 +14,7 @@ using IIIF.Serialisation;
 using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json.Linq;
 using Orchestrator.Infrastructure.IIIF;
+using Orchestrator.Infrastructure.IIIF.Manifests;
 using Orchestrator.Tests.Integration.Infrastructure;
 using Test.Helpers;
 using Test.Helpers.Data;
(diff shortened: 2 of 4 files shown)


src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
```diff
@@ -509,20 +509,73 @@ private void AddAdjunctsToCanvas(Canvas canvas, Asset asset)
...
+            Space = adjunct.AssetId.Space,
+            AssetPath = adjunct.AssetId.Asset,
+            RoutePrefix = AssetDeliveryChannels.Adjunct,
+            CustomerPathValue = adjunct.AssetId.Customer.ToString(),
         };
+        return assetPathGenerator.GetFullPathForRequest(transcodeRequest, BuilderUtils.UseNativeFormatForAssets, false);
     }
 }
+
+public class AdjunctOutput(string type) : ExternalResource(type), IPaintable;
```
donaldgray: Why the custom type? `ExternalResource` is the appropriate type to use for what we're adding.

src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
```diff
@@ -12,6 +12,7 @@ public static class AssetDeliveryChannels
     public const string File = "file";
     public const string None = "none";
     public const string Default = "default";
+    public const string Adjunct = "adjunct-annotations";
```
donaldgray: This doesn't seem right to sit here. This is adjuncts specific, nothing to do with assets or delivery-channels.

src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
```diff
@@ -509,20 +509,73 @@ private void AddAdjunctsToCanvas(Canvas canvas, Asset asset)
...
+            if (inlineAnnotationPage == null)
+            {
+                currentCanvas.Annotations ??= [];
+                currentCanvas.Annotations.Add(new AnnotationPage
+                {
+                    Id = GetPathForAdjunct(adjunct),
+                    Label = new LanguageMap("en", "Inline annotations"),
+                    Items =
+                    [
+                        new GeneralAnnotation(adjunct.Motivation)
```
donaldgray: Readability/maintenance - the construction of the `GeneralAnnotation` is duplicated in both branches of `if`, suggest instantiating it outside the of the `if` - single place to create it, means that we can't change 1 and forget about the other.

```cs
var generalAnnotation = new GeneralAnnotation(adjunct.Motivation){ ... }};

if (inlineAnnotationPage == null)
{
    ...
```

src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
```diff
@@ -509,20 +509,73 @@ private void AddAdjunctsToCanvas(Canvas canvas, Asset asset)
...
+    private string GetPathForAdjunct(Adjunct adjunct)
+    {
+        var transcodeRequest = new BasicPathElements
+        {
+            Space = adjunct.AssetId.Space,
+            AssetPath = adjunct.AssetId.Asset,
+            RoutePrefix = AssetDeliveryChannels.Adjunct,
+            CustomerPathValue = adjunct.AssetId.Customer.ToString(),
         };
+        return assetPathGenerator.GetFullPathForRequest(transcodeRequest, BuilderUtils.UseNativeFormatForAssets, false);
```
donaldgray: I'm not sure about using `assetPathGenerator` here. The issue is that the path format could differ based on the incoming hostname - I'm not sure whether we want to do that, my current thinking is that we probably don't want to. Although we will have a better idea once we've worked through how IIIF-Presentation will handle these.

Other `id` values in NamedQuery (Canvas, Annotations etc) use `GetCanvasId()` - either directly or as a base - which _always_ uses canonical path, the format of this never differs. This is then a guaranteed format that we can use when processing (e.g. https://github.com/dlcs/iiif-presentation/blob/af3c39a8a0d123c256d12d505ac61912b4a8d40f/src/IIIFPresentation/Repository/Paths/PathParser.cs#L22-L24) to identify AssetId, or in this case AdjunctId. We do rewrite the path of the actual content as that's where it can be fetched from on 'nice' paths - with CDN etc potentially in front - but the skeleton structure around it is often useful to have a standard format we can reliably parse.

src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
```diff
@@ -812,13 +813,23 @@ await dbFixture.DbContext.Images
...
         await dbFixture.DbContext.SaveChangesAsync();
 
         List<AnnotationPage> expectedAnnos =
         [
-            new() { Id = "https://mets.example/w3c", Label = new LanguageMap("en", "Line-level annos") }
+            new() { Id = "https://mets.example/w3c", Label = new LanguageMap("en", "Line-level annos") },
+            new() { Id = $"http://localhost/adjunct-annotations/{id}", Label = new LanguageMap("en", "Inline annotations"), 
+                Items = 
+                [
+                    new GeneralAnnotation(null){Id = $"http://localhost/adjunct-annotations/{id}/inlineAnnotation1", Body = [new AdjunctOutput("AnnotationPage"){ Id = "https://inline.example/1", Format = "text/xml"}]},
```
donaldgray: Small, both these are `new GeneralAnnotation(null)` - should we confirm `motivation` is set if specified?

src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
JackLewis-digirati: I think this had previously been something which didn't quite work with `ExternalResource`, but after some changes it does work - will remove the custom type

src/protagonist/Orchestrator.Tests/Integration/ManifestHandlingTests.cs
JackLewis-digirati: Will look to add something, but the `BuildManifest_InlineAnnotation` test does check for `motivation` being set correctly

src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
JackLewis-digirati: I can use `GetCanvasId`, but it does output in a different format than the one specified in #1113 i.e.: `http://localhost/adjunct-annotations/99/1/Get_V3ManifestWithAdjuncts_OutputsAdjunctsInCorrectLocation` vs `http://localhost/iiif-img/99/1/Get_V3ManifestWithAdjuncts_OutputsAdjunctsInCorrectLocation/canvas/c/1/adjunct-annotations` by default - is that  what it should be, or in a different format?

donaldgray: It should be the pattern in the ticket (`/adjunct-annotations/`).

Sorry, I wasn't the clearest - I mentioned `GetCanvasId()` just to point out how other paths are generated, rather than saying to use it directly. Internally that method does call `assetPathGenerator`, specifying param to use native path format always. 

On a re-look it might be fine to use what's there but always setting `useNativeFormat=true`?

```cs
// maybe this will do .,,
// but "transcodeRequest" is wrong!!
assetPathGenerator.GetFullPathForRequest(transcodeRequest, true, false);

// rather than this ...
assetPathGenerator.GetFullPathForRequest(transcodeRequest, BuilderUtils.UseNativeFormatForAssets, false);
```

JackLewis-digirati: sure, I can do this, though by the `transcodeRequest` being wrong, do you mean around how the `RoutePrefix` is in the incorrect location?

donaldgray: No, was just looking at the variable name (it's not a transcode, it's an adjunct)

src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
```diff
@@ -509,20 +510,63 @@ private void AddAdjunctsToCanvas(Canvas canvas, Asset asset)
...
+        {
+            var inlineAnnotationPage = currentCanvas.Annotations?.FirstOrDefault(a => !a.Items.IsNullOrEmpty());
+        
+            if (inlineAnnotationPage == null)
+            {
+                currentCanvas.Annotations ??= [];
+                currentCanvas.Annotations.Add(new AnnotationPage
+                {
+                   Id = GetPathForAdjunct(adjunct),
+                    Label = new LanguageMap("en", "Inline annotations"),
```
donaldgray: Tiny but indent is off here and line below

donaldgray (approved):

Issue #1098, opened 2026-01-28
Advertised adjunct paths can be rewritten


donaldgray:
#1079 allowed hosted assets to be output on generated Manifests.

This ticket is to use the ["pathTemplate"](https://github.com/dlcs/protagonist/blob/develop/docs/path-templates.md) logic to allow adjunct ids to be rewritten on Manifests.

This allows Protagonist to advertise alternative ids for items, dependant on the incoming hostname.

* `prefix` will be `adjuncts`
* `assetPath` should suffice for `<asset-id>/<adjunct-id>`

> [!WARNING]
> `assetPath` being `<asset-id>/<adjunct-id>` means that all custom paths must have `<asset-id>/<adjunct-id>` together, which I think is a safe limitation for now as adjunctId is only unique for an individual assetId. We can extend later if there is a requirement to do so.

For reference, the canonical path is:

* `/adjuncts/{customer}/{space}/{asset-id}/{adjunct-id}`
  * e.g. `https://dlcs.example/adjuncts/2/99/foo/bar`

Which still conforms to the format of `/{prefix}/{version}/{customer}/{space}/{assetPath}`, where
* `{prefix}` is `adjuncts`
* `{version}` isn't relevant so ignored
* `{customer}` is `2`
* `{space}` is `99`
* `{assetPath}` is `foo/bar` (`{assetId}/{adjunctId}`)

Example path-templates and output for adjunct `bar` for asset `2/99/foo`:
* `/{prefix}/{version}/{customer}/{space}/{assetPath}` => `/adjuncts/2/99/foo/bar` (default)
* `/{prefix}/{assetPath}` => `/adjuncts/foo/bar` (no customer or space - common in single Customer envs)
* `/supplementary/{assetPath}` => alternative "supplementary" prefix
-----
* [x] Manifests support rewritten path as default (ie regardless of incoming hostname they non-canonical format is output)
* [x] Manifest support rewritten paths based on incoming hostname
* [x] Rewrite rules support string [formatters](https://github.com/dlcs/protagonist/blob/develop/docs/path-templates.md#formatters)
* [x] Rewritten paths can include additional path elements (with exception of between Asset and Adjunct it, as per above warning)
* [x] Rewritten paths can include omitted path elements
* [x] Path template elements can be reordered

> [!NOTE]
> Make sure rewrite rules are working on dev environment.

> [!NOTE]
[...]

PR #1129, opened 2026-03-05
Initial RFC for stub assets

donaldgray:
## What does this change?

RFC for how we can store 'stub' assets that only serve as a container for adjuncts.

### DIFF
--- a/docs/rfcs/022-stub-assets.md
+++ b/docs/rfcs/022-stub-assets.md
@@ -1,6 +1,6 @@
 # Stub Assets and Adjuncts
 
-This RFC looks at how we can enable "stub" assets to store adjuncts and how we can carry out bulk adjunct operations for multiple assets. 
+This RFC looks at how we can enable "stub" assets to store adjuncts and how we can carry out bulk adjunct operations for multiple assets.
 
 Stub assets are assets that won't ever have any binary content themselves, they exist purely to store adjuncts.
 
@@ -22,7 +22,7 @@ We could leverage the `none` channel but what about `"origin"` and `"mediaType"`
 }
 ```
 
-We could require that the consumer pass some known placeholder values, similar to the "unobtainable" role from [ADR-0010](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md), or they are free to use these columns for their own means - they are effectively ignored.
+We could require that the consumer pass some known placeholder values, similar to the "unobtainable" role from [ADR-0010](https://github.com/dlcs/protagonist/blob/develop/docs/adr/0010-replace-maxunauthorised.md), or they are free to use these properties for their own means - they are effectively ignored.
 
 ```json
 {
@@ -56,17 +56,15 @@ When creating spaces, customers will get the next available int identifier, or t
 >
 > If we supported JWT tokens and claims we could specify what consumers can use this space.
 
-### Specific Logic
+### Validation/Logic Changes
 
-If we use space 0 for _all_ stub assets we can make the validation logic work slightly differently - all fields will act as they do with other spaces with the exception that:
+If we use space 0 for _all_ stub assets we can make updates to the validation logic to help. 
 
-* `"origin"` and `"mediaType"` are optional. If not supplied they gain placeholder values.
-* `"deliveryChannel"` must be `none`, any other value is rejected.
+In space 0, all fields will act as they do with other spaces with the exception that `"deliveryChannel"` must be `none`, any other value is rejected.
 
-> [!NOTE]
-> Do we want to do this? I'm not sure if different validation logic is better than the consumer passing random values? Restricting use of none makes sense.
-
-The API can shortcut assets and auto-finish them, no need to notify Engine at all.
+When using the `none` channel:
+* `"origin"` and `"mediaType"` are optional. If not supplied they gain placeholder values (this will be considerably simpler to implement over changing logic to check for existence).
+* The API can shortcut assets and auto-finish them, no need to notify Engine at all.
 
 ### Routing / Paths
 
@@ -121,6 +119,15 @@ For bulk reading of adjuncts we should support an `include` parameter for assetQ
 * Con: The returned Image Hydra model has an existing "adjuncts" property, which is normally a URI but this would involve it being a collection of adjuncts. This is valid for json+ld, but may present some serialisation challenges.
 * Con: It won't be possible to query for adjuncts only.
 
+An alternative representation of a separate `?include` parameter would be to encode it within the query parameter itself, e.g.
+
+```json
+{
+   "manifest": [ "blah" ],
+   "includeAdjuncts": true
+}
+```
+
 ### Bulk Writing
 
 `POST /customers/{c}/spaces/{s}/images/{i}/adjuncts` allows bulk adding of adjuncts to a single asset. We need an alternative endpoint that would allow adding adjuncts to multiple assets at once. This is akin to how a batch can be created for images in multiple different spaces.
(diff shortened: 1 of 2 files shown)


docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Specific Logic
+
+If we use space 0 for _all_ stub assets we can make the validation logic work slightly differently - all fields will act as they do with other spaces with the exception that:
+
+* `"origin"` and `"mediaType"` are optional. If not supplied they gain placeholder values.
+* `"deliveryChannel"` must be `none`, any other value is rejected.
+
+> [!NOTE]
+> Do we want to do this? I'm not sure if different validation logic is better than the consumer passing random values? Restricting use of none makes sense.
```
tomcrane: Maybe the requirement is actually the same for all spaces:

`origin` and `mediaType` are REQUIRED **_unless_** the `deliveryChannel` is "none"

This would allow you to reserve asset slots for later use.

The space 0 constraint is then that you can't ever change the deliveryChannel from `none` for assets in that space. Maybe that just moves the ick somewhere else though.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+
+The API can shortcut assets and auto-finish them, no need to notify Engine at all.
+
+### Routing / Paths
+
+Would we want to output space 0 on output paths? By default we will but it could be useful to make this a configurable option. `{space}` is the path template for space, we could support a `{spaceNot0}` or `{positiveSpace}` to only output the space path component if it is > 0.
```
tomcrane, line 71: seems easier to just output /0/

But adjuncts will presumably be subject to rewrite rules as well

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+
+> [!NOTE]
+> Is this of use? Or would we need to see what path rewrite requirements would be before we determine what to do here?
+
+An alternative might be to have different values based on whether the space is 0 or not. A common customer configuration is to have a Customer/Space per environment, in these instances we would likely want to differentiate between space 0 and not. E.g.
+
+* Canonical image: `https://example.dlcs/iiif-img/2/10/asset`
+* Custom image: `https://customer.host/images/asset`
+* Canonical asset adjunct: `https://example.dlcs/adjuncts/2/10/asset/mets.xml`
```
tomcrane, line 80: I wonder if we need an extra path element to avoid collisions. e.g., in the public URL:

`https://example.dlcs/adjuncts/2/10/asset/adjuncts/mets.xml`

`adjuncts` is reserved, and I can have whatever I like after that.

Dumb scenario: I have an adjunct and I really want to call it `info.json` - pushing it one path element along under /adjuncts/info.json avoids the obvious collision. We could just ban `info.json` as an adjunct `id` but we might want other .../asset/xxx reserved slugs for whatever reasons, and easier to just reserve /adjuncts/ now.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+These are useful, but in scenarios like IIIF-Presentation where we may want to get all adjuncts for a number of assets, or upsert adjuncts to a number of different assets it would be useful to have alternative endpoints for these.
+
+### Bulk Reading
+
+[AssetQuery](https://dlcs.github.io/public-docs/api-doc/asset-queries/) syntax allows for querying multiple assets based on specified criteria. This is supported on:
+
+* `/customers/{customer}/allImages`
+* `/customers/{customer}/spaces/{space}/images`
+
+For bulk reading of adjuncts we should support an `include` parameter for assetQuery syntax, rather than introducing new endpoints. This would include inline adjuncts with each asset. E.g. `/customers/99/allImages?q={{\"manifests\":[\"whatever\"]}}&include=adjuncts`.
```
tomcrane, line 115: If the q param is always an {{ object }} then maybe it can have:

```
{
   "manifest": [ "blah" ],
   "includeAdjuncts": true
}
```

Is that neater? Allows a query object to passed around. Not sure which is better.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+* Raise notification on completion
+* GET all batches `/customers/{c}/adjunctQueue/batches` (equivalent of `/customers/{c}/queue/batches`)
+* GET batch details `/customers/{c}/adjunctQueue/batches/{id}` (equivalent of `/customers/{c}/queue/batches/{id}`)
+* GET batch adjuncts `/customers/{c}/adjunctQueue/batches/{id}/adjuncts` (equivalent of `/customers/{c}/queue/batches/{id}/assets`)
+* POST add batch `/customers/{c}/adjunctQueue` (equivalent of `/customers/{c}/queue`)
+
+I don't think we would need all the functionality of batches immediately - no priority batches or recent/active batch endpoints.
+
+> [!NOTE]
+> What `@type` would this be? `vocab:Batch` has `"*Images"` properties that are not relevant to adjuncts.
```
tomcrane, line 154: OTOH this is already "wrong" in the sense that they aren't just images... although they are `vocab:Image` but that's for another day.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+>
+> While there are a lot of shared properties, it is a different object. `vocab:AdjunctBatch` perhaps?
```
tomcrane, line 156: Perhaps. I'm not totally averse to reusing `Batch` for this, even with Images; with the idea that it's really an alias for "members" or "resources" or something; the things being batched.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Bulk Deleting
+
+Support bulk delete operations, similar to current POST `/customers/{customerId}/deleteImages`. Suggested endpoint is `/customers/{customerId}/deleteAdjuncts`, similar to the asset equivalent this will accept a list of `id` values.
+
+## `manifests` Property
+
+Rename `manifests` column/property to `scopes`. `manifests` is too prescriptive for use, it can also be Collections. `scopes` keeps options for reuse open in the future. To fully make this change there will be changes to NQs to support `scopes` parameter, PATCH `/customers/{customerId}/allImages`, update `assetQuery` syntax etc. The `scopes` parameter is an internal use list of associated references.
```
tomcrane, line 164: Need to reflect this in the other PR

tomcrane (approved):

docs/rfcs/022-stub-assets.md
donaldgray: Made updates:
* Space 0 assets must be `none`
* `origin` and `mediaType` are optional if using `none`

docs/rfcs/022-stub-assets.md
donaldgray: As discussed on Slack, leaving this for now. See `[!WARNING]` on https://github.com/dlcs/protagonist/issues/1098, adding a `/adjuncts/` slug would break the rule set out there. Would also differ from canonical path as outlined in docs.

docs/rfcs/022-stub-assets.md
donaldgray: Added as option.

docs/rfcs/022-stub-assets.md
```diff
@@ -0,0 +1,195 @@
...
+Not all batch functionality needs to be implemented immediately - no priority batches or recent/active batch endpoints.
+
+> [!NOTE]
+> What `@type` would this be? `vocab:Batch` has `"*Images"` properties that are not relevant to adjuncts.
+>
+> While there are a lot of shared properties, it is a different object. `vocab:AdjunctBatch` perhaps?
+
+### Bulk Deleting
+
+Support bulk delete operations, similar to current POST `/customers/{customerId}/deleteImages`. Suggested endpoint is `/customers/{customerId}/deleteAdjuncts`, similar to the asset equivalent this will accept a list of `id` values.
```
JackLewis-digirati, line 160: is there potentially a need to have an endpoint that will let you delete both assets and adjuncts in the same call? Could see it potentially being useful in a manifest update scenario (if we decide to support actual deletion via a manifest)

donaldgray: We could revisit if `/deleteImages` and `/deleteAdjuncts` are too cumbersome to work with. Initial thought is that deleting different types of resources in the same API call feels awkward.

JackLewis-digirati (approved):

PR #1130, opened 2026-03-06
Handle very small maxWidth values and "sizes" in info.json

### LINKED ISSUE #1123: `/iiif-img/` `info.json` shows thumbnail sizes available even when `"maxWidth"` is too low
If the value of `"maxWidth"` is too low (e.g, `10` - a value smaller than the smallest thumbnail size) for an image, requesting its `/thumbs/` `info.json` returns 404 as there's nothing to serve.

However, requesting its `/thumbs/` `info.json` returns various sizes that shouldn't have been generated.

[Example](https://github.com/user-attachments/files/25738039/test_b.json)

donaldgray:
## What does this change?

Fixes #1123 . See final comment, logic is:

* Prevent creation of image with `maxWidth` < 256
* If building info.json and no thumbnails found, use those from image-server but filter out any that exceed `maxWidth`.

### DIFF
--- a/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
+++ b/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
@@ -80,17 +80,19 @@ public class HydraImageValidatorTests
     [Fact]
     public void DeliveryChannel_ValidationError_DeliveryChannelMissingChannel()
     {
-        var model = new Image { DeliveryChannels =
-        [
-            new DeliveryChannel
-            {
-                Policy = "none"
-            },
-            new DeliveryChannel
-            {
-                Channel = "file"
-            }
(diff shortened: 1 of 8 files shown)


src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
```diff
@@ -79,85 +80,90 @@ public void Created_Provided()
...
                 Channel = "none"
             },
-            new DeliveryChannel()
+            new DeliveryChannel
             {
                 Channel = "file"
             }
-        } };
+        ]
+        };
```
JackLewis-digirati, line 117: tiny - formatting of the `]` and `}`?  - There's a few of these in this class

JackLewis-digirati (approved):

PR #1138, opened 2026-03-17
Add ability to delete hosted adjuncts

### LINKED ISSUE #1099: Cleanup hosted adjuncts
A DELETE endpoint is available for adjuncts, implemented in #1073. This currently just removes the DB record, which is all that was required for #1073 as it was for adjuncts with an `externalId`. Now that we can have hosted adjuncts, we may need to clean up any deleted adjuncts, or those that are no longer required (ie the Asset was deleted, or adjunct was changed in such a way that the hosted asset is no longer needed).

This ticket is to ensure that when an adjunct or asset is deleted, or  any hosted adjuncts are removed from DLCS storage by the cleanup-handler. This should happen when:

* [x] An adjunct is deleted (`DELETE /customers/{customer}/spaces/{space}/images/{image}/adjuncts/{adjunct}`)
* [x] An individual asset is deleted (`DELETE /customers/{customer}/spaces/{space}/images/{image}`)
* [x] Assets are bulk deleted (`POST /customers/{customer}/deleteImages`)
* [x] API raises an SNS notification for create/update/delete operations on adjuncts. If SNS raising fails, record somewhere for manual intervention.
* [x] Asset is changed from "Hosted" to "External" via PUT.

> [!NOTE]
> Cleanup-handler may already be deleting stored adjuncts when an asset is deleted, assuming adjuncts are stored in storage bucket with asset-id as prefix

JackLewis-digirati:
## What does this change?

This PR allows hosted adjuncts to be removed from S3 when either deleted OR changed from a hosted to external adjunct by raising notfications in SNs.  This will raise notifications for all of these operations, but only a subset will cause the cleanup handler to run

## Infrastructure Changes

> [!NOTE]
> This PR requires infrastructure changes.
>
> - **Required:**
>   - SQS queues for adjunct updating and deletion
>   - SNS topic for adjunct modification

> [!NOTE]
> There has been a change to SNS messages to reduce duplication in messaging between adjunct and asset, but has the knock-on impact that messages produced before the rollout will be in a different format to afterwards.  This should be transparent to the program as handling has been introduced to support both the old and new message formats, but this should be removed in a future version. 

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | CleanupHandler | `AWS:SQS:AdjunctDeleteNotificationQueueName` | N | Delete notification queue name | `null` |
> | CleanupHandler | `AWS:SQS:AdjunctUpdateNotificationQueueName` | N | Update notification queue name | `null` |
> | API | `AWS:SNS:AdjunctModifiedNotificationTopicArn` | N | Update notification topic | `null` |
>
> The application *will* start without these values present, but not having them means adjuncts will not be cleaned up properly


src/protagonist/DLCS.Model/Messaging/UpdatedNotificationRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿using DLCS.Model.Assets;
+using DLCS.Model.PathElements;
+
+namespace DLCS.Model.Messaging;
+
+public class UpdatedNotificationRequest<T> where T : IDeliverable
```
donaldgray: I think this could be more clearly named, maybe `DeliverableUpdatedNotification` to show that it is for an `IDeliverable`, also dropping `Request` as it's an actual notification, rather than a request to create a notification?

Same with Created and Deleted

src/protagonist/DLCS.Model/Messaging/Adjunct/IngestAdjunctRequest.cs
```diff
@@ -1,9 +1,8 @@
 ﻿using System;
 using System.Diagnostics;
 using DLCS.Core.Types;
-using DLCS.Model.Assets;
 
-namespace DLCS.Model.Messaging;
+namespace DLCS.Model.Messaging.Adjunct;
```
donaldgray: Small - why move? If this is moved to a sub-namespace, should the others be moved too for consistency? (e.g. should `IngestAssetRequest` be moved to `.Asset`?)

src/protagonist/DLCS.Model/Assets/Adjunct.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace DLCS.Model.Assets;
 
-public class Adjunct : IDeliverable
+public class Adjunct : ICloneable, IDeliverable
```
donaldgray: Should `IDeliverable : ICloneable` ?

Both `Asset` and `Adjunct` are `IDeliverable`. IIRC cloning is required for raising notifications, so that we can see the "before" and "after". Given that `IDeliverable` is the generic constraint when raising notifications any future `IDeliverable` objects could be used in notifications so if we add `ICloneable` to `IDeliverable` it'd force any implementors to add clone implementation too.

src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using DLCS.Model.Customers;
+using DLCS.Model.Messaging.Adjunct;
```
donaldgray:
```suggestion
```

src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
```diff
@@ -26,12 +27,20 @@ public Task<bool> PublishToCustomerCreatedTopic(CustomerCreatedNotification mess
...
     public Task<bool> PublishToBatchCompletedTopic(BatchCompletedNotification message,
         CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Asynchronously publishes a message to an Adjunct Modified SNS topic
+    /// </summary>
+    /// <param name="messages">A collection of notifications to send</param>
+    /// <param name="cancellationToken">Current cancellation token</param>
+    /// <returns>Boolean representing the overall success/failure status of all requests</returns>
+    Task<bool> PublishToAdjunctModifiedTopic(IReadOnlyList<DeliverableModifiedNotification> messages, CancellationToken cancellationToken);
```
donaldgray: Could we have a single `PublishModified` method? The parameter provided to both `PublishToAssetModifiedTopic` and `PublishToAdjunctModifiedTopic` is `IReadOnlyList<DeliverableModifiedNotification>`. 

Unlikely but it's possible I could pass a list of adjunct containing `DeliverableModifiedNotification` to the `PublishToAssetModifiedTopic` method. Having a single method that takes the `DeliverableModifiedNotification` and works out where to send them feels safer? Any consumer can effectively just say "send this" and the `ITopicPublisher` impl can work out where to go.

src/protagonist/DLCS.AWS/Settings/SQSSettings.cs
```diff
@@ -39,6 +39,16 @@ public class SQSSettings
     /// </summary>
     public string? DeleteNotificationQueueName { get; set; }
     
+    /// <summary>
+    /// Name of queue for handling notifications that adjuncts have been deleted
+    /// </summary>
+    public string? AdjunctDeleteNotificationQueueName { get; set; }
+    
+    /// <summary>
+    /// Name of queue for handling notifications that adjuncts have been deleted
```
donaldgray:
```suggestion
    /// Name of queue for handling notifications that adjuncts have been updated
```

src/protagonist/DLCS.AWS/SNS/TopicPublisher.cs
```diff
@@ -87,13 +76,39 @@ public async Task<bool> PublishToBatchCompletedTopic(BatchCompletedNotification
...
+        if (messages.Count == 1)
+        {
+            var singleMessage = messages[0];
+            return await PublishToDeliverableModifiedTopic(singleMessage, topicArn, cancellationToken);
+        }
+
+        const int maxSnsBatchSize = 5;
+        var allBatchSuccess = true;
+        var batchIdPrefix = Guid.NewGuid();
+        logger.LogDebug("Publishing SNS batch {BatchPrefix} containing {ItemCount} items", batchIdPrefix,
```
donaldgray: Small - is it worth logging part of the topicArn? Not sure how easy it is to parse just then name or something. Only thinking is that previously this always sent to 1 single Topic but now it's 1 of 2, so if there was an issue it may be of use to know which one.

It might be unnecessary though if it'd be clear from correlation-id or something where it's going.

src/protagonist/CleanupHandlerTests/CleanupHandlerTests.csproj
```diff
@@ -25,6 +25,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\API.Tests\API.Tests.csproj" />
```
donaldgray: This doesn't seem right?

src/protagonist/API/Features/Adjuncts/Requests/CreateOrUpdateAdjunct.cs
```diff
@@ -135,6 +141,12 @@ public async Task<ModifyEntityResult<Adjunct>> Handle(CreateOrUpdateAdjunct requ
                     WriteResult.NotFound);
             }
         }
+        
+        var adjunctModificationRecord = existingAdjunct == null
```
donaldgray: Can this use the same logic as `return ModifyEntityResult<Adjunct>.Success()` call for working out whether it's a create or update?

```suggestion
        var adjunctModificationRecord = isCreate
```

src/protagonist/API/Features/Adjuncts/Requests/DeleteAdjunct.cs
```diff
@@ -1,35 +1,60 @@
...
         return new ResultMessage<DeleteResult>(string.Empty, DeleteResult.Deleted);
     }
+    
+    private async Task RaiseNotification(DeleteAdjunct request, Adjunct deletedAdjunct,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            var deleted = NotificationRecord<Adjunct>.Delete(deletedAdjunct, request.DeleteFrom); 
+            logger.LogDebug("Sending delete asset notification for {AssetId}", request.AssetId);
```
donaldgray:
```suggestion
            logger.LogDebug("Sending delete adjunct notification for {AssetId}/{AdjunctId}", request.AssetId, request.Id);
```

src/protagonist/API/Features/Adjuncts/Requests/DeleteAdjunct.cs
```diff
@@ -1,35 +1,60 @@
...
+            await adjunctNotificationSender.SendAdjunctModifiedMessage(deleted, cancellationToken);
+        }
+        catch (Exception ex)
+        {
+            // Don't return error because notification failed
+            logger.LogWarning(ex, "Error raising delete notification for {AssetId}", request.AssetId);
```
donaldgray:
```suggestion
            logger.LogWarning(ex, "Error raising delete notification for {AssetId}/{AdjunctId}", request.AssetId, request.Id;
```

src/protagonist/API/Infrastructure/Messaging/General/ModificationSender.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace API.Infrastructure.Messaging.General;
+
+public class ModificationSender(
+    ITopicPublisher topicPublisher, 
+    IPathCustomerRepository customerPathRepository,
+    ILogger<ModificationSender> logger)
+{
+    public async Task SendAdjunctModifiedMessage<T>(IReadOnlyCollection<NotificationRecord<T>> notifications,
```
donaldgray:
```suggestion
    public async Task SendModifiedMessage<T>(IReadOnlyCollection<NotificationRecord<T>> notifications, 
```

src/protagonist/API/Infrastructure/Messaging/General/NotificationRecord.cs
```diff
@@ -2,38 +2,36 @@
...
 using DLCS.Model.Messaging;
 
-namespace API.Infrastructure.Messaging;
+namespace API.Infrastructure.Messaging.General;
 
-/// <summary>
-/// Represents a change to a single asset - the relevant status before/after change and the change type
-/// </summary>
-public class AssetModificationRecord
+public class NotificationRecord<T> where T : class, IDeliverable
```
donaldgray, line 11: Maintain the XML comments (with changes)?

src/protagonist/API/appsettings-Development-Example.json
```diff
@@ -23,7 +23,8 @@
       "AdjunctQueueName": "dlcs-adjunct"
     },
     "SNS": {
-      "AssetModifiedNotificationTopicArn": "arn:aws:sns:eu-west-1:{AWS account}:dlcsspinup-asset-modified-notifications",
+      "AssetModifiedNotificationTopicArn": "arn:aws:sns:eu-west-1:{AWS account}:{prefix}-asset-modified-notifications",
+      "AdjunctModifiedNotificationTopicArn": "arn:aws:sns:eu-west-1:{AWS account}:{prefix}-adjunct-modified-notifications",
```
donaldgray: Tiny, consistency with below:
```suggestion
      "AssetModifiedNotificationTopicArn": "arn:aws:sns:{region}:{account}:{prefix}-asset-modified-notifications",
      "AdjunctModifiedNotificationTopicArn": "arn:aws:sns:{region}:{account}:{prefix}-adjunct-modified-notifications",
```

src/protagonist/CleanupHandler/Adjunct/AdjunctDeletedHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            return true;
+        }
+
+        var adjunct = request.Deliverable;
+        
+        // if the item exists in the db, assume the adjunct has been reingested after delete
+        if (dbContext.Adjuncts.Any(a => a.Id == adjunct.Id && a.AssetId == adjunct.AssetId))
+        {
+            logger.LogInformation("Adjunct {Adjunct} can be found in the database, so will not be deleted",
```
donaldgray: Include AssetId, Adjunct id alone isn't unique.

src/protagonist/API/Infrastructure/PathHelper.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using DLCS.Model.PathElements;
+
+namespace API.Infrastructure;
+
+public static class PathHelper
+{
+    private static Dictionary<int, CustomerPathElement> customerPathElements = new();
+    
+    public static async Task<CustomerPathElement> GetCustomerPathElement(
```
donaldgray: I'm not sure about this, particularly the `private static Dictionary<int, CustomerPathElement> customerPathElements = new();`. This is effectively creating an `int: CustomerPathElement` lookup that's cached for the lifecycle of the API.

Originally this was a method in `AssetNotificationSender`, which is a scoped dependecy. The `customerPathElements` property was a readonly prop so was scoped to the current request. It's really unlikely that in practical terms it would cause an issue but if a customer name was updated it would never reflect without restarting API service entirely. 

The expensive lookup of hitting DB to fetch the id:name values is already handled within `customerPathRepository` so I think having this as a request-scoped lookup is much better.

src/protagonist/CleanupHandler/Adjunct/AdjunctDeletedHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+public class AdjunctDeletedHandler(
+    IAdjunctBucketOperations adjunctBucketOperations,
+    DlcsContext dbContext,
+    IOptions<CleanupHandlerSettings> handlerSettings,
+    ILogger<AdjunctDeletedHandler> logger)
+    : IMessageHandler
+{
+    private readonly CleanupHandlerSettings settings = handlerSettings.Value;
+
+    public async Task<bool> HandleMessage(QueueMessage message, CancellationToken cancellationToken = default)
```
donaldgray, line 20: Set correlation id for tracking?

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
```diff
@@ -0,0 +1,54 @@
+﻿using CleanupHandler.Infrastructure;
+using DLCS.AWS.SQS;
+using DLCS.Repository;
+using DLCS.Web.Logging;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace CleanupHandler.Adjunct;
```
donaldgray, line 10: If there is an `.Adjunct` namespace, should there be a `.Asset`?

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        
+        using (LogContextHelpers.SetCorrelationId(message.MessageId))
+        {
+            var adjunctBefore = request.DeliverableBeforeUpdate!;
+
+            var adjunctAfter = await dlcsContext.Adjuncts.FirstOrDefaultAsync(a => a.Id == request.DeliverableAfterUpdate!.Id);
+
+            if (adjunctAfter == null)
+            {
+                logger.LogInformation("Adjunct {Id} was not found in the database for use in after calculation",
```
donaldgray: As with delete, need AssetId in all logging messages

src/protagonist/CleanupHandler/Adjunct/AdjunctDeletedHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        var request = MessageParser.TryParseDeleteMessage<DLCS.Model.Assets.Adjunct>(message, logger);
+        if (request == null) return false;
+
+        // This means it's not a hosted asset, so nothing to do
```
donaldgray: Should we log that?

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    public async Task<bool> HandleMessage(QueueMessage message, CancellationToken cancellationToken = default)
+    {
+        var request = MessageParser.TryParseUpdatedMessage<DLCS.Model.Assets.Adjunct>(message, logger);
+        if (request == null) return false;
+        
+        using (LogContextHelpers.SetCorrelationId(message.MessageId))
+        {
+            var adjunctBefore = request.DeliverableBeforeUpdate!;
+
+            var adjunctAfter = await dlcsContext.Adjuncts.FirstOrDefaultAsync(a => a.Id == request.DeliverableAfterUpdate!.Id);
```
donaldgray: > [!CAUTION]
> This could pick up another adjunct with the same id associated with a different asset

I wonder if we need a helper to avoid this, `.GetAdjunct(assetId, adjunctId)` or something.

Can also pass CancellationToken to this.

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+            
+            await adjunctBucketOperations.DeleteFromOriginBucket(adjunctBefore, settings);
+
+            return true;
+        }
+    }
+
+    // We only cleanup when the adjunct has moved from hosted to unhosted
+    private bool NoCleanupRequired(DLCS.Model.Assets.Adjunct adjunctAfter,
+        DLCS.Model.Assets.Adjunct adjunctBefore) => adjunctBefore.Origin == null && adjunctAfter.Origin != null && adjunctAfter.ExternalId == null;
```
donaldgray: Small, the line wrapping on this seems odd, difficult to read with first param on 1 line and 2nd param and the expression body on the next.

```suggestion
    private static bool NoCleanupRequired(DLCS.Model.Assets.Adjunct adjunctAfter, DLCS.Model.Assets.Adjunct adjunctBefore) 
        => adjunctBefore.Origin == null && adjunctAfter.Origin != null && adjunctAfter.ExternalId == null;
```

src/protagonist/CleanupHandler/Infrastructure/DeliverableQueueType.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace CleanupHandler.Infrastructure;
+
+// These are the same, but this avoids issues when the services are registered
```
donaldgray: Alternative is to have 1 enum with the different types?

```cs
public enum CleanupMessageType
{
   AssetDeleted,
   AssetUpdated,
   AdjunctDeleted,
   AdjunctUpdated
}
```

IMO this is a little cleaner and more explanatory (_"here the types of the messages that are handled"_). It's used to wire up different listeners in the listener manager. It would also mean that you only need 1 `QueueHandlerResolver` registered in DI container. Then, if we wanted other items in here, say `DeletedSpace`, you'd only need to append to 1 enum and extend that 1 handler resolver rather than add a new `public enum SpaceQueueType`

src/protagonist/DLCS.Model/Assets/IDeliverable.cs
```diff
@@ -24,4 +24,9 @@ public interface IDeliverable : IOriginItem
     /// Returns <see cref="AssetId"/> that is or is parent to this deliverable
     /// </summary>
     AssetId GetAssetId();
+
+    /// <summary>
+    /// Returns the id of the deliverable
+    /// </summary>
+    string Identifier();
```
donaldgray: This can be removed. `IDeliverable : IOriginItem` and `IOriginItem` already has `string Identifier();`

src/protagonist/API/Infrastructure/Messaging/General/ModificationSender.cs
```diff
@@ -0,0 +1,112 @@
...
+            DeliverableAfterUpdate = deliverableAfter, 
+            CustomerPathElement = customerPathElement
+        };
+
+        return JsonSerializer.Serialize(request, serializerOptions);
+    }
+
+    private async Task<string>GetSerialisedCreatedNotification<T>(T deliverableAfter, JsonSerializerOptions serializerOptions) where T : IDeliverable
+    {
+
```
donaldgray, line 105:
```suggestion
```

src/protagonist/API/Infrastructure/Messaging/General/ModificationSender.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            DeliverableBeforeUpdate = deliverableBefore,
+            DeliverableAfterUpdate = deliverableAfter, 
+            CustomerPathElement = customerPathElement
+        };
+
+        return JsonSerializer.Serialize(request, serializerOptions);
+    }
+
+    private async Task<string>GetSerialisedCreatedNotification<T>(T deliverableAfter, JsonSerializerOptions serializerOptions) where T : IDeliverable
```
donaldgray:
```suggestion
    private async Task<string> GetSerialisedCreatedNotification<T>(T deliverableAfter, JsonSerializerOptions serializerOptions) where T : IDeliverable
```

src/protagonist/API/Infrastructure/Messaging/General/ModificationSender.cs
```diff
@@ -0,0 +1,112 @@
...
+using DLCS.Core.Collections;
+using DLCS.Core.Strings;
+using DLCS.Model.Assets;
+using DLCS.Model.Messaging;
+using DLCS.Model.PathElements;
+using Microsoft.Extensions.Logging;
+
+namespace API.Infrastructure.Messaging.General;
+
+public class ModificationSender(
```
donaldgray, line 16: Comment around what this does?

src/protagonist/API/Infrastructure/Messaging/Asset/AssetNotificationSender.cs
```diff
@@ -0,0 +1,64 @@
...
+using API.Infrastructure.Messaging.General;
+using DLCS.Core.Collections;
+using DLCS.Model.Assets;
+
+namespace API.Infrastructure.Messaging.Asset;
+
+/// <summary>
+/// Class that handles raising notifications for modifications made to assets (Create/Update/Delete)
+/// </summary>
+public class AssetNotificationSender(
```
donaldgray: Thought as discussed - can this be shared with `AdjunctNotificationSender`, so that there is a single sender?

The biggest difference looks to be the serialiser settings but the exclusions are by type so I think a single unified dictionary would work - e.g. the below handles both `Asset` and `ImageDeliveryChannel` so adding `Adjunct` to that should handle it properly.

donaldgray (changes requested):

donaldgray:
Hit review submit without adding comment:

Overall functionally looks like it's pretty much there, most comments are small/tidy with a couple of exceptions (static dict and missing AssetId filter).

Also noted Asset and Adjunct cleanup are implemented slightly inconsistently. I'm not sure if it's worth changing so just flagging:
* Assets do all db work via a repo, Adjuncts to it directly on dbContext in handler
* Assets do all S3 work in handler, Adjuncts to it via separate interface

> This should be transparent to the program as handling has been introduced to support both the old and new message formats, but this should be removed in a future version.

Raise a ticket to do this so that we don't lose sight.

src/protagonist/DLCS.Model/Messaging/Adjunct/IngestAdjunctRequest.cs
JackLewis-digirati: I moved them as it felt like it could get messy in the future if more of these were added and I was making changes, but given these have been rolled back, it's probably best to change it back - will do this

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
donaldgray: Also occurred to me - this should be `SingleOrDefault`, not `FirstOrDefault`. The latter could return multiple but soldier on, the former is what we want. If there are multiple matches, that's an error condition.

src/protagonist/API/Infrastructure/Messaging/General/ModificationSender.cs
```diff
@@ -0,0 +1,125 @@
...
+                    { "messageType", notification.ChangeType.ToString() }
+                };
+
+                changes.Add(new DeliverableModifiedNotification(serialisedNotification, attributes));
+            }
+        }
+
+        var typeParameterType = typeof(T);
+        
+        switch (typeParameterType) {
```
donaldgray: Small/opinion, this isn't particularly easy on the eye. I think it's the `case not null` that feels odd.

Maybe a switch expression?
```cs
typeParameterType switch
{
    _ when typeParameterType == typeof(Adjunct) => await topicPublisher.PublishToDeliverableModifiedTopic(changes, DeliverableTopicType.Adjunct, cancellationToken);,
    _ when typeParameterType == typeof(Asset) => await topicPublisher.PublishToDeliverableModifiedTopic(changes, DeliverableTopicType.Asset, cancellationToken),
    _ => throw new InvalidOperationException($"Deliverable type not supported - {typeParameterType}")
};
```

or use an expression to get the `DeliverableTopicType` only then execute based on that after?

```cs
var topicType = typeParameterType switch
{
    _ when typeParameterType == typeof(Adjunct) => DeliverableTopicType.Adjunct,
    _ when typeParameterType == typeof(Asset) => DeliverableTopicType.Asset,
    _ => throw new InvalidOperationException($"Deliverable type not supported - {typeParameterType}")
};

await topicPublisher.PublishToDeliverableModifiedTopic(changes, topicType, cancellationToken)
```

src/protagonist/API.Tests/Infrastructure/Messaging/DeliverableNotificationSenderTests.cs
```diff
@@ -13,104 +12,106 @@
 using Microsoft.Extensions.Logging.Abstractions;
 using System.Text.Json;
 using System.Text.Json.Nodes;
+using API.Infrastructure;
+using API.Infrastructure.Messaging.General;
 
 namespace API.Tests.Infrastructure.Messaging;
 
-public class AssetNotificationSenderTests
+public class DeliverableNotificationSenderTests
```
donaldgray, line 20: Rename file to match class

src/protagonist/API/Infrastructure/DeliverableNotificationSender.cs
```diff
@@ -0,0 +1,68 @@
...
+        ReferenceHandler = ReferenceHandler.IgnoreCycles,
+        TypeInfoResolver = new DefaultJsonTypeInfoResolver
+        {
+            Modifiers = { AssetSerialiserContractModifier }
+        }
+    };
+
+    private static void AssetSerialiserContractModifier(JsonTypeInfo typeInfo)
+    {
+        // Collection of properties to ignore when serialising Asset object, by containing type
```
donaldgray:
```suggestion
        // Collection of properties to ignore when serialising objects, by containing type
```

src/protagonist/API/Infrastructure/IDeliverableNotificationSender.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using API.Infrastructure.Messaging.General;
+using DLCS.Model.Assets;
+
+namespace API.Infrastructure;
+
+public interface IDeliverableNotificationSender
+{
+    /// <summary>
+    /// Broadcast a change to the status of an Asset, for any subscribers.
```
donaldgray:
```suggestion
    /// Broadcast a change to the status of an deliverable, for any subscribers.
```

src/protagonist/API/Infrastructure/IDeliverableNotificationSender.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    Task SendDeliverableModifiedMessage<T>(NotificationRecord<T> notification,
+        CancellationToken cancellationToken = default) where T : class, IDeliverable;
+
+    /// <summary>
+    /// Broadcast a change to the status of multiple Assets, for any subscribers.
```
donaldgray:
```suggestion
    /// Broadcast a change to the status of multiple deliverables, for any subscribers.
```

src/protagonist/API/API.csproj
```diff
@@ -45,4 +45,8 @@
         </Content>
     </ItemGroup>
 
+    <ItemGroup>
+      <Folder Include="Infrastructure\Messaging\Asset\" />
```
donaldgray: Remove

src/protagonist/CleanupHandler/Adjunct/AdjunctDeletedHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+            // This means it's not a hosted asset, so nothing to do
+            if (request.Deliverable!.Origin == null)
+            {
+                logger.LogDebug("Adjunct {Asset}/{Adjunct} does not have an origin, so no deletion required",
+                    adjunct.AssetId, adjunct.Id);
+                return true;
+            }
+
+            // if the item exists in the db, assume the adjunct has been reingested after delete
+            if (dbContext.Adjuncts.Any(a => a.Id == adjunct.Id && a.AssetId == adjunct.AssetId))
```
donaldgray: Use the helper that's been introduced

```suggestion
            if (dbContext.Adjuncts.FindAdjunct(adjunct.Id, adjunct.AssetId).Any())
```

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+            
+            await adjunctBucketOperations.DeleteFromOriginBucket(adjunctBefore, settings);
+
+            return true;
+        }
+    }
+
+    // We only cleanup when the adjunct has moved from hosted to unhosted
+    private bool NoCleanupRequired(DLCS.Model.Assets.Adjunct adjunctBefore, DLCS.Model.Assets.Adjunct adjunctAfter) =>
+        adjunctBefore.ExternalId != null || (adjunctBefore.Origin == null && adjunctAfter.Origin != null);
```
donaldgray: Should this use `string.IsNullOrEmpty()` rather than `== null` (I can't recall whether there are situations where it could be empty or if it's guaranteed to always be null)?

src/protagonist/CleanupHandler/Asset/AssetUpdatedHandler.cs
```diff
@@ -196,13 +208,13 @@ private async Task CleanupRemoved(ImageDeliveryChannel deliveryChannelRemoved, A
                 CleanupFileDeliveryChannel(assetAfter, s3Objects.objectsToRemove);
                 break;
             default:
-                logger.LogDebug("policy {PolicyName} does not require any changes for asset {AssetId}",
+                logger.LogDebug("policy {PolicyName} does not require any changes for DLCS.Model.Assets.Asset {AssetId}",
```
donaldgray: Was this a find/replace update? "asset" seems simpler (same with other log messages)?

src/protagonist/DLCS.AWS/Settings/SNSSettings.cs
```diff
@@ -7,6 +7,11 @@ public class SNSSettings
     /// </summary>
     public string? AssetModifiedNotificationTopicArn { get; set; }
     
+    /// <summary>
+    /// Name of the SNS topic for notifying that adjuncts have been modified
```
donaldgray:
```suggestion
    /// Arn of the SNS topic for notifying that adjuncts have been modified
```

src/protagonist/DLCS.AWS/SNS/ITopicPublisher.cs
```diff
@@ -26,12 +17,22 @@ public Task<bool> PublishToCustomerCreatedTopic(CustomerCreatedNotification mess
     /// <returns>Boolean representing the overall success/failure status of request</returns>
     public Task<bool> PublishToBatchCompletedTopic(BatchCompletedNotification message,
         CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Asynchronously publishes a message to an Adjunct Modified SNS topic
```
donaldgray:
```suggestion
    /// Asynchronously publishes a message to deliverable modified SNS topic
```

src/protagonist/DLCS.AWS/SNS/TopicPublisher.cs
```diff
@@ -87,13 +61,45 @@ public async Task<bool> PublishToBatchCompletedTopic(BatchCompletedNotification
...
+        if (messages.Count == 1)
+        {
+            var singleMessage = messages[0];
+            return await PublishToDeliverableModifiedTopic(singleMessage, topicArn, cancellationToken);
+        }
+
+        const int maxSnsBatchSize = 5;
+        var allBatchSuccess = true;
+        var batchIdPrefix = Guid.NewGuid();
+        Arn.TryParse(topicArn, out var arn);
```
donaldgray: If we're parsing should we handle the TryParse failing?

src/protagonist/DLCS.Repository/Adjuncts/AdjunctX.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using DLCS.Core.Types;
+using DLCS.Model.Assets;
+
+namespace DLCS.Repository.Adjuncts;
+
+public static class AdjunctX
+{
+    public static IEnumerable<Adjunct> FindAdjunct(this IQueryable<Adjunct> adjuncts, string adjunctId, AssetId assetId) =>
```
donaldgray: As discussed, return `IQueryable`, rather than `IEnumerable`. The latter means that more work is done in the client, the former takes advantage of DB where possible.

donaldgray (approved):
Approving - the only thing that really needs done is the `IQueryable` / `IEnumerable` change.

Small but is it worth renaming the PR to avoid confusion when generating release notes - the "ability to delete adjuncts" was technically done in #1082, this is just the cleanup side of it.

src/protagonist/CleanupHandler/Adjunct/AdjunctUpdatedHandler.cs
JackLewis-digirati: `ExternalId` is a URI, so I think this should be ok?

donaldgray: What about `Origin`?

PR #1146, opened 2026-03-19
Use orchestrator url for hosted adjuncts id in manifest builder

### LINKED ISSUE #1079: Generated manifests include hosted manifests
Extend generated manifests to also support hosted adjuncts, in addition to externally hosted adjuncts.

The same rules from (#1074) are used for generation of IIIF, all that changes is the `"id"` value. It will now contain an Orchestrator path.

(#1074) may be filtering in `where external_id is not null` - this should be extended to `where coalesce(external_id, origina) is not null`. This will avoid items with no content (future implementation) from appearing until we are ready.

p-kaczynski:
## What does this change?

Implementes #1079 

Previously the `ManifestV3Builder` included only external adjuncts. This filters the db query for adjuncts that are either external or hosted and applies a logic split when setting the `Id` property of generated adjunct manifest objects.

### DIFF
--- a/src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
+++ b/src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
@@ -36,6 +36,7 @@ public class ManifestV3Builder : ManifestBuilderBase<Manifest>
     private readonly IIIIFAuthBuilder authBuilder;
     private readonly ILogger<ManifestV3Builder> logger;
     private const string AdjunctAnnotationRoutePrefix = "adjunct-annotations";
+    private const string AdjunctRoutePrefix = "adjuncts";
 
     /// <summary>
     /// Implementation of <see cref="IBuildManifests{T}"/> responsible for generating IIIF v3 manifest
@@ -502,7 +503,7 @@ public class ManifestV3Builder : ManifestBuilderBase<Manifest>
                     canvas.Annotations ??= [];
                     canvas.Annotations.Add(new AnnotationPage
                     {
-                        Id = adjunct.ExternalId.ToString(),
+                        Id = GetAdjunctId(adjunct),
                         Label = adjunct.Label,
                     });
                     break;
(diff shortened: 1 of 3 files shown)


src/protagonist/Orchestrator/Infrastructure/IIIF/Manifests/ManifestV3Builder.cs
```diff
@@ -551,15 +552,29 @@ void CreateInlineAnnotation(Canvas currentCanvas, Adjunct adjunct)
...
+            => adjunct switch
+            {
+                { Origin: null, ExternalId: { } externalUri } => externalUri.ToString(),
+                { Origin: not null, ExternalId: null } => assetPathGenerator.GetFullPathForRequest(new BasicPathElements
+                {
+                    Space = adjunct.AssetId.Space,
+                    AssetPath = $"{adjunct.AssetId.Asset}/{adjunct.Id}",
+                    RoutePrefix = AdjunctRoutePrefix,
+                    CustomerPathValue = adjunct.AssetId.Customer.ToString(),
+                }),
```
donaldgray, line 572: The ticket this PR is resolving is to handle outputting canonical paths for adjuncts only, rather than rewritten paths. Rewritten paths are for ticket #1098. However, to enforce canonical paths _only_ we'd need to set `useNativeFormat: true` when calling `GetFullPathForRequest()` which seem pointless as this should handle rewrites as-is, which is good as that's what `GetFullPathForRequest` and `BasicPathElements` etc should handle.

In hindsight the split of tickets was probably not required, the restriction that assetPath is _always_ `<asset-id>/<adjunct-id>` (captured in the `"Warning"` note in #1098) was added after I initially split the tickets. Enforcing this rule means that our existing path rewrite rules work out of the box.

I think it's worth keeping #1098 as a separate ticket but I don't think any code changes will be required (I'm 95% but could be overlooking something). It'll just be to confirm that's the case, maybe add tickets, update dev environment etc and have an explicit ticket for testers to confirm this works.

donaldgray (approved):

PR #1165, opened 2026-03-30
Support ?include=adjuncts query parameter

### LINKED ISSUE #1144: Bulk reading of adjuncts
Extend endpoints that accept [Asset Query Syntax](https://dlcs.github.io/public-docs/api-doc/asset-queries/) to accept the new [`?include=adjuncts`](https://dlcs.github.io/public-docs/api-doc/asset-queries/#include-children) property. See https://github.com/dlcs/protagonist/blob/develop/docs/rfcs/022-stub-assets.md#bulk-reading for more information.

In addition to supporting the `?q=` and shortcut parameters we now need to support the `?include=adjuncts` parameter. When specified, this instructs the returned Assets to include child adjuncts as embedded resources in the [`"adjuncts"`](https://dlcs.github.io/public-docs/api-doc/asset/#adjuncts) property if they exist, rather than returning a hypermedia link. E.g.

## Without `?include=adjuncts` (as-is)
```jsonc
{
  "member": [{
    "@context": "https://api.dlcs.example/contexts/Image.jsonld",
    "@id": "https://api.dlcs.example/customers/7/spaces/49/images/1b",
    "@type": "vocab:Image",
    "id": "1b",
    "space": 49,
    // adjuncts is a URI
    "adjuncts": "https://api.dlcs.example/customers/7/spaces/49/images/1b/adjuncts"
  }]
}
```

## With `?include=adjuncts`
```jsonc
{
  "member": [{
    "@context": "https://api.dlcs.example/contexts/Image.jsonld",
    "@id": "https://api.dlcs.example/customers/7/spaces/49/images/1b",
    "@type": "vocab:Image",
    "id": "1b",
    "space": 49,
    // adjuncts is an array of embedded adjuncts
    "adjuncts": [{ 
      "@id": "https://api.dlcs.example/customers/7/spaces/49/images/1b/my-adjunct"
    }]
  }]
}
```

AC:
* [ ] Update all asset query syntax to support `?include=`
  * [ ] GET `/customers/{customer}/allImages`
  * [ ] GET `/customers/{customer}/spaces/{space}/images`
  * [ ] GET `/customers/{customer}/batches/{batchId}/images`
  * [ ] GET `/customers/{customer}/batches/{batchId}/assets`
* [ ] Any value other than `adjuncts` is ignored
* [ ] On Asset hydra model, `"adjuncts"` is a URI by default. If payload is response to an `?include=` request then it is a collection of embedded Adjuncts, or an empty array if there are none.
* [ ] HydraCollection `"view"` properties (e.g. "next", "prev" etc) should reflect `?include=adjuncts` if included.

----
## Implementation Note

`AssetFilter` models the incoming Asset Query Syntax. See helpers in `AssetConverter` for how this is generated.

donaldgray:
## What does this change?

Extends asset query accepting endpoints to handle `?include` query parameter, the only valid value being `adjuncts` - all others are ignored.

> [!NOTE]
> The high additions/deletions numbers are primarily due to moving existing classes/methods around.

### Summary of changes

All handlers that used `AssetFilter` now take `AssetQueryModel`. This new model contains enough to build appropriate sql queries to requests.

HydraModel `Asset.Adjunct` can be `string` or `Adjunct[]` object. To avoid custom serialisation I opted to use a `JToken`. The default HydraModel plumbing will continue to set this to the appopriate URI, `AssetConverter.ToHydra()` may overwrite this if `?include=adjuncts` has been specified.

Moved a few classes + methods from the DLCS.Models and DLCS.Repository into the API project. These were all solely used for API AssetQuery filtering so made sense to move these together. Including
* Introduce `AssetQueryModel`, which contains `AssetFilter` (was in DLCS.Model) and new `AssetInclude` object.
* Move `IPagedRequest`, `IAssetFilterableRequest` and `IOrderableRequest` as well as paging models + helpers.
* Move parsing logic out of `AssetConverter` into `AssetQueryConverter`. Simplified call from controllers to build the model - all controllers made the same 2 calls to get the model : build from `q` -> override with specific query parameters. This is now done in 1 call.
* Move majority of `AssetQueryX` from repository => API project. This contained helpers for filtering/ordering etc. Added `.IncludeRelated()` helpers.

All handlers that apply filters call `.IncludeRelated()` separately. Initially I aimed to add it to `.ApplyAssetFilter()` but the "get batch assets" endpoint had called `.Select()` in IQueryable before this method was hit (because `BatchAssets` is entry point for query, not `Assets`), which cause the `.Include()` to fail. I think this is pretty clean though - it keeps the "filter" and "include" as 2 distinct steps and you could have one without the other.

Updated `GetSpaceImages` - this could take CustomerId in request OR find the current customer, which isn't used - this was a legacy/initial implementation that isn't used in any codepaths.

Resolves #1144


src/protagonist/API/Features/Assets/Query/AssetQueryConverter.cs
```diff
@@ -0,0 +1,135 @@
...
+        var q = request.GetFirstQueryParamValue(QueryParameters.Query);
+        if (!q.HasText()) return null;
+
+        var imageQuery = ImageQuery.Parse(q);
+        return imageQuery?.ToAssetFilter();
+    }
+
+    /// <summary>
+    /// Inspect the request for string1, number1 etc metadata fields.
+    /// Create a new AssetFilter if present, or add to the one passed in.
```
JackLewis-digirati: tiny - `<see cref="AssetFilter" />`?

src/protagonist/API/Features/Assets/Query/AssetQueryConverter.cs
```diff
@@ -0,0 +1,135 @@
...
+        return assetFilter;
+
+        AssetFilter SafeFilter() => assetFilter ??= new AssetFilter();
+    }
+
+    /// <summary>
+    /// Convert the Hydra <see cref="ImageQuery"/> to a <see cref="AssetFilter"/>.
+    /// </summary>
+    /// <remarks>
+    /// ImageQuery is a Hydra object, map to similar object for business logic use
```
JackLewis-digirati: tiny - `<see cref="ImageQuery"/>`?

src/protagonist/API/Infrastructure/Requests/IOrderableRequest.cs
```diff
@@ -0,0 +1,10 @@
+namespace API.Infrastructure.Requests;
+
+/// <summary>
+/// Represents a request that can be ordered
+/// </summary>
+public interface IOrderableRequest
+{
+    string? Field { get; set; }
```
JackLewis-digirati, line 11: missing comments?

JackLewis-digirati (approved):

src/protagonist/API/Features/Assets/Query/AssetQueryConverter.cs
donaldgray: Added

src/protagonist/API/Features/Assets/Query/AssetQueryConverter.cs
donaldgray: Added

src/protagonist/API/Infrastructure/Requests/IOrderableRequest.cs
donaldgray: I thought they were self-explanatory tbh but added comments. Also extended to `IPagedRequest` as that didn't have any at field level and was copied from same source.

PR #1168, opened 2026-03-31
Add provides to adjunct

### LINKED ISSUE #1153: Accept 'provides' for annotations
Support the `provides` property for adjuncts: https://dlcs.github.io/public-docs/api-doc/adjuncts/#provides

> Only used when iiifLink is "inlineAnnotation". It appears as the provides property of the annotation, as in the example above.
> 
> The value of provides can be a single string or a comma-separated list of strings. The value will always appear in the resulting IIIF as an array of provides values, whether there is one or more than one. The values should be taken from the list defined in the IIIF Presentation specification (e.g., “audioDescription” or “subtitles”).

* [ ] Support new `provides` string property on an adjunct. This is optional and only used for `inlineAnnotation` types but can be supplied for other types. Do not validate provided values against a whitelist, accept any string values (see Tom's comment, below)

> [!NOTE]
> `provides` is a IIIF Presentation 4 property so we are only storing the value for now, it won't be output until we have support for P4.

See #1112 for similar `motivation` property implemented for `inlineAnnotations`

JackLewis-digirati:
Resolves #1153

## What does this change?

This PR adds the `provides` field to adjuncts in preparation of presentation 4, which is a list of strings like this:

```
{
    "id": "someAdjunctId",
    "@type": "Image",
    "externalId": "https://some-location.com/an-adjunct",
    "iiifLink": "inlineAnnotation",
    "provides": "test"
}
```

This works similarly to `motivation` in that it can be provided for any `iiifLink` type, but will only be valid for `inlineAnnotation` once presentation 4 is supported.  Similarly, the values for `provides` are not checked, but should match the ones outlined in the [documentation](https://preview.iiif.io/api/prezi-4/presentation/4.0/model/#provides)

## Database Migration

> [!NOTE]
> Details of migration. What does it change? Is it breaking/non-breaking?
>
> - **Migration name:** `20260331163532_Add provides to adjuncts`
> - **What it does:** (adds column `provides` to the `adjunct` table, with a type of `text`)
> - **Breaking Change?** No

### DIFF
--- a/src/protagonist/DLCS.Model/Assets/Adjunct.cs
+++ b/src/protagonist/DLCS.Model/Assets/Adjunct.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using DLCS.Core.Types;
 using IIIF.Presentation.V3.Strings;
@@ -90,6 +91,11 @@ public class Adjunct : IDeliverable
     /// The reason why this adjunct exists.
     /// </summary>
     public string? Motivation { get; set; }
+    
+    /// <summary>
+    /// An additional set of features or functionality this adjunct provides
+    /// </summary>
+    public List<string>? Provides { get; set; }
 
     public Asset Asset { get; set; } = null!;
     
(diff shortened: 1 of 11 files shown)


src/protagonist/DLCS.Model/Assets/Adjunct.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
```
donaldgray:
```suggestion
```

donaldgray (approved):

PR #1169, opened 2026-04-01
Implement bulk delete of adjuncts

### LINKED ISSUE #1151: Bulk delete adjuncts
Support bulk deleting of adjuncts via `POST /customers/{customerId}/deleteAdjuncts` (similar to `POST /customers/{customerId}/deleteImages`).

This accepts a Hydra collection:

```jsonc
{
 "member": [
    { "id": "1/1/foo", "adjunct": ["ocr.txt", "mets.xml"] }
  ]
}
```

* [x] Validate request (see `ImageIdListValidator` for existing impl. of this)
  * [x] Reject if members empty
  * [x] Reject if list contains duplicates
  * [x] Reject if total count of `adjuncts` is greater than maximum allowed (Images uses `MaxImageListSize` config value)
* [x] Return 400 if no adjuncts are deleted due to them not being in the DB OR due to an error deleting.
* [x] Return 204 if at least 1 adjunct deleted (if member contains 10 values but only 1 adjunct was deleted this is treated as a success).
* [x] Raise an adjunct deleted notification for every adjunct that was removed from DB as there may be further cleanup.

---
## Implementation Details

`/deleteImages` accepts a collection of `"id"` values, each `"id"` is an AssetId. This was based on Deliverator implementation.

```json
{
 "member": [
    { "id": "1/1/foo" },
    { "id": "1/99/bar" }
  ]
}
```

JackLewis-digirati:
Resolves #1151

## What does this change?

This PR adds anew endpoint to allow the bulk deletion of adjuncts from assets and can be found at `POST /customers/{customerId}/deleteAdjuncts`.  This also requires a request `body` like this:

```json
{
 "member": [
    { "id": "1/1/foo", "adjunct": ["ocr.txt", "mets.xml"] }
  ]
}
```


src/protagonist/DLCS.Repository/Adjuncts/AdjunctX.cs
```diff
@@ -2,11 +2,26 @@
...
+using LinqKit.Core;
 
 namespace DLCS.Repository.Adjuncts;
 
 public static class AdjunctX
 {
     public static IQueryable<Adjunct> FindAdjunct(this IQueryable<Adjunct> adjuncts, string adjunctId, AssetId assetId) =>
         adjuncts.Where(a => a.Id == adjunctId && a.AssetId == assetId);
+
+    public static IQueryable<Adjunct> FindAdjuncts(this IQueryable<Adjunct> adjuncts,
```
donaldgray, line 21: Comment? Worth testing in isolation?

src/protagonist/API/Features/Customer/CustomerAdjunctsController.cs
```diff
@@ -0,0 +1,72 @@
...
+using Hydra.Collections;
+using Hydra.Model;
+using MediatR;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
+
+namespace API.Features.Customer;
+
+/// <summary>
+/// Controller for handling bulk requests for assets associated with a customer
```
donaldgray:
```suggestion
/// Controller for handling bulk requests for adjuncts associated with a customer 
```

src/protagonist/DLCS.Model/AdjunctIdentifierOnly.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using DLCS.Core.Types;
+
+namespace DLCS.Model;
+
+/// <summary>
+/// An object that contains the minimal properties to identify an adjunct
+/// </summary>
+public class AdjunctIdentifierOnly
```
donaldgray: Think this can be named better as it's not just adjunct ids.

src/protagonist/API/Features/Customer/Validation/AdjunctIdListValidator.cs
```diff
@@ -0,0 +1,40 @@
...
+            .Must(m => m?.ConvertToDictionary().Any(kvp => kvp.Value.Distinct().Count() == kvp.Value.Count) ?? true)
+            .WithMessage((_, mem) =>
+            {
+                var dupes = mem!.ConvertToDictionary().Where(kvp => kvp.Value.Distinct().Count() != kvp.Value.Count).ToList();
+                return $"Members contains {dupes.Count} duplicate Id(s): {string.Join(",", dupes.Select(d => $"asset Id: {d.Key} : adjunct id: {string.Join(',', d.Value.Distinct())}"))}";
+            });
+        
+        var maxBatch = apiSettings.Value.MaxImageListSize;
+        RuleFor(c => c)
+            .Must(m => m.Flatten().Count() <= maxBatch)
```
donaldgray: Seeing a warning for this in IDE

src/protagonist/API.Tests/Features/Adjuncts/Validation/AdjunctIdListValidatorTests.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        var adjuncts = new AdjunctIdentifierOnly[2];
+        adjuncts[0] = new AdjunctIdentifierOnly
+        {
+            Id = AssetIdGenerator.GetAssetId().ToString(),
+            Adjunct = ["first"]
+        };
+        adjuncts[1] = new AdjunctIdentifierOnly
+        {
+            Id = AssetIdGenerator.GetAssetId(asset: $"{nameof(Valid_WhenSingleAdjunctMultipleAsset)}_1").ToString(),
```
donaldgray: This can be simplified, it takes a property to do exactly what you're doing here (ie add a value after the main asset id to vary it).

```suggestion
            Id = AssetIdGenerator.GetAssetId(assetPostfix: "_1").ToString(),
```

src/protagonist/API/Features/Customer/Requests/DeleteMultipleAdjunctsById.cs
```diff
@@ -0,0 +1,94 @@
...
+using DLCS.Core.Types;
+using DLCS.Model.Assets;
+using DLCS.Repository;
+using DLCS.Repository.Adjuncts;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace API.Features.Customer.Requests;
+
+public class DeleteMultipleAdjunctsById(
```
donaldgray, line 16: XML comment? I think all other mediatr handlers have a simple explanatory comment.

src/protagonist/API/Features/Customer/Validation/AdjunctIdListValidator.cs
```diff
@@ -0,0 +1,40 @@
...
+            .WithMessage($"Maximum adjuncts in single batch is {maxBatch}");
+    }
+    
+    protected override bool PreValidate(ValidationContext<AdjunctIdentifierOnly[]?> context, ValidationResult result)
```
donaldgray: This is now in 3 classes `BulkAssetPatchValidator`, `AdjunctIdListValidator` and `ImageIdListValidator` - worth refactoring to share?

src/protagonist/API/Features/Adjuncts/Infrastructure/AdjunctIdentifierOnlyX.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                adjunctDictionary[adjunctIdentifier.Id].AddRange(adjunctIdentifier.Adjunct);
+            }
+        }
+        
+        return adjunctDictionary;
+    }
+
+    /// <summary>
+    /// Flattens a list of <see cref="AdjunctIdentifierOnly"/> into a serikes of key value pairs of combinations
```
donaldgray: Looking at the usage of this; there's no need to create a new KVP, all you need is the count of total adjuncts.

There's a typo in xml comment

```suggestion
    /// Flattens a list of <see cref="AdjunctIdentifierOnly"/> into a series of key value pairs of combinations
```

donaldgray (commented):
Overall LGTM, one observation is around the validation and number of times we're iterating the list of members. Currently:

* Members is passed to the validator.
* Internally the validator calls `.ConvertToDictionary()`
* If validation fails, `.ConvertToDictionary()` is called again to get return message.
* If validation passes, `.ConvertToDictionary()` is called when passing the mediatr request (so, always twice).
* Inside the mediatr handler it's _kinda_ called again in `ConvertAssetIds` - but this time to `AssetId` and it throws if customer is not valid.

Instead, could you:
* Call `.ConvertToDictionary()` once in the controller. This can return `Dict<AssetId, List<str>>` but still de-dup (validating AssetIds are valid and Customer is correct)
* Pass this Dictionary to the validator, don't need to convert to check for duplicates.
* Pass this Dictionary to the mediatr handler too

src/protagonist/API/Features/Customer/Validation/AdjunctIdListValidator.cs
JackLewis-digirati: `BulkAssetPatcher` is slightly different as it's validating a property inside the object, rather than the object itself, but sharing between the other 2 is fine

donaldgray (approved):

PR #1170, opened 2026-04-08
Immediate completion of none channel

### LINKED ISSUE #1154: API auto-finish none channel
Assets with the `none` channel are sent to the Engine but the Engine does no work on them, it automatically finishes it, see https://github.com/dlcs/protagonist/blob/8748621a873f0fec95ce3ed9010e044fac74edef/src/protagonist/Engine/Ingest/IngestExecutor.cs#L109-L116

The logic to auto-complete the Asset can be moved to the API, rather than making an unnecessary request to the Engine.

* [x] Any assets with `none` channel are automatically finished in API, without being sent to Engine.
* [x] This could result in a Batch being 'completed' instantly, which should raise a notification.

JackLewis-digirati:
Resolves #1154

## What does this change?

This PR modifies the `None` delivery channel to complete without needing to be ingested - this PR will also raise a `completed` batch notification immediately if required


src/protagonist/DLCS.Model/Assets/ImageDeliveryChannelX.cs
```diff
@@ -47,6 +47,19 @@ public static class ImageDeliveryChannelX
...
             throwIfNotFound);
     
+    /// <summary>
+    /// Get ImageDeliveryChannel record for None channel, optionally throwing if not found.
+    /// </summary>
+    /// <param name="imageDeliveryChannels">Collection of <see cref="DeliveryChannelPolicy"/></param>
+    /// <param name="throwIfNotFound">If true, and image not found, then exception will the thrown</param>
+    /// <returns><see cref="DeliveryChannelPolicy"/>, if found.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if image policy not found and throwIfNotFound = true</exception>
+    public static ImageDeliveryChannel? GetNoneChannel( //todo: test
```
donaldgray: ` //todo: test`

It looks like callers of this don't do anything with the channel, they just check it exists. Would it be better to use `.HasDeliveryChannel()` or `.HasSingleDeliveryChannel()` instead?

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -106,15 +106,6 @@ public async Task<IngestResult> IngestAsset(Asset asset, CustomerOriginStrategy
         var sw = Stopwatch.StartNew();
         var context = new IngestionContext(asset);
 
-        // If the asset has the `none` delivery channel specified, skip processing and mark the ingest as being complete
-        if (asset.HasSingleDeliveryChannel(AssetDeliveryChannels.None))
```
donaldgray, line 110: This is still required.

The notification sent to the engine doesn't contain the delivery-channels, it contains the bare minimum of information for Engine to then read from DB. Engine always operates on the _current_ state of the Asset, not the state of the asset at the time the ingest request was queued. Between the API queueing asset and that asset being picked up it could have been assigned back to `None` only.

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
```diff
@@ -8,19 +8,28 @@
...
 {
-    private readonly IDefaultDeliveryChannelRepository defaultDeliveryChannelRepository;
-    private readonly IDeliveryChannelPolicyRepository deliveryChannelPolicyRepository;
-    private readonly ILogger<DeliveryChannelProcessor> logger;
     private const string FileNonePolicy = "none";
 
-    public DeliveryChannelProcessor(IDefaultDeliveryChannelRepository defaultDeliveryChannelRepository,
-        IDeliveryChannelPolicyRepository deliveryChannelPolicyRepository, ILogger<DeliveryChannelProcessor> logger)
+    /// <summary>
+    /// Adds details of the delivery channel policy to the delivery channels
```
donaldgray: Why is this needed? It feels odd to expose this, ideally any DeliveryChannel wrangling should be in `ProcessImageDeliveryChannels()`, keep it all contained to single place and prevent consumers needing to then do more d-c _stuff_ externally? 

Is it to make sure that the full policy is populated for raising notification? If so, from reading code wouldn't that be covered by `AddExplicitNoneChannel()` already?

src/protagonist/Engine/Ingest/IngestExecutor.cs
```diff
@@ -106,15 +106,6 @@ public async Task<IngestResult> IngestAsset(Asset asset, CustomerOriginStrategy
...
-        {
-            context.WithStorage();
```
donaldgray, line 112: Reading this I'm realising this is a change in behaviour - previously Engine would create an "empty" `ImageStorage` record for `none` channel. We'll need to make sure the API does this too - initially that feels like Engine concerns bleeding into API but I _think_ it's okay as it'll only ever be setting up an "empty" record to shorten processing. I don't think it'll be an issue for new assets but if an asset moves from "not-none" to `none` channel we'll need to clear down any previous `ImageStorage` values.

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -110,12 +110,18 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
...
             {
-                var deliveryChannelChanged = await deliveryChannelProcessor.ProcessImageDeliveryChannels(
+                var deliveryChannelsRequireEngineNotification = await deliveryChannelProcessor.ProcessImageDeliveryChannels(
                     assetFromDatabase, updatedAsset, assetBeforeProcessing.DeliveryChannelsBeforeProcessing);
-                if (deliveryChannelChanged)
+                if (deliveryChannelsRequireEngineNotification)
                 {
                     requiresEngineNotification = true;
                 }
+                // After processing delivery channels, the none channel was set - so no need to do ingestion
```
donaldgray: `ProcessImageDeliveryChannels()` should return `false` for None channel - if it is returning `true` then something is amiss?

src/protagonist/API/Features/DeliveryChannels/DataAccess/DeliveryChannelPolicyRepository.cs
```diff
@@ -10,29 +11,34 @@
...
+
+        return deliveryChannelPolicies.RetrieveDeliveryChannel(customerId, channel, policy);
     }
+    
+    public async Task<DeliveryChannelPolicy> RetrieveDeliveryChannelPolicy(int customerId, string channel, int policyId)
+    {
+        var deliveryChannelPolicies = await RetrieveFromCache(customerId);
 
-    public async Task<DeliveryChannelPolicy> RetrieveDeliveryChannelPolicy(int customerId, string channel, string policy)
+        return deliveryChannelPolicies.First(dcp => dcp.Id == policyId && dcp.Channel == channel);
```
donaldgray: Are you expecting there to be multiple matches, in which case you want the first? (`.First()` works). 

Or are you expecting there to be one and only one match? (if so, it should be `.Single()`).

donaldgray (changes requested):

src/protagonist/API/Features/Image/Ingest/DeliveryChannelProcessor.cs
JackLewis-digirati: not quite - the issue was a catch-22 - effectively the `HydraModel` needs details of the `DeliveryChannelPolicy` itself to populate the correct values on return.  However, by filling these values in, it acts like there's a new `DCP` that needs to be added to the table and so gets a conflict on `database.SaveChanges()`.  Previously, this didn't matter as we fill out the `ImageDeliveryChannel` enough to link to the `DCP` (this is the `AddExplicitNone` method) to allow the DB save and further down the line, the engine notification logic discards the `asset` and retrieves it again from DB (which has all the additional details). 

As we're no longer going through the batch processing logic for `None` channel, these details need to be filled out another way - I could have retrieved the asset from the DB again, but it felt like more than was needed as we should already have the details cached from the initial processing etc. These details are set after the database save, which is what this method is doing

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
JackLewis-digirati: `ProcessImageDeliveryChannels` does return false, but `requiresEngineNotification` can technically be set to `false` (even if it doesn't currently) in `AssetPreparer.PrepareAssetForUpsert` - so this was an attempt to only change it if we get back the `none` channel and the `AssetPreparer` has decided we need to ingest (otherwise continue as previous behaviour)

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -44,7 +44,9 @@ public async Task<bool> UpdateIngestedDeliverable(IDeliverable deliverable, Imag
                 }
             }
 
-            await UpsertImageStorage(assetId, imageStorage, cancellationToken);
+            await DlcsContext.ImageStorages.UpsertImageStorageRecord(imageStorage, cancellationToken);
+
+            //await imageStorageRepository.UpsertImageStorageRecord(imageStorage, cancellationToken);
```
donaldgray: Commented out code
```suggestion
```

src/protagonist/DLCS.Repository.Tests/Storage/ImageStorageXTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using DLCS.Model.Assets;
+using DLCS.Repository.Storage;
+using Test.Helpers.Data;
+using Test.Helpers.Integration;
+
+namespace DLCS.Repository.Tests.Storage;
+
+[Trait("Category", "Integration")]
+[Collection(DatabaseCollection.CollectionName)]
+public class ImageStorageXTests
```
donaldgray, line 13: Rename file to match type

src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
```diff
@@ -143,6 +127,16 @@ public async Task<ModifyEntityResult<Asset>> Handle(CreateOrUpdateImage request,
             return await IngestAndGenerateResult(assetAfterSave, existingAsset != null, cancellationToken);
         }
 
+        if (assetAfterSave.HasSingleDeliveryChannel(AssetDeliveryChannels.None))
```
donaldgray, line 132: Comment as to why this is necessary? I remember the reason because we discussed it yesterday but worth a quick outline here for future ref.

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -110,12 +101,30 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
...
+                var deliveryChannelsRequireEngineNotification = await deliveryChannelProcessor.ProcessImageDeliveryChannels(
                     assetFromDatabase, updatedAsset, assetBeforeProcessing.DeliveryChannelsBeforeProcessing);
-                if (deliveryChannelChanged)
+                if (deliveryChannelsRequireEngineNotification)
                 {
                     requiresEngineNotification = true;
                 }
+
+                if (updatedAsset.ImageDeliveryChannels.Count == 1 &&
+                    updatedAsset.HasSingleDeliveryChannel(AssetDeliveryChannels.None))
```
donaldgray: `updatedAsset.ImageDeliveryChannels.Count == 1 &&` is unnecessary as `.HasSingleDeliveryChannel()` is already handling that.

```suggestion
                if (updatedAsset.HasSingleDeliveryChannel(AssetDeliveryChannels.None))
```

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -110,12 +101,30 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
...
+                    // no engine notification, so 0 out the image storage record
+                    var imageStorage = new ImageStorage
+                    {
+                        Id = updatedAsset.GetAssetId(),
+                        Customer = updatedAsset.Customer,
+                        Space = updatedAsset.Space,
+                        LastChecked = DateTime.UtcNow
+                    };
+                    
+                    await assetRepository.UpsertImageStorageRecord(imageStorage, cancellationToken);
```
donaldgray: Rather than `AssetProcessor` creating a "0" item, passing it to `assetRepository.UpsertImageStorageRecord()`, which directly calls `ImageStorages.UpsertImageStorageRecord();` is it cleaner for this call to just be 

```cs
// no need to notify the engine with the none channel
requiresEngineNotification = false;

// Dodgy name but caller doesn't need to create 
await assetRepository.ResetImageStorage(updatedAsset.GetAssetId(), cancellationToken);
```

Ultimately a small change but maybe cleaner, means the assetRepo call isn't just a pass-thru, it's creating the empty object and setting it. If we added a new field to that entity the repo is where the change would take place, rather than caller. This seems like a cleaner separation of concerns (ie the repo does all the data things, the caller just tells it what needs doing)?

donaldgray (approved):

PR #1171, opened 2026-04-14
Create batch of adjuncts

### LINKED ISSUE #1149: Adjunct Batch Creation
Implement API endpoint `POST /customers/{customer}/adjunctQueue`. This is the equivalent of `/customers/{customer}/queue` and works in an identical way, with the difference being this creates an `AdjunctBatch` resource. 

An adjunct batch is created via a HydraCollection of `Adjunct`. Each `member` needs to be a valid `Adjunct`, following validation rules applied to creation of individual adjuncts. Unlike 'basic' adjunct endpoints, it's possible to specifc the id of the Asset that the adjunct is for, allowing adjuncts to be ingested for multiple assets in a single batch.

E.g. the following creates 3 adjuncts; 2 for `99/100/xxy_page_1` and 1 for `99/200/xxy_page_2`

```jsonc
{
  "@type": "Collection",
  "member": [ 
    {
      "id": "mets.xml",
      "asset": "99/100/xxy_page_1",
      // other standard fields
    },
    {
      "id": "ocr.txt",
      "asset": "https://api.dlcs.example/customers/99/spaces/100/images/xxy_page_1",
      // other standard fields
    },
    {
      "id": "mets.xml",
      "asset": "99/200/xxy_page_2",
      // other standard fields
    }
  ]
}
```

* [ ] Add `POST /customers/{customer}/adjunctQueue`
* [ ] Validate each member in accordance with individual adjunct validation rules. Additional validation rules:
  * [ ] Reject if `"asset"` missing
  * [ ] Reject if `"member"` is null or empty
  * [ ] Reject if there are duplicates adjuncts specified
  * [ ] Reject if `"member"` size exceeds `MaxBatchSize`
* [ ] Accept `"asset"` as fully qualified URI or short form
* [ ] Notify engine / mark as completed using same rules as done when adding adjunct to individual asset.
* [ ] Store batchId value against Adjunct in db. This is overwritten each time but used to identify 'current' batch.
* [ ] Store history of batch:adjunctBatch (similar to `BatchAssets`)
* [ ] Add `"batch"` property to adjunct hydra model that exposes the full AdjunctBatch URI

> [!NOTE]
> Validation happens up front. Reject the entire batch if 1 issue found, no partial implementation.

### LINKED ISSUE #1150: Adjunct gains `"asset"`
Internally an adjunct stores a relationship to the parent asset. Extend the Adjunct hydramodel to output `"asset"` property.

This will be a fully qualified URI, which is the same as the parent Assets `"@id"`.

donaldgray:
## What does this change?

Adds support for creating batches of adjuncts. The concept of `Batch` already exists for assets, so this introduces a new `AdjunctBatch` concept - it is very similar to asset/batches but different enough to warrant new entities etc.

### Extend Hydra Model
Resolves #1150 

Adjunct HydraModel gains `Asset` property 

### API + Engine handle AdjunctBatch

Add new entities:
* `AdjunctBatch` - this is the overall record of items/success/failed etc counts for group of adjuncts (equivalent of `Batch` for assets)
  * `Batch` and `AdjunctBatch` are `IDeliverableBatch` as there is some commonality between them.
  * Rename `BatchAssetStatus` -> `BatchStatus` as it's shared between the 2 different batch types.
* `AdjunctBatchAdjunct` - collection of all `Adjunct`s in an `AdjunctBatch`, maintains a record of processing for adjunct-batches (equivalent of `BatchAsset`)
  * `BatchAsset` and `AdjunctBatchAdjunct` are both `IDeliverableBatchItem`

Refactored adjunct processing code out of `CreateOrUpdateAdjunct` and into `AdjunctUpsertService`. This handles mapping of fields and raising notifications. 

As part of extracting to `AdjunctUpsertService` I updated engine notifications to be sent in batches, which is more efficient. This has resulted in a slight loss of identifying which actual message sending failed as we're no longer sending one at a time but this should be a very rare occurrence so the trade-off seems worthwhile.

New controller for `adjunctQueue` - currently only supporting POST operations to create a batch.

`CreateAdjunctBatch` mediatr request, this
* Validates incoming payload and finds existing adjuncts from DB.
* Creates a db transaction
* Iterates through all adjuncts in payload and calls adjunctUpsertService to process
* Creates batch + assigns batchId to all adjuncts
* Creates `AdjunctBatchAdjunct` record with correct status.
* Commits transaction
* Send notifications (Engine + Create/Update)

Extend `AdjunctConverter` to include an overload of `ToDlcsModel()` - this allows either the short or long-form assetId to be passed.

Use `"adjunct"` queue name, rather than `"default"`. Saving `"default"` for assets, introducing a new one will make it clearer the differences. This change won't be immediately obvious but is surface in customer-queue endpoints, it allows customers to see how many items are outstanding in each type of queue.

Refactor `IEngineClient` to make use of `IDeliverable` for single async ingest requests. Stick to `Asset` and `Adjunct` for batch requests as there are more differences (namely priority or not). Internally refactor `EngineClient` to share more internally based in `IDeliverable` type.

Resolves #1149

### Other Changes

Add `HydraExceptionFilter` - this catches any unhandled exceptions. It's essentially replacing `HydraController.HandleHydraRequest()` so calls to that were removed. The resulting controller code is now clearer to read.

## Database Migration

> [!NOTE]
>
> - **Migration name:** `Add AdjunctBatch`
> - **What it does:** Adds `AdjunctBatches` table (with sequence `adjunct_batch_id_sequence` and index on `(Customer, Submitted)`), `AdjunctBatchAdjuncts` junction table, and a nullable `Batch` FK column on `Adjuncts`.
> - **Breaking Change?** No — all new tables/columns; existing data unaffected.


src/protagonist/API/Converters/AdjunctConverter.cs
```diff
@@ -89,6 +118,56 @@ public static DLCS.HydraModel.Adjunct ToHydra(this Adjunct adjunct, UrlRoots url
...
+                }
+            }
+
+            return false;
+        }
+
+        // Try short form: customer/space/assetId
+        try
+        {
+            assetId = AssetId.FromString(assetField);
```
JackLewis-digirati, line 165: tiny - worth doing the short form first as it's less work and likely more common?

src/protagonist/DLCS.HydraModel/AdjunctBatch.cs
```diff
@@ -0,0 +1,63 @@
...
+    public int Errors { get; set; }
+
+    [RdfProperty(Description = "Date the batch was finished, if it has finished",
+        Range = Names.XmlSchema.DateTime, ReadOnly = true, WriteOnly = false)]
+    [JsonProperty(Order = 15, PropertyName = "finished")]
+    public DateTime? Finished { get; set; }
+
+    [HydraLink(Description = "Collection of adjuncts currently claimed by this batch",
+        Range = Names.Hydra.Collection, ReadOnly = true, WriteOnly = false)]
+    [JsonProperty(Order = 20, PropertyName = "currentAdjuncts")]
```
JackLewis-digirati, line 56: is there a reason for the jump in ordering from 15 to 20?

src/protagonist/DLCS.Repository.Tests/Messaging/EngineClientTests.cs
```diff
@@ -121,6 +122,198 @@ public async Task AsynchronousIngest_QueuesMessage(int? batchId)
...
     }
 
+    [Fact]
+    public async Task AsynchronousIngest_QueuesMessage_ForAdjunct()
+    {
+        // Arrange
+        var adjunct = new Adjunct
+        {
+            Id = "adjunct-1",
+            AssetId = AssetId.FromString("99/1/ingest-asset"),
```
JackLewis-digirati, line 132: `AssetIdGenerator.GetAssetId()`? (several in this PR)

JackLewis-digirati (approved):

src/protagonist/DLCS.HydraModel/AdjunctBatch.cs
donaldgray: It was to add a buffer between properties (which finish at 15) and then the hypermedia links (starting at 20). If we want to add more props or links then no need to shuffle numbers straight away. (FWIW PR #1173 removes these links, I added them assuming that the endpoints would be implemented but they're not).

src/protagonist/API/Converters/AdjunctConverter.cs
donaldgray: As discussed on Slack - would still need to switch on Uri.TryParse(), in which case there'd be no saved work. If trying short form initially then we'd use exceptions for code flow. Possible solution is to add AssetId.TryParse(), if worthwhile can do so in a follow up PR.

src/protagonist/DLCS.Repository.Tests/Messaging/EngineClientTests.cs
donaldgray: I thought they were all safe tbh, the `AssetIdGenerator.GetAssetId()` is important in places where we're saving to state (ie DB) shared between tests but I _think_ that all of these are isolated to tests or mocks

Issue #1111, opened 2026-02-18
Support `inlineAnnotation` adjuncts - Parent ticket


donaldgray:
Add a new `iiifLink` type: `inlineAnnotation`.

`iiifLink` type of `annotation` is currently supported, see #1073 and #1074 for implementation. The types `annotation` and `inlineAnnotation` are closely linked:
* `annotation` - used when the adjunct itself is an `AnnotationPage`, the link is then to a [`referenced`](https://iiif.io/api/presentation/3.0/#12-terminology) resource.
* `inlineAnnotation` - used when the the adjunct itself is the body of an `Annotation`. A platform managed `AnnotationPage` is added by the platform, effectively allowing an `embedded` resource.

## Annotation

Example of `annotation` type, for annotation named "my-annotation-page".

This will be rendered in final Manifest as:

```json
{
  "annotations": [
    {
        "id": "https://dlcs.example/adjunct/2/1/foo/my-annotation-page",
        "type": "AnnotationPage",
        "label": { "en": [ "Adjunct is an annotation page" ] }
    }]
}
```

Fetching `https://dlcs.example/adjunct/2/1/foo/my-annotation-page` should return an `AnnotationPage`:

```json
{
  "@context": "http://iiif.io/api/presentation/3/context.json",
  "id": "...",
  "type": "AnnotationPage",
  "items": [ ... ]
}
```

## InlineAnnotation

Example of `InlineAnnotation` type, for annotation named "my-annotation-page".

This will be rendered in final Manifest as:

```json
{
  "annotations": [
    {
        "id": "https://dlcs.example/adjunct-annotations/2/1/foo",
        "type": "AnnotationPage",
        "label": { "en": [ "Inline annotations" ] },
        "items": [
          {
            "id": "https://dlcs.example/adjunct-annotations/2/1/foo/my-annotation-page",
            "type": "Annotation",
            "motivation": [..],
            "body": [{
              "id": "https://dlcs.example/adjunct/2/1/foo/my-annotation-page",
              "type": "Text",
              "format": "text/plain",
              "label": { "en": [ "Adjunct is an inline annotation" ] },
            }]
          }
        ],
        "target": "https://dlcs.example/iiif-img/2/1/foo/canvas/c/1"
    }]
}
```

where the `AnnotationPage` is a "synthetic" annotation page added by the platform, with a non-dereferenceable id. The `body` of the annotation is constructed using the payload sent to API.

christinemacdonald:
[...]

PR #1176, opened 2026-04-17
RFC: Hosted Annotation Adjuncts `id` Property Handling

### LINKED ISSUE #1114: Ensure `"id"` for hosted `annotations` adjunct is correct
The need for this ticket arose when writing #1111 with the introduction of `inlineAnnotations`. 

We can have 2 related types of adjuncts: `annotation` and `inlineAnnotation`. The former is a referenced resource - the referenced resource, when fetched, should be an `AnnotationPage`. We don't verify this in the API or Engine level, all we do is enforce the API user specified `"@type": "AnnotationPage"` if the `iiifLink` is `annotation`.

When loading an external `"AnnotationPage"` the `"id"` value should reflect the path that it is being requested on. How do we do this?

Possibilities and their pros/cons are:

* Write the `"id"` value when saving the adjunct.
  * Pro: Orchestrator unchanged, serving adjunct is straightforward and will always have the same flow (ie we use YARP to proxy from S3).
  * Con: If s3-ambient origin then we cannot touch the adjunct, we just stream from source.
  * Con: If there are rewrite rules setup Engine will need to know about these, this doesn't seems like a concern of Engines.
  * Con: Related to above - if rewrite rules are setup, or changed, the saved `AnnotationPage` will only ever have 1 single `"id"` value. This in itself may not be an issue but if they were to change we'd need to bulk update.
* Write the `"id"` value when serving from Orchestrator.
  * Pro: The `"id"` can be updated on the fly so it will always reflect the path being requested, regardless of rewrite rules etc.
  * Con: This is a different approach to how Orchestrator will work for other adjuncts. We'd likely need to break away from serving via YARP. Orchestrator already does very similar for info.json requests but these are on completely different code paths from other 'serving' behaviours and info.json files are known to be small.
  * Con: We'd lose some of the 'nice' YARP behaviour like supporting range-requests. This may not be an issue.
  * Con: We can't guarantee that the adjunct is indeed an `AnnotationPage` so we'd need an exit strategy if it was something else.
    * Could we combat this by having a step in Engine that validates that an `annotation` is indeed an `AnnotationPage`?
    * Should we have a size limit on `annotations`?

Outcome:
* [x] Write up RFC on how we will approach this problem.

p-kaczynski:
As per #1114 an RFC documenting recommended approach to the described issue, as well as a "bonus" section that tackles question of Annotation Adjuncts JSON validation.

### DIFF
--- a/docs/rfcs/023-hosted-adjunct-id.md
+++ b/docs/rfcs/023-hosted-adjunct-id.md
@@ -3,13 +3,13 @@
 This RFC looks at how to ensure that the `id` property is correct, i.e. it represents the current, configuration-specific
 location (e.g. affected by path rewrite).
 
-This is specific to `iiifLink` of `annotation` and `inline-annotation`. The API user has specified that the
+This is specific to `iiifLink` of `annotation` only. The API user has specified that the
 adjunct has `"type": "AnnotationPage"`, and it is _expected_ that the adjunct binary content is a valid `AnnotationPage`.
 
 > [!NOTE]
 > There is currently no verification at any point that this content indeed conforms to the expectations
 
-# Proposed Solutions Overview
+## Proposed Solutions Overview
 
 Two contrasting solutions were proposed:
 
@@ -18,12 +18,12 @@ Two contrasting solutions were proposed:
 
 This RFC will argue for the option of `id` rewrite on-the-fly. This is primarily due to a need to handle scenarios with `s3-ambient`
 stored items. As those are immutable, and hence they ultimately require _some_ processing on the output, they make development effort
-for the solution `2` required in both cases, and therefore much more sensible asd a general solution.
+for the solution `2` required in both cases, and therefore much more sensible and a general solution.
 
 The main downsides that will likely need either addressing or at least observation is the performance hit that comes with
 output manipulation vs serving pre-modified content.
 
-# Implementation Overview
+## Implementation Overview
 
 When working on the `iiif-presentation` a similar scenario has been encountered before, and a `StreamingJsonProcessor` class
 has been created. It is a static implementation of the .NET `System.Text.Json`-based JSON processor which operates on
@@ -32,16 +32,22 @@ a stream, without e.g. loading entire contents into memory, allowing it to work 
 Similarities with previous scenario continue, as `iiif-presentation` also includes a "plugin" class for the
 aforementioned `StreamingJsonProcessor` specifically designed to modify the `id` property of a JSON document.
 
-Therefore, uplifting both the processor and the `id`-rewrite plugin into e.g. a dedicated NuGet package could allow us
+Therefore, uplifting both the processor and the `id`-rewrite plugin into Protagonist would allow us
 to completely reuse the code, speeding up resolution of this issue.
 
 Using this processor the remaining implementation steps are as follows:
 * Identify that the request is for an `annotation`/`inline-annotation` adjunct
 * Determine the correct `id` value
 * Obtain the S3 stream response within Orchestrator and stream into JSON processor
-* Stream the processor output to user response
+* Stream the processor output to user response, returning a 5xx exception should the binary be unprocessable.
 
-# Downsides and Considerations
+> [!NOTE]
+> Decision is to internalise `StreamingJsonProcessor`, rather than creating a dedicated NuGet package.
+>
+> The classes are relatively small and the additional overhead of maintenance (build and pipelines etc)> as deemed
+> to not be worth it. We can revisit in the future should circumstances change.
+
+## Downsides and Considerations
 
 Proposed approach is not free from downsides and trade-offs, the main one being that Orchestrator becomes part
 of the "content streaming pipeline", which despite optimizations already present in the `StreamingJsonProcessor` it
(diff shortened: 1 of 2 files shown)


docs/rfcs/023-hosted-adjunct-id.md
```diff
@@ -0,0 +1,91 @@
+# Annotation Adjuncts `id` Property
+
+This RFC looks at how to ensure that the `id` property is correct, i.e. it represents the current, configuration-specific
+location (e.g. affected by path rewrite).
+
+This is specific to `iiifLink` of `annotation` and `inline-annotation`. The API user has specified that the
```
donaldgray: It's only `annotation` as that references a 'full' `AnnotationPage`. 

`inline-annotation` types are embedded into the page in a "synthetic" `AnnotationPage` see #1111

docs/rfcs/023-hosted-adjunct-id.md
```diff
@@ -0,0 +1,91 @@
...
+# Proposed Solutions Overview
+
+Two contrasting solutions were proposed:
+
+1. Write the `id` value when saving the adjunct to S3
+2. Rewrite the `id` value when outputting requested adjunct
+
+This RFC will argue for the option of `id` rewrite on-the-fly. This is primarily due to a need to handle scenarios with `s3-ambient`
+stored items. As those are immutable, and hence they ultimately require _some_ processing on the output, they make development effort
+for the solution `2` required in both cases, and therefore much more sensible asd a general solution.
```
donaldgray: small typo, 
```suggestion
for the solution `2` required in both cases, and therefore much more sensible as a general solution.
```

docs/rfcs/023-hosted-adjunct-id.md
```diff
@@ -0,0 +1,91 @@
...
+After upload to S3 finishes (or it's determined to exist in `s3-ambient` bucket), byte length is
+obtained from S3. If it exceeds a globally configured value, the supposed annotation page JSON is
+not loaded, and if it's invalid it will have to be caught at a different point, possibly by the user.
+
+This in itself doesn't protect from the streaming processor abuse for Orchestrator, but rather protects Engine
+from potentially running out of memory or just having degraded performance.
+
+## Size Limits Summary
+
+Both of the aforementioned limits could be set, a smaller one that limits "Engine validates IIIF AnnotationPage JSON" step,
```
donaldgray, line 95: I think it makes sense to validate in both places. What would be a reasonable limit? Is it 1MB, 5MB, 10MB? Sensible route would be to make this limit configurable and move up/down depending on findings but it'd be good to get a starting idea.

I think it can err towards the larger size in some cases - we have CDN and varnish sitting infront of Orchestrator so those should be able to take at least some of the load off of repeated requests for larger adjuncts.

`s3-ambient` may be more difficult - if the origin is `s3://my-bucket/anno/whatever.json` - we'd have no idea if the file is changed after being ingested, so we may just need to be safe/cautious in Orchestrator. Stream the first XX bytes and abort if it's larger or something.

docs/rfcs/023-hosted-adjunct-id.md
```diff
@@ -0,0 +1,91 @@
...
+
+When working on the `iiif-presentation` a similar scenario has been encountered before, and a `StreamingJsonProcessor` class
+has been created. It is a static implementation of the .NET `System.Text.Json`-based JSON processor which operates on
+a stream, without e.g. loading entire contents into memory, allowing it to work on arbitrarily large JSON documents.
+
+Similarities with previous scenario continue, as `iiif-presentation` also includes a "plugin" class for the
+aforementioned `StreamingJsonProcessor` specifically designed to modify the `id` property of a JSON document.
+
+Therefore, uplifting both the processor and the `id`-rewrite plugin into e.g. a dedicated NuGet package could allow us
+to completely reuse the code, speeding up resolution of this issue.
```
donaldgray, line 36: Does it warrant a nuget package, or can it be embedded? Pros/cons - ease of maintenance vs risk of drift etc

docs/rfcs/023-hosted-adjunct-id.md
```diff
@@ -0,0 +1,91 @@
...
+
+Using this processor the remaining implementation steps are as follows:
+* Identify that the request is for an `annotation`/`inline-annotation` adjunct
+* Determine the correct `id` value
+* Obtain the S3 stream response within Orchestrator and stream into JSON processor
+* Stream the processor output to user response
```
donaldgray: What will happen if it encounters non-JSON (say XML)? Will it throw an obvious exception that we could catch and transform to an appropriate status code (500 or 502 maybe)

donaldgray (commented):
LGTM, aside from a typo couple of observations, rather than anything wrong or missing.

docs/rfcs/023-hosted-adjunct-id.md
donaldgray: Validation at Engine stage has been done, see https://github.com/dlcs/protagonist/issues/1195

docs/rfcs/023-hosted-adjunct-id.md
donaldgray: #1194 work has started, decision was not to use a nuget package and embed code instead as it's only a few classes. The trade off of additional build pipeline etc wasn't worth it. Can reevaluate at a later date if this isn't the case.

donaldgray (approved):

PR #1178, opened 2026-04-17
UpdateHandlers abort if item not in DB

donaldgray:
## What does this change?

Alter `AdjunctUpdateHandler` and `AssetUpdateHandler` to return `true` rather than `false`. By returning `false` the item is requeued and then moved to DLQ, whereas returning `true` marks the message as handled.

This can only happen if an item is Updated then Deleted prior to being picked up by Updated handler. In this case the `*DeleteHandler` will have handled removing derivatives.


p-kaczynski (approved):
That makes sense to me

PR #1187, opened 2026-04-29
Set certain fields to optional for the 'None' channel

### LINKED ISSUE #1155: 'none' channel makes some fields optional
An asset with `none` delivery-channel is not ingested by Engine - it is effectively a placeholder. To make creation of `none` assets easier (e.g. for "stub" assets), assets with `none` delivery-channel should have slightly different validation logic. "origin" and "mediaType" are optional and given a placeholder if omitted.

* [x] `"origin"` is optional, if `none` delivery-channel. Default to: `"https://example.org/origin"`
* [x] An "origin" of  `"https://example.org/origin"` is invalid for any other delivery-channels
* [x] `"mediaType"` is optional, if `none` delivery-channel. Default to: "example/example" (uses valid [`example/`](https://www.iana.org/assignments/media-types/media-types.xhtml#example) type)
* [x] An "mediaType" of  `"example/example"` is invalid for any other delivery-channels
* [x] Legacy mode should continue to require both `"origin"` and `"mediaType"` as they can't supply delivery-channel.

### LINKED ISSUE #1182: BUG: 500 returned when we send invalid parameter names in member body
When we send a payload with invalid properties such as:

```json
{
  "member": [
    {
      "idx": "1/1/foo",
      "adjunctx": [
        "ocr.txt",
        "mets.xml"
      ]
    }
  ]
}
```

Then we receive a 500 error response. Should be a 400.

```json
{
  "@type": "Error",
  "statusCode": 500,
  "status": 500,
  "title": "Unexpected error",
  "detail": "Object reference not set to an instance of an object.",
  "instance": "https://api.dlcs.digirati.io",
  "description": "Object reference not set to an instance of an object."
}
```

JackLewis-digirati:
## What does this change?

Resolves #1155
Resolves #1182

This PR modifies the `none` channel to make the `mediaType` and `origin` optional when specified.  

It will also update the origin to be `https://example.org/origin` and the mediatype to `example/example` if they haven't been set.

However, if the `none` channel isn't specified, the code will act the same way, outside of activwely rejecting any asset qwith the placeholder values set

Additionally, the bug around returning `500` when deleting adjuncts with incorrect fields has been fixed by adding`[ApiController]` to the `CustomerAdjunctController` class, so it should now return a `400` response

### DIFF
--- a/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
+++ b/src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
@@ -345,7 +345,7 @@ public class HydraImageValidatorTests
     [InlineData(null)]
     [InlineData("")]
     [InlineData(" ")]
-    public void MediaType_NullOrEmpty_OnCreateWhenNoneDeliveryChannel(string mediaType)
+    public void MediaType_CanBeNullOrEmpty_OnCreateWhenNoneDeliveryChannel(string mediaType)
     {
         var model = new Image
         {
@@ -354,5 +354,33 @@ public class HydraImageValidatorTests
         };
         var result = Sut.TestValidate(model, options => options.IncludeRuleSets("default", "create"));
         result.ShouldNotHaveValidationErrorFor(a => a.MediaType);
+    }
+    
+    [Fact]
+    public void PlaceHolderValues_NotAllowed_WhenNotNoneChannel()
+    {
+        var model = new Image
+        {
+            MediaType = AssetDeliveryChannels.NoneChannelMediaTypePlaceholder,
+            Origin = AssetDeliveryChannels.NoneChannelOriginPlaceholder,
+            DeliveryChannels = [new DeliveryChannel {Channel = AssetDeliveryChannels.Image }]
+        };
+        var result = Sut.TestValidate(model, options => options.IncludeRuleSets("default"));
+        result.ShouldHaveValidationErrorFor(a => a.MediaType).WithErrorMessage("'example/example' is not a valid mediaType");
+        result.ShouldHaveValidationErrorFor(a => a.Origin).WithErrorMessage("'https://example.org/origin' is not a valid origin");
+    }
+    
+    [Fact]
+    public void PlaceHolderValues_Allowed_WhenNoneChannel()
+    {
+        var model = new Image
+        {
+            MediaType = AssetDeliveryChannels.NoneChannelMediaTypePlaceholder,
+            Origin = AssetDeliveryChannels.NoneChannelOriginPlaceholder,
+            DeliveryChannels = [new DeliveryChannel {Channel = AssetDeliveryChannels.None }]
+        };
+        var result = Sut.TestValidate(model, options => options.IncludeRuleSets("default"));
+        result.ShouldNotHaveValidationErrorFor(a => a.MediaType);
+        result.ShouldNotHaveValidationErrorFor(a => a.Origin);
     }
 }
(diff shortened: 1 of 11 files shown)


src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
```diff
@@ -68,6 +72,21 @@ public static bool HasSingleDeliveryChannel(this Asset asset, string deliveryCha
     public static bool DoesNotHaveDeliveryChannel(this Asset asset, string deliveryChannel)
         => !asset.HasDeliveryChannel(deliveryChannel);
     
+    /// <summary>
+    /// Checks if all specified channels are the 'none' channel (and at least one is present)
```
donaldgray: The _and at least one is present_ isn't correct here. See https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/delivery-channels#the-none-channel

> This resource is only valid on asset.deliveryChannels, **and only when it is the one and only member.**

'none' is only valid when it is the only member. Multiple 'none' is invalid.

Also missing unit tests for this, other `AssetDeliveryChannel` methods have tests.

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -56,6 +68,9 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
         RuleFor(a => a.Finished).Empty().WithMessage("Should not include finished");
         RuleFor(a => a.Created).Empty().WithMessage("Should not include created");
     }
+    
+    static bool IsNoneOnly(DLCS.HydraModel.Image image)
```
donaldgray: Missing access modifier.

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -20,9 +20,21 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
         
         RuleSet("create", () =>
         {
-            RuleFor(a => a.MediaType).NotEmpty().WithMessage("Media type must be specified");
+            RuleFor(a => a.MediaType).NotEmpty()
```
donaldgray: Small, new line would make it more readable/keep it consistent with below methods.

```suggestion
            RuleFor(a => a.MediaType)
                .NotEmpty()
```

src/protagonist/API.Tests/Features/Images/Validation/HydraImageValidatorTests.cs
```diff
@@ -339,4 +340,19 @@ public void MaxWidth_CanBeMaxSize()
         var result = Sut.TestValidate(model);
         result.ShouldNotHaveValidationErrorFor(a => a.MaxWidth);
     }
+    
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    [InlineData(" ")]
+    public void MediaType_NullOrEmpty_OnCreateWhenNoneDeliveryChannel(string mediaType)
```
donaldgray: Should this have something in title to signify the expectation that there's no error - my assumption is that these are verifying invalid data? `MediaType_CanBeNullOrEmpty_OnCreateWhenNoneDeliveryChannel` looks to be how other tests do it?

src/protagonist/API/Features/Image/Validation/HydraImageValidator.cs
```diff
@@ -20,9 +20,21 @@ public HydraImageValidator(IOptions<ApiSettings> apiSettings)
...
+                .When(a => !IsNoneOnly(a))
+                .WithMessage("Media type must be specified");
         });
-        
+
+        RuleFor(a => a.Origin)
```
donaldgray, line 29: Missing tests for this + similar rule for MediaType.

donaldgray (changes requested):
LGTM, only real required change is the `IsNoneOnly` logic handling multiple 'none' channels. Other comments are mostly around missing tests.

donaldgray (approved):

PR #1191, opened 2026-05-01
Allow for stub assets to be added to space 0

### LINKED ISSUE #1186: Stub asset/space 0 handling
See RFC for details on stub assets: https://github.com/dlcs/protagonist/blob/develop/docs/rfcs/022-stub-assets.md

As outlined in RFC, all stub assets are in space 0 and must use the ['none'](https://dlcs.github.io/public-docs/api-doc/delivery-channels/#the-none-channel) delivery-channel. This ticket is to tighten up some rules around how space 0 is used, to make it easier to use stub assets:

* [ ] "None" delivery channel is the only accepted type for assets in space 0
* [ ] DefaultDeliveryChannels should be configured in such a way that space 0 uses "None" by default. New and existing customers should receive these.
* [ ] Ensure that all existing customers have a space 0
* [ ] Ensure that new customers receive a space 0
* [ ] Prevent deletion of space 0
* [ ] Do not return space 0 in API space listing responses
* [ ] (_Stretch goal_) Prevent changing DefaultDeliveryChannel for space 0, to ensure it stays as "None" only.

See ticket #1152 for related changes to how "none" channel assets are processed.

JackLewis-digirati:
## What does this change?

Resolves #1186

Allows space 0 to be used for stub assets, while disallowing the use of space 0 for any delivery channel which isn't `none`.  This additionally adds a migration to make sure default delivery channels will automatically specify the `none` channel if space 0 is selected.

Additionally, due to space now being `null` for setting global delivery channels, there's essentially a new set of endpoints for `DefaultDeliverChannel` management to omit the space.

| Operation | Route |                                                                                                                                                                                                                                                                                                                                                                                                          
  |---|---|                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
  | Get one | `GET /customers/{id}/defaultDeliveryChannels/{ddcId}` |
  | Create | `POST /customers/{id}/defaultDeliveryChannels` |                                                                                                                                                                                                                                                                                                                                                                    
  | Update | `PUT /customers/{id}/defaultDeliveryChannels/{ddcId}` |
  | Delete | `DELETE /customers/{id}/defaultDeliveryChannels/{ddcId}` |

> [!Note]
> These endpoints work the exact same way as the previous endpoints, and go through the same code as the previous space-based endpoints

### Requirements

| # | Requirement | Covered? | Where | Tests |                                                                                                                                                                                                                                                                                                                                                                                 
  |---|---|---|---|---|      
  | 1 | `none` is the only accepted delivery channel for space 0 assets | ✅ | `CreateOrUpdateImage.cs`, `CreateBatchOfImages.cs` | `Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400`, `Put_Asset_InSpaceZero_WithNoneDeliveryChannel_CreatesAsset`, `Post_CreateBatch_400_IfSpaceZeroAssetHasNonNoneDeliveryChannel` |                                                                                              ─
  | 2 | Space 0 uses `none` DDC by default; new and existing customers receive it | ✅ | Migration (`SpaceZeroHandling`), `DapperNewCustomerDeliveryChannelRepository` | `CreateNewCustomer_CreatesSpaceZero`, `Post_CreateDefaultDeliveryChannel_ForSpaceZero_Returns400` (verifies row exists) |
  | 3 | All existing customers have a space 0 | ✅ | Migration (`SpaceZeroHandling`) | Covered by migration; `CreateNewCustomer_CreatesSpaceZero` validates new customer path |                                                                                                                                                                                                                                                  ─
  | 4 | New customers receive a space 0 | ✅ | `CreateCustomer.cs` | `CreateNewCustomer_CreatesSpaceZero` |
  | 5 | Prevent deletion of space 0 | ✅ | `DeleteSpace.cs` | `DeleteSpace_Returns_Conflict_ForSpaceZero` |
  | 6 | Space 0 not returned in API space listings | ✅ | `SpaceRepository.GetPageOfSpaces` | `GetSpaces_DoesNotReturn_SpaceZero` |
  | 7 | _(Stretch)_ Prevent changing DDCs for space 0 | ✅ | `DefaultDeliveryChannelHelper` | `Post_CreateDefaultDeliveryChannel_ForSpaceZero_Returns400`, `Put_UpdateDefaultDeliveryChannel_ForSpaceZero_Returns400`, `Delete_DeleteDefaultDeliveryChannel_ForSpaceZero_Returns409`, `GetSpaceZeroErrorMessage_ReturnsError_WhenSpaceIsZero` |

## Database Migration

> [!NOTE]
> - **Migration name:** `20260501110241_SpaceZeroHandling`
> - **What it does:**
>   - Makes `space` nullable in the `DefaultDeliveryChannel` table
>   - Modifies the "global" space in the `DefaultDeliveryChannel` table to be `null` instead of `0`
>   - Adds space `0` to all existing customers
>   - Adds a new template `DefaultDeliveryChannel` that sets the `none` channel if space `0` is specified (used for new customers)
>   - Updates all existing customers with the new `DefaultDeliveryChannel`
> - **Breaking Change?** Yes - would break API in this case

### DIFF
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -207,7 +207,7 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
     {
         // Arrange
         var customerAndSpace = await CreateCustomerAndSpace();
-        var assetId = new AssetId(customerAndSpace.customer, 0, nameof(Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400));
+        var assetId = AssetIdGenerator.GetAssetId(customerAndSpace.customer, 0);
         var hydraImageBody = $@"{{
             ""@type"": ""Image"",
             ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -255,6 +255,44 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
         var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
         asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == AssetDeliveryChannels.None);
     }
+    
+    [Fact]
+    public async Task Put_UpdateAsset_InSpaceZero_WithNoSpecifiedDeliveryChannel_UpdatesAssetWithNoneChannelSuccessfully()
+    {
+        // Arrange
+        var customerAndSpace = await CreateCustomerAndSpace();
+        var assetId = AssetIdGenerator.GetAssetId(customer: customerAndSpace.customer, space: AssetDeliveryChannels.StubAssetSpace);
+    
+        var noneDeliveryChannel = new List<ImageDeliveryChannel>
+        {
+            new()
+            {
+                Channel = AssetDeliveryChannels.None,
+                DeliveryChannelPolicyId = KnownDeliveryChannelPolicies.None
+            }
+        };
+        
+        await dbContext.Images.AddTestAsset(assetId, customer: customerAndSpace.customer, space: AssetDeliveryChannels.StubAssetSpace, 
+            origin: $"https://example.org/{assetId.Asset}.tiff", imageDeliveryChannels: noneDeliveryChannel);
+        
+        await dbContext.SaveChangesAsync();
+        
+        var hydraImageBody = $@"{{
+            ""@type"": ""Image"",
+            ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
+            ""family"": ""I"",
+            ""mediaType"": ""image/tiff""
+        }}";
+
+        // Act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customerAndSpace.customer).PutAsync(assetId.ToApiResourcePath(), content);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
+        asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == AssetDeliveryChannels.None);
+    }
 
     [Fact]
     public async Task Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToDefault()
--- a/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
+++ b/src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
@@ -203,6 +203,60 @@ public class ModifyAssetTests : IClassFixture<ProtagonistAppFactory<Startup>>
     }
     
     [Fact]
+    public async Task Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400()
+    {
+        // Arrange
+        var customerAndSpace = await CreateCustomerAndSpace();
+        var assetId = new AssetId(customerAndSpace.customer, 0, nameof(Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400));
+        var hydraImageBody = $@"{{
+            ""@type"": ""Image"",
+            ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
+            ""family"": ""I"",
+            ""mediaType"": ""image/tiff"",
+            ""deliveryChannels"": [
+            {{
+                ""channel"": ""iiif-img"",
+                ""policy"": ""default""
+            }}]
+        }}";
+
+        // Act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customerAndSpace.customer).PutAsync(assetId.ToApiResourcePath(), content);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+    }
+
+    [Fact]
+    public async Task Put_Asset_InSpaceZero_WithNoneDeliveryChannel_CreatesAsset()
+    {
+        // Arrange
+        var customerAndSpace = await CreateCustomerAndSpace();
+        var assetId = new AssetId(customerAndSpace.customer, 0, nameof(Put_Asset_InSpaceZero_WithNoneDeliveryChannel_CreatesAsset));
+        var hydraImageBody = $@"{{
+            ""@type"": ""Image"",
+            ""origin"": ""https://example.org/{assetId.Asset}.tiff"",
+            ""family"": ""I"",
+            ""mediaType"": ""image/tiff"",
+            ""deliveryChannels"": [
+            {{
+                ""channel"": ""none"",
+                ""policy"": ""none""
+            }}]
+        }}";
+
+        // Act
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var response = await httpClient.AsCustomer(customerAndSpace.customer).PutAsync(assetId.ToApiResourcePath(), content);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.Created);
+        var asset = dbContext.Images.Include(i => i.ImageDeliveryChannels).Single(x => x.Id == assetId);
+        asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == AssetDeliveryChannels.None);
+    }
+
+    [Fact]
     public async Task Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToDefault()
     {
         var customerAndSpace = await CreateCustomerAndSpace();
--- a/src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
+++ b/src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
@@ -102,7 +102,7 @@ public class CreateCustomerHandler : IRequestHandler<CreateCustomer,  ModifyEnti
             {
                 Customer = newCustomerId,
                 Id = 0,
-                Name = "stub-space",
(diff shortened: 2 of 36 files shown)


src/protagonist/API/Features/DeliveryChannels/Requests/DefaultDeliveryChannels/CreateDefaultDeliveryChannel.cs
```diff
@@ -72,19 +75,34 @@ public async Task<ModifyEntityResult<DefaultDeliveryChannel>> Handle(
             return ModifyEntityResult<DefaultDeliveryChannel>.Failure("Failed to find linked delivery channel policy", WriteResult.BadRequest);
         }
 
+        var space = request.Space;
+        var duplicate = await dbContext.DefaultDeliveryChannels.AnyAsync(
```
JackLewis-digirati, line 82: This isn't as nice as it worked previously.  However, the issue is that postgres effectively treats `NULL != NULL` for unique indexes, which means saving to the database doesn't throw an error like it did previously.

There is a nicer way of doing this in [PG15+](https://stackoverflow.com/questions/8289100/create-unique-constraint-with-null-columns) where we can set `NULLS NOT DISTINCT` on unique indexes, but this would mean more changes as we support PG14

JackLewis-digirati:
### Core concept                                                                                                                                                                                                                                                                                                                                                                                                              
Introduces **space 0** as a reserved "stub space" per customer that can hold assets using only the
`none` delivery channel. Simultaneously, the representation of **customer-wide default delivery                                                                                                                                                                                                                                                                                                                               
channels** shifts from `Space = 0` to `Space = NULL` at the database level.

---

### Database migration (`SpaceZeroHandling`)
- Makes `DefaultDeliveryChannels.Space` nullable (`int` → `int?`)
- Migrates existing `Space = 0` rows (old customer-wide defaults) to `Space = NULL`
- Creates a `Space = 0` stub-space entry in the `Spaces` table for all existing customers
- Seeds a `*/*` → `none` default delivery channel at `Space = 0` for all customers (the policy
  that prevents stub assets from being processed)

### Model changes
- `DefaultDeliveryChannel.Space` changed from `int` to `int?`
- `DlcsContext` config and model snapshot updated to reflect nullable `Space`
- `HydraModel.DefaultDeliveryChannel`: URL construction now treats `Space = null` (not `Space = 0`)
  as the customer-level route

### Controller / request handlers
All `DefaultDeliveryChannels` endpoints change the `space` route parameter default from
`int space = 0` to `int? space = null`, so requests without a space segment correctly represent
customer-wide defaults.

**Space 0 guard** added to Create, Update, and Delete handlers via a shared
`DefaultDeliveryChannelHelper.GetSpaceZeroErrorMessage` — returning 400/409 to prevent direct
manipulation of the stub-space delivery channel.

**Duplicate detection** in Create now does an explicit pre-check rather than relying solely on the
DB unique constraint (which doesn't catch `NULL = NULL` in PostgreSQL 14).

### Customer creation
When a new customer is created, a `Space = 0` stub-space row is also created for them.

### New customer delivery channel seeding (Dapper)
The SQL that copies delivery channels to new customers is updated to copy `Space = NULL` rows
(previously it only copied `Space = 0`) and preserves the original `Space` value rather than
hardcoding `0`.

### Asset / batch validation
Assets in space 0 are restricted to the `none` delivery channel only — enforced in both
single-asset (`CreateOrUpdateImage`) and batch (`CreateBatchOfImages`) write paths.

### Space repository
`GetPageOfSpaces` now excludes `Id = 0` from results and counts, so the stub space is invisible
to normal space listings.

### Space deletion
`DeleteSpace` now returns 409 Conflict if the caller attempts to delete space 0.

### delivery channel handling
Special case added to allow the stub asset space to work when it would normally not be allowed due to global channels nominally being selected for the none space

### Asset processor
Small change to logic so that none to none asset updates do not try to force reingestion

src/protagonist/API/Features/Image/Requests/CreateOrUpdateImage.cs
```diff
@@ -76,6 +76,15 @@ public async Task<ModifyEntityResult<Asset>> Handle(CreateOrUpdateImage request,
                 WriteResult.FailedValidation);
         }
 
+        if (!AssetDeliveryChannels.IsValidForSpaceZero(
+                assetBeforeProcessing.Asset.Space,
+                assetBeforeProcessing.DeliveryChannelsBeforeProcessing?.Select(dc => dc.Channel)))
+        {
+            return ModifyEntityResult<Asset>.Failure(
+                "Assets in space 0 must use the 'none' delivery channel only",
```
donaldgray:
```suggestion
                "Assets in space 0 can only use the 'none' delivery channel",
```

src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
```diff
@@ -83,6 +85,13 @@ public static bool IsNoneOnly(IEnumerable<string>? channels)
         return e.MoveNext() && e.Current == None && !e.MoveNext();
     }
 
+    /// <summary>
+    /// Returns true if the asset's delivery channels are valid for space 0 (only 'none' allowed),
+    /// or if the asset is not in space 0.
+    /// </summary>
+    public static bool IsValidForSpaceZero(int space, IEnumerable<string>? channels)
```
donaldgray: I think this would read clearer as an API specific method that operates on `AssetBeforeProcessing` as that's what it's actually operating on. Usages would then be

```cs
// this
var hasInvalidSpace0Asset = request.AssetsBeforeProcessing.Any(a => a.IsValidForSpaceZero());

// rather than
var hasInvalidSpace0Asset = request.AssetsBeforeProcessing.Any(a =>
    !AssetDeliveryChannels.IsValidForSpaceZero(
        a.Asset.Space,
        a.DeliveryChannelsBeforeProcessing?.Select(dc => dc.Channel)));
```

src/protagonist/API/Features/Queues/Requests/CreateBatchOfImages.cs
```diff
@@ -151,6 +154,20 @@ await ingestNotificationSender.SendIngestAssetsRequest(engineNotificationList, r
         return ModifyEntityResult<Batch>.Success(batch, WriteResult.Created);
     }
 
+    private static ModifyEntityResult<Batch>? ValidateSpace0DeliveryChannels(CreateBatchOfImages request)
```
donaldgray: Could/should this be shared with equivalent in `CreateOrUpdateImage` - single place to do check + single place to store error message.

src/protagonist/API/Features/DeliveryChannels/Helpers/DefaultDeliveryChannelHelper.cs
```diff
@@ -0,0 +1,13 @@
...
+
+public enum SpaceZeroOperation { Create, Modify, Delete }
+
+public static class DefaultDeliveryChannelHelper
+{
+    /// <summary>
+    /// Returns an error message if <paramref name="space"/> is 0, otherwise null.
+    /// The <paramref name="operation"/> controls which message is returned.
+    /// </summary>
+    public static string? GetSpaceZeroErrorMessage(int? space, SpaceZeroOperation operation)
```
donaldgray: Small - should this just be `"Default delivery channels for space 0 cannot be changed"`, seems to capture the error succinctly, also some permutations don't read right:

* `"Default delivery channels for space 0 cannot be changed for the create operation"`
* `"Default delivery channels for space 0 cannot be changed for the modify operation"`
* `"Default delivery channels for space 0 cannot be changed for the delete operation"`

src/protagonist/API/Features/DeliveryChannels/DataAccess/DefaultDeliveryChannelRepository.cs
```diff
@@ -31,9 +32,17 @@ public DefaultDeliveryChannelRepository(IAppCache appCache,
 
     public async Task<List<DeliveryChannelPolicy>> MatchedDeliveryChannels(string mediaType, int space, int customerId)
     {
-        var completedMatch = new List<DeliveryChannelPolicy>();
```
donaldgray: Small - additional whitespace

src/protagonist/API/Features/DeliveryChannels/DataAccess/DefaultDeliveryChannelRepository.cs
```diff
@@ -31,9 +32,17 @@ public DefaultDeliveryChannelRepository(IAppCache appCache,
...
         
         var orderedDefaultDeliveryChannels = await OrderedDefaultDeliveryChannels(space, customerId);
+
+        // this is the stub asset space, which is handled differently
```
donaldgray: Small - worth expanding how this is different? (I think because it doesn't need to look at mediaType?)

src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
```diff
@@ -96,6 +97,19 @@ await dbContext.Queues.AddRangeAsync(
                 new Queue { Customer = newCustomerId, Name = QueueNames.Priority, Size = 0 }
             );
 
+            // Create space 0 for stub assets
+            await dbContext.Spaces.AddAsync(new DLCS.Model.Spaces.Space
+            {
+                Customer = newCustomerId,
+                Id = 0,
+                Name = "stub-space",
```
donaldgray: I'm not sure if this would be surfaced anywhere but would `stub-assets` be a clearer indication of what this is used for?

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -202,6 +202,98 @@ public async Task Put_NewImageAsset_Creates_Asset_WithDeliveryChannelsSetToNone(
         imageStorage.ThumbnailSize.Should().Be(0);
     }
     
+    [Fact]
+    public async Task Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400()
+    {
+        // Arrange
+        var customerAndSpace = await CreateCustomerAndSpace();
+        var assetId = new AssetId(customerAndSpace.customer, 0, nameof(Put_Asset_InSpaceZero_WithNonNoneDeliveryChannel_Returns400));
```
donaldgray:
```suggestion
        var assetId = AssetIdGenerator.GetAssetId(customerAndSpace.customer, 0);
```

donaldgray (approved):
LGTM

PR #1193, opened 2026-05-08
Allow create batch of images to work with a minimal payload in the stub asset space

JackLewis-digirati:
## What does this change?

This PR fixes some issues identified with `createBatchOfImages` to work with a minimal payload in the stub asset space, which is used by iiif-presentation for manifest level adjuncts

Original PR here - #1191

### DIFF
--- a/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
+++ b/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
@@ -1972,6 +1972,7 @@ public class CustomerQueueTests : IClassFixture<ProtagonistAppFactory<Startup>>
         const int customerId = 1901;
         var assetId = AssetIdGenerator.GetAssetId(customerId, 0);
         await dbContext.Customers.AddTestCustomer(customerId);
+        await dbContext.DefaultDeliveryChannels.AddTestDefaultDeliveryChannels(customerId);
         await dbContext.Spaces.AddTestSpace(customerId, 0, "stub-assets");
         await dbContext.CustomerStorages.AddTestCustomerStorage(customerId);
         await dbContext.Queues.AddAsync(new Queue { Customer = customerId, Name = "default", Size = 0 });
--- a/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
+++ b/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
@@ -2019,4 +2019,36 @@ public class CustomerQueueTests : IClassFixture<ProtagonistAppFactory<Startup>>
                     A<CancellationToken>._))
             .MustHaveHappened();
     }
+
+    [Fact]
+    public async Task Post_CreateBatch_400_IfSpaceNotProvided()
+    {
+        // Arrange
+        const int customerId = 1902;
+        await dbContext.Customers.AddTestCustomer(customerId);
+        await dbContext.CustomerStorages.AddTestCustomerStorage(customerId);
+        await dbContext.Queues.AddAsync(new Queue { Customer = customerId, Name = "default", Size = 0 });
+        await dbContext.SaveChangesAsync();
+
+        var hydraImageBody = """
+        {
+            "@context": "http://www.w3.org/ns/hydra/context.jsonld",
+            "@type": "Collection",
+            "member": [
+                {
+                  "id": "my-asset"
+                }
+            ]
+        }
+        """;
+
+        var content = new StringContent(hydraImageBody, Encoding.UTF8, "application/json");
+        var path = $"/customers/{customerId}/queue";
+
+        // Act
+        var response = await httpClient.AsCustomer(customerId).PostAsync(path, content);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+    }
 }
--- a/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
+++ b/src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
@@ -2025,12 +2025,6 @@ public class CustomerQueueTests : IClassFixture<ProtagonistAppFactory<Startup>>
     public async Task Post_CreateBatch_400_IfSpaceNotProvided()
     {
         // Arrange
-        const int customerId = 1902;
-        await dbContext.Customers.AddTestCustomer(customerId);
-        await dbContext.CustomerStorages.AddTestCustomerStorage(customerId);
-        await dbContext.Queues.AddAsync(new Queue { Customer = customerId, Name = "default", Size = 0 });
-        await dbContext.SaveChangesAsync();
-
         var hydraImageBody = """
         {
             "@context": "http://www.w3.org/ns/hydra/context.jsonld",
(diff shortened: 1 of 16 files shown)


src/protagonist/API/Features/Queues/Validation/QueuePostValidator.cs
```diff
@@ -33,15 +34,20 @@ public QueuePostValidator(IOptions<ApiSettings> apiSettings)
             .Must(m => (m?.Length ?? 0) <= maxBatch)
             .WithMessage($"Maximum assets in single batch is {maxBatch}");
 
-        RuleForEach(c => c.Members).SetValidator(new HydraImageValidator(apiSettings),
-            "default", "create");
-        
+        RuleForEach(c => c.Members).SetValidator(new HydraImageValidator(apiSettings), "default");
```
donaldgray: Why is `"create"` being dropped here?

src/protagonist/API/Features/Queues/CustomerQueueController.cs
```diff
@@ -418,9 +418,19 @@ private static List<AssetBeforeProcessing> CreateAssetsBeforeProcessing(int cust
...
         var assetsBeforeProcessing = images.Members!
             .Select(i => new AssetBeforeProcessing(i.ToDlcsModel(customerId),
-                i.DeliveryChannels.ToInterimModel())).ToList();
+                GetDeliveryChannelsForAsset(i))).ToList();
         return assetsBeforeProcessing;
     }
+
+    private static DeliveryChannelsBeforeProcessing[]? GetDeliveryChannelsForAsset(DLCS.HydraModel.Image image)
+    {
+        // Space 0 assets with no delivery channels are stub assets - treat as 'none' channel implicitly
```
donaldgray: Why is this needed here? This feels like it should be further down the chain? i.e. the API just receives the payload and passes it down to the request handler to worry about what the actual meaning is. We don't have any other special handling when delivery-channels aren't provided.

Where's the equivalent for this in single-asset handling (e.g. via PUT)? Can they/should they have shared logic?

src/protagonist/API/Converters/AssetConverter.cs
```diff
@@ -260,17 +260,15 @@ private static Asset GetAssetWithIdentifiers(Image hydraImage, int customerId, i
         {
             throw new BadRequestException("Asserted space does not agree with supplied space.");
         }
-        
-        if (hydraImage.Space <= 0)
+
+        if (hydraImage.Space == 0 && spaceId.HasValue)
```
donaldgray: As discussed on slack - need to differentiate "space not provided" and "space defaulted to 0"

donaldgray (commented):
I think it's worth adding a link to the original issue too, for tracking?

src/protagonist/API/Features/Queues/Validation/QueuePostValidator.cs
JackLewis-digirati: unintended duplication - will replace this back

src/protagonist/API/Features/Queues/CustomerQueueController.cs
JackLewis-digirati: yeah, makes sense - will move this check so it's shared with the single image and further down the process

src/protagonist/API.Tests/Integration/CustomerQueueTests.cs
```diff
@@ -1964,4 +1964,92 @@ public async Task Post_CreateBatch_400_IfSpaceZeroAssetHasNonNoneDeliveryChannel
...
+                    A<Batch>.That.Matches(b => b.Id == model.Id.GetLastPathElementAsInt()),
+                    A<CancellationToken>._))
+            .MustHaveHappened();
+    }
+
+    [Fact]
+    public async Task Post_CreateBatch_400_IfSpaceNotProvided()
+    {
+        // Arrange
+        const int customerId = 1902;
```
donaldgray: Small but you should be able to remove this DB setup and just use the default customer as it should get rejected up front.

src/protagonist/API.Tests/Integration/ModifyAssetTests.cs
```diff
@@ -256,6 +256,26 @@ public async Task Put_Asset_InSpaceZero_WithNoneDeliveryChannel_CreatesAsset()
         asset.ImageDeliveryChannels.Should().ContainSingle(x => x.Channel == AssetDeliveryChannels.None);
     }
     
+    [Fact]
+    public async Task Put_CreateAsset_InSpaceZero_WithNoDeliveryChannel_CreatesAssetWithNoneChannelAndPlaceholders()
+    {
+        // Arrange - minimal stub asset payload: no origin, no mediaType, no deliveryChannels
+        var customerAndSpace = await CreateCustomerAndSpace();
```
donaldgray: Is it worth using the default customer for this test? They'll have space 0 created by default and saves a couple of http calls to set up.

src/protagonist/API/Features/Image/Ingest/AssetProcessor.cs
```diff
@@ -82,7 +82,9 @@ public async Task<ProcessAssetResult> Process(AssetBeforeProcessing assetBeforeP
 
             var existingAsset = assetFromDatabase?.Clone();
             var isNoneChannel = AssetDeliveryChannels.IsNoneOnly(
-                assetBeforeProcessing.DeliveryChannelsBeforeProcessing?.Select(dcp => dcp.Channel));
+                assetBeforeProcessing.DeliveryChannelsBeforeProcessing?.Select(dcp => dcp.Channel))
```
donaldgray: Storing `assetBeforeProcessing.DeliveryChannelsBeforeProcessing?.Select(dcp => dcp.Channel)` as a variable would avoid iterating twice and make this easier to read.

src/protagonist/API/Features/Queues/CustomerQueueController.cs
```diff
@@ -435,9 +435,9 @@ private void UpdateMembers(int customerId, IList<DLCS.HydraModel.Image>? members
         {
             for (int i = 0; i < members.Count; i++)
             {
-                if (Settings.LegacyModeEnabledForSpace(customerId, members[i].Space))
+                if (Settings.LegacyModeEnabledForSpace(customerId, members[i].Space.GetValueOrDefault()))
```
donaldgray: Extract `members[i].Space.GetValueOrDefault()` to var?

General comment on `.GetValueOrDefault()` - I don't think it's used elsewhere in the codebase, everywhere else uses `?? 0` null-coalescing operator in this sort of scenario but it's used throughout this PR. Is that intentional?

src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
```diff
@@ -80,11 +80,17 @@ public static bool DoesNotHaveDeliveryChannel(this Asset asset, string deliveryC
...
     {
         if (channels == null) return false;
-        
+
         using var e = channels.GetEnumerator();
         return e.MoveNext() && e.Current == None && !e.MoveNext();
     }
 
+    /// <summary>
+    /// Checks if an asset is a stub asset: in space 0 with no delivery channels specified
```
donaldgray, line 89: A stub asset explicitly specifying `none` channel is also valid?

src/protagonist/API/Features/Queues/CustomerQueueController.cs
JackLewis-digirati: it was suggested by Claude, but I felt like it was a nicer way of handling default values than `?? 0` - I can swap it back to use that to keep continuity, but it felt like something we'd want to use going forward?

src/protagonist/DLCS.Model/Assets/AssetDeliveryChannels.cs
JackLewis-digirati: `IsNoneOnly` picks that up and it's used wherever `IsStubAsset` is used - I could add this in, but it's redundant in the current calls and `IsNoneOnly` picks up assets in other spaces that would be valid for those checks, as such it's probably best to rename this to be clearer, potentially something like `IsPotentialStubAsset`

donaldgray (approved):

PR #1201, opened 2026-05-14
Handle bucket copy for files greater than 50GB

donaldgray:
## What does this change?

Fixes #1200 

Fixes issue in s3 -> s3 copy operations and optimises it. Summary of changes

* Dynamic part sizing - replaces hardcoded 5MB part size. Size of origin file is read from metadata and part-size determined based on that, with a default of 16MB (matching `TransferUtility` from AWS which I'm assuming was optimised). Part size increases for files over 160GB to stay within S3's 10,000-part limit
* Parallel part copying - `Parallel.ForEachAsync` with bounded concurrency (default 4, configurable via appsetting) replaces sequential loop
* Cancellation handling - `OperationCanceledException` now caught explicitly, aborting the multipart upload before returning
* Removed `OverflowException` handling as this was due to too many parts, which is now prevented.

## Infrastructure Changes

> [!NOTE]
> This PR requires infrastructure changes.
>
> - **Optional:** For environments where larger AV files are expected, timebased queue `visibility_timeout` could be increased to allow for handling. In practice this won't have much affect as messages are only retried once.

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | Engine | `AWS:S3:CopyPartConcurrency` | N | Number of concurrent part uploads | 4 |

### DIFF
--- a/src/protagonist/DLCS.AWS/S3/S3BucketWriter.cs
+++ b/src/protagonist/DLCS.AWS/S3/S3BucketWriter.cs
@@ -4,20 +4,16 @@ using Amazon.S3;
 using Amazon.S3.Model;
 using Amazon.S3.Transfer;
 using DLCS.AWS.S3.Models;
+using DLCS.AWS.Settings;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace DLCS.AWS.S3;
 
-public class S3BucketWriter : IBucketWriter
+public class S3BucketWriter(IAmazonS3 s3Client, IOptions<AWSSettings> awsOptions, ILogger<S3BucketWriter> logger)
+    : IBucketWriter
 {
-    private readonly IAmazonS3 s3Client;
-    private readonly ILogger<S3BucketWriter> logger;
-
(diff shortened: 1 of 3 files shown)


src/protagonist/DLCS.AWS/S3/S3BucketWriter.cs
```diff
@@ -134,11 +117,15 @@ public async Task<LargeObjectCopyResult> CopyLargeObject(ObjectInBucket source,
             success = true;
             return new LargeObjectCopyResult(LargeObjectStatus.Success, objectSize);
         }
-        catch (OverflowException e)
+        catch (OperationCanceledException)
```
JackLewis-digirati, line 120: does this need a unit test?

JackLewis-digirati (approved):

src/protagonist/DLCS.AWS/S3/S3BucketWriter.cs
donaldgray: Added test - create 2 part copy and cancel CTS after first part uploaded.

PR #1203, opened 2026-05-20
Use null for the global space storage record

JackLewis-digirati:
## What does this change?

Links #1192

This PR updates `CustomerStorage` to work with adjuncts and is purely the protagonist side of this work.  

It goes further than the ticket implies and adds additional fields to `CustomerStorage` that show adjunct size and number.

It makes somechanges to logic, such as making sure that all customers have `CustomerStorage`
 records to move some creation logic into a more logical place

Finally, this PR fixes an issue with `ImageStorage`, where values weren't surface in the API, so now the response will contain `adjunctSize` values.

responses have been changed to look like the following:

`{{baseUrl}}/customers/:customerId/spaces/:spaceId/images/:imageId/storage`:
```
{
    "thumbnailSize": 0,
    "size": 0,
    "adjunctSize": 7487616,
    "lastChecked": "2026-05-22T10:27:08.4928440Z",
    "checkingInProgress": false
}
```

`{{baseUrl}}/customers/:customerId/spaces/:spaceId/storage`:

```
{
    "numberOfStoredImages": 0,
    "totalSizeOfStoredImages": 0,
    "totalSizeOfThumbnails": 0,
    "numberOfStoredAdjuncts": 1,
    "totalSizeOfStoredAdjuncts": 11231424,
    "storagePolicy": "large"
}
```

> [!Warning]
> In order to get the overall customer storage value the `{{baseUrl}}/customers/:customerId/storage` is now the *only* way to retrieve this value.  Previously, retrieving space `0` could be used as a stand-in that will no longer work as this is now the stub asset space
>  This is potentially a breaking change and should be flagged to customers

## Database Migration

> [!NOTE]
> - **Migration name:** `20260518133605_StopSpaceZeroCustomerStorage`
> - **What it does:** This PR contains a fairly complex migration that updates Customer storage `0` records to `null` and adds a new `0` record that will pick up the storage policy based on the "new" `null` value.  It also ensures every customer that doesn't have a `null` space now has one. Finally the table itself is modified to change the primary key from a composite of `space` and `customer` to a generated `int` as `space` is now nullable with indexes added to the table to continue to ensure uniqueness.  Finally, columns are added to customer storage for `NumberOfStoredAdjuncts` and `TotalSizeOfStoredAdjuncts`
> - **Breaking Change?** Yes, this will remove storage for adjuncts - created adjuncts will not be tracked for storage prior to this migration


src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
```diff
@@ -0,0 +1,143 @@
...
+            // StoragePolicy is inherited from the customer's aggregate (NULL-space) row.
+            migrationBuilder.Sql(@"
+INSERT INTO ""CustomerStorage"" (""Customer"", ""Space"", ""StoragePolicy"", ""NumberOfStoredImages"", ""TotalSizeOfStoredImages"", ""TotalSizeOfThumbnails"", ""NumberOfStoredAdjuncts"", ""TotalSizeOfStoredAdjuncts"")
+SELECT s.""Customer"", 0, agg.""StoragePolicy"", 0, 0, 0, 0, 0
+FROM ""Spaces"" s
+JOIN ""CustomerStorage"" agg ON agg.""Customer"" = s.""Customer"" AND agg.""Space"" IS NULL
+WHERE s.""Id"" = 0
+AND NOT EXISTS (
+    SELECT 1 FROM ""CustomerStorage"" cs WHERE cs.""Customer"" = s.""Customer"" AND cs.""Space"" = 0
+);");
```
JackLewis-digirati, line 79: There's currently no tests on this migration and it's quite complicated, but they would be fairly tightly coupled and expensive - worth doing anyway?

src/protagonist/DLCS.Repository/Storage/CustomerStorageRepository.cs
```diff
@@ -25,75 +24,42 @@ public CustomerStorageRepository(
...
-            }
-        }
+        var aggregateRow = await dlcsContext.CustomerStorages
+            .AsNoTracking()
+            .SingleOrDefaultAsync(cs => cs.Customer == customerId && cs.Space == null, cancellationToken);
 
-        // Which one of these should we return!!!
-        return space0Summary;
-        // return sumSummary;
+        return aggregateRow == null
```
donaldgray: The below method has `// The aggregate row is seeded by CreateCustomer and backfilled by migration - its absence is a bug.` but this method silently deals with a missing record. 

Should we align these? Or, if not, should we log a warning if there's no `null` space, or create one on the fly?

src/protagonist/DLCS.Repository/DlcsContext.cs
```diff
@@ -285,10 +285,24 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
         modelBuilder.Entity<CustomerStorage>(entity =>
         {
-            entity.HasKey(e => new { e.Customer, e.Space });
+            entity.HasKey(e => e.Id);
 
             entity.ToTable("CustomerStorage");
 
+            entity.Property(e => e.Id).UseIdentityByDefaultColumn();
+
+            // One row per real space per customer (Space IS NOT NULL)
```
donaldgray, line 294: Annoyingly PG15 supports this automatically with `NULLS NOT UNIQUE` which would avoid requiring multiple

src/protagonist/Engine.Tests/Integration/AdjunctIngestTests.cs
```diff
@@ -201,12 +204,66 @@ public async Task IngestAdjunct_Success_ReingestUpdatesAdjunctSize()
     }
 
     [Fact]
-    public async Task IngestAsset_Error_ExceedAllowance()
+    public async Task IngestAdjunct_Success_UpdatesCustomerStorage()
     {
-        // prep customer
-        await dbContext.Customers.AddTestCustomer(CustomerForLimits);
-        await dbContext.Spaces.AddTestSpace(CustomerForLimits, SpaceExceedLimit);
+        // Seed per-space storage row before parent-asset ingest so both rows are updated by that ingest
```
donaldgray, line 206: IMO the use of a "seedContext" makes this less-clear as you're now dealing with 2 different DbContexts, I think it's simpler just using 1 context but `dbContext.ChangeTracker.Clear();` after saving the "pre" stuff.

This also keeps these tests in-line with others.

src/protagonist/Engine/Data/EngineAssetRepository.cs
```diff
@@ -178,21 +178,28 @@ private async Task<bool> FinishBatchedItem<T>(IDeliverableBatchItem batchItem, I
         return batch?.Finished.HasValue ?? false ? batch : default;
     }
 
-    private async Task IncreaseCustomerStorage(ImageStorage imageStorage, CancellationToken cancellationToken)
+    private async Task IncreaseCustomerStorage(ImageStorage imageStorage, bool isAdjunct,
+        CancellationToken cancellationToken)
     {
         try
         {
+            // skip image/thumb size delta when the deliverable is an adjunct
```
donaldgray, line 186: Won't these already be 0 if it's an adjunct?

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
```diff
@@ -0,0 +1,143 @@
...
+                defaultValue: 0L);
+
+            // All existing Space=0 rows were aggregate rows.
+            // Real stub-asset Space=0 tracking has not started yet, so this is safe to do wholesale.
+            migrationBuilder.Sql(@"UPDATE ""CustomerStorage"" SET ""Space"" = NULL WHERE ""Space"" = 0;");
+
+            // Ensure every customer has an aggregate (NULL-space) row; some may never have had a Space=0 row.
+            migrationBuilder.Sql(@"
+INSERT INTO ""CustomerStorage"" (""Customer"", ""Space"", ""StoragePolicy"", ""NumberOfStoredImages"", ""TotalSizeOfStoredImages"", ""TotalSizeOfThumbnails"", ""NumberOfStoredAdjuncts"", ""TotalSizeOfStoredAdjuncts"")
+SELECT c.""Id"", NULL, 'default', 0, 0, 0, 0, 0
```
donaldgray, line 62: Will this move every customer onto the 'default' storage-policy, without respecting what their current policy is?

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
donaldgray: Tightly coupled to what?

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
```diff
@@ -0,0 +1,143 @@
...
+FROM ""Customers"" c
+WHERE NOT EXISTS (
+    SELECT 1 FROM ""CustomerStorage"" cs WHERE cs.""Customer"" = c.""Id"" AND cs.""Space"" IS NULL
+);");
+
+            // Ensure every customer that has a Space=0 (stub-assets) also has a per-space CustomerStorage row for it.
+            // StoragePolicy is inherited from the customer's aggregate (NULL-space) row.
+            migrationBuilder.Sql(@"
+INSERT INTO ""CustomerStorage"" (""Customer"", ""Space"", ""StoragePolicy"", ""NumberOfStoredImages"", ""TotalSizeOfStoredImages"", ""TotalSizeOfThumbnails"", ""NumberOfStoredAdjuncts"", ""TotalSizeOfStoredAdjuncts"")
```
donaldgray, line 72: We should set `""NumberOfStoredAdjuncts"", ""TotalSizeOfStoredAdjuncts""` correctly, rather than leaving as zero, or will something else handle that?

src/protagonist/DLCS.Model/Assets/AdjunctX.cs
```diff
@@ -3,10 +3,9 @@
 public static class AdjunctX
 {
     /// <summary>
-    /// Semantic helper for code readability. If `origin` has been provided in adjunct, it needs to be ingested into
-    /// hosted adjunct
+    /// Returns true if the adjunct is or will be hosted by DLCS (i.e. has an origin to ingest from).
+    /// 
+    /// NOTE: External adjuncts have no origin and are not ingested.
```
donaldgray: Should `NOTE:` be `<remarks>`?

src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
```diff
@@ -110,6 +110,30 @@ await dbContext.Spaces.AddAsync(new DLCS.Model.Spaces.Space
                 MaxUnauthorised = -1
             }, cancellationToken);
 
+            // Create null-space aggregate CustomerStorage row for storage limit checks
```
donaldgray: Small, this isn't just for storage limit checks. It is used for that in shared instance but on dedicated-instances it's not, it's used for tracking total storage only (no limit check or enforcement)

src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
```diff
@@ -110,6 +110,30 @@ await dbContext.Spaces.AddAsync(new DLCS.Model.Spaces.Space
...
+                StoragePolicy = StoragePolicy.DefaultStoragePolicyName,
+                NumberOfStoredImages = 0,
+                TotalSizeOfStoredImages = 0,
+                TotalSizeOfThumbnails = 0
+            }, cancellationToken);
+
+            // Pre-seed a per-space row for space 0 (stub-assets) so that Engine storage increments
+            // hit an existing row rather than silently no-oping on their first UPDATE.
+            // Other spaces get their per-space row created lazily by the Engine on first ingest.
+            await dbContext.CustomerStorages.AddAsync(new CustomerStorage
```
donaldgray, line 126:
```
// Pre-seed a per-space row for space 0 (stub-assets) so that Engine storage increments
// hit an existing row rather than silently no-oping on their first UPDATE.
// Other spaces get their per-space row created lazily by the Engine on first ingest.
```

Why would other per-space rows be created lazily by Engine but Space 0 isn't?

src/protagonist/API/Features/Adjuncts/AdjunctUpsertService.cs
```diff
@@ -87,14 +95,34 @@ public async Task<AdjunctDocument> HandleAdjunct(Adjunct adjunct, Adjunct? dbAdj
         return new AdjunctDocument(dbAdjunct, existingAdjunct);
     }
 
+    private async Task IncrementAdjunctCount(int customerId, int space, CancellationToken cancellationToken)
```
donaldgray: These "load -> update -> save" can lead to out of sync data. Is it worth using `UpdateFromQuery`?

src/protagonist/API/Features/Adjuncts/Requests/DeleteAdjunct.cs
```diff
@@ -36,12 +37,28 @@ public async Task<ResultMessage<DeleteResult>> Handle(DeleteAdjunct request, Can
         }
         
         dbContext.Adjuncts.Remove(adjunct);
+
+        if (adjunct.IsHosted())
+        {
+            var storageRows = await dbContext.CustomerStorages
```
donaldgray: `UpdateFromQuery()`?

Should these floor at 0, I think there's outstanding bug for assets with negative counts

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
JackLewis-digirati: apologies - it would be coupled onto the names of the migration - though granted, this is unlikely to change

src/protagonist/Engine/Data/EngineAssetRepository.cs
JackLewis-digirati: it's pulling from the `ImageStorage` record, which has details of the image itself and the adjuncts attached to it, so these values are filled out

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
JackLewis-digirati: the `"UPDATE ""CustomerStorage"" SET ""Space"" = NULL WHERE ""Space"" = 0;"` is what moves existing customers., while this is creating a default policy for any customers that don't have a storage policy setup (i.e.: never created an image in the previous code) - will add a comment to clarify

src/protagonist/API/Features/Customer/Requests/CreateCustomer.cs
JackLewis-digirati: space 0 is always the stub asset space, so we create it at customer creation as it can only be this - by contrast other spaces are customer created, and only matter for storage if the engine manages to ingest that first asset

JackLewis-digirati: After a call - issue is that `Other spaces get their per-space row created lazily by the Engine on first ingest` implies there's special handling in engine for space 0, need to clarify

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
JackLewis-digirati: after call - local testing on queries is fine

src/protagonist/API/Features/Adjuncts/Requests/DeleteAdjunct.cs
JackLewis-digirati: updated to use this, along with guards added to floor at zero

src/protagonist/DLCS.Repository/Storage/CustomerStorageRepository.cs
JackLewis-digirati: this now logs a warning

src/protagonist/DLCS.Repository/Migrations/20260518133605_StopSpaceZeroCustomerStorage.cs
JackLewis-digirati: The recalculators should be able to reconcile this with previous `ImageStorage` records - and this should only be an issue in the dev env, as others should have the migrations applied for `ImageStorage` at a similar/same time

src/protagonist/API/Features/Adjuncts/AdjunctUpsertService.cs
```diff
@@ -87,14 +96,33 @@ public async Task<AdjunctDocument> HandleAdjunct(Adjunct adjunct, Adjunct? dbAdj
...
 
+    private Task IncrementAdjunctCount(int customerId, int space, CancellationToken cancellationToken) =>
+        dbContext.CustomerStorages
+            .Where(cs => cs.Customer == customerId && (cs.Space == null || cs.Space == space))
+            .UpdateFromQueryAsync(cs => new CustomerStorage
+            {
+                NumberOfStoredAdjuncts = cs.NumberOfStoredAdjuncts + 1
+            }, cancellationToken);
+
+    private Task DecrementAdjunctStorage(int customerId, int space, long adjunctSize, CancellationToken cancellationToken) =>
```
donaldgray: This feels like something that should be reusable by `DeleteAdjunct`? The handling is identical, including floor handling etc.

donaldgray (approved):

PR #1204, opened 2026-05-26
Validate "annotations" adjuncts are JSON

donaldgray:
## What does this change?

Resolves #1195 

Validate that hosted `"iiifLink": "annotations"` are valid JSON. Abort if not the case. 

Summary
* Extend `AssetToS3.CopyOriginToStorage()` to accept an optional validation function. If passed this is called after the object is downloaded. 
  * _This will affect direct s3->s3 bucket copies as the object will be downloaded first_. Passing this validator function makes everything use the "indirect" path.
  * This function returns an error message. If `null` is returned then the validation passes, if not-null is returned then an error has occurred and an exception is thrown. Returning string for error/null for okay is a strategy used elsewhere and is documented in xml comments and is an easy to understand approach.
* ``FileChannelWorker` will treat `"annotations"` and non-`"annotations"` slightly differently to handle validation step.

> [!CAUTION]
> This changes `S3AmbientOriginStrategy` very slightly. It now detects a null return value and sets `.IsEmpty` in returned object if this is the case. 
> This is a widely used method but should not affect any handling as `Stream.IsNull()` detection will still work.

### DIFF
--- a/src/protagonist/Engine/Ingest/Persistence/AssetToS3.cs
+++ b/src/protagonist/Engine/Ingest/Persistence/AssetToS3.cs
@@ -17,7 +17,7 @@ public interface IAssetToS3
     /// Copy <see cref="IOriginItem"/> provided by the supplied <paramref name="context"/> from Origin to DLCS storage.
     /// Configuration determines if this is a direct S3-S3 copy, or S3-disk-S3.
     /// When <paramref name="validator"/> is provided the copy always goes via local disk so the validator can inspect
-    /// the file; the S3 write is skipped if the validator returns false.
+    /// the file; the S3 write is skipped if the validator returns an error.
     /// </summary>
     /// <param name="destination"><see cref="ObjectInBucket"/> where file is to copied to</param>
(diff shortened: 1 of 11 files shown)


src/protagonist/Engine/Ingest/Persistence/AssetToS3.cs
```diff
@@ -16,15 +16,21 @@ public interface IAssetToS3
     /// <summary>
     /// Copy <see cref="IOriginItem"/> provided by the supplied <paramref name="context"/> from Origin to DLCS storage.
     /// Configuration determines if this is a direct S3-S3 copy, or S3-disk-S3.
+    /// When <paramref name="validator"/> is provided the copy always goes via local disk so the validator can inspect
+    /// the file; the S3 write is skipped if the validator returns false.
```
JackLewis-digirati: tiny, but a don't think the validator can return false - maybe something like
```suggestion
    /// the file; the S3 write is skipped if the validator returns an error.
```

JackLewis-digirati (approved):
LGTM - only thing is the name of this PR could be improved as it will end up on release notes

PR #1205, opened 2026-05-27
Update wording of adjunct error message

JackLewis-digirati:
## What does this change?

This PR updates an error message to be closer to what is actually happening so that a user does not get confused with the issue being related to the database, as opposed to being a more general processing error


src/protagonist/API.Tests/Integration/AdjunctUnexpectedErrorTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        // Act
+        var response = await httpClient.AsCustomer(assetId.Customer).PutAsync(path, content);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.InternalServerError);
+
+        Func<Task> tryRead = () => response.ReadAsHydraResponseAsync<Error>();
+        (await tryRead.Should().ThrowAsync<DlcsException>())
+            .Which.Message.Should()
+            .Be($"Unknown error processing adjuncts for {assetId}");
```
donaldgray:
```suggestion
            .WithMessage($"Unknown error processing adjuncts for {assetId}");
```

donaldgray (approved):

PR #1206, opened 2026-05-28
Disallow spaces that aren't a positive integer

### LINKED ISSUE #997: Customer is able to create a space assigned to `0` or a negative number via PUT
Space numbers start at `1` - if a fresh new customer creates one via POST, it'll be set to `1` automatically.
But! There appears to be nothing stopping a customer from creating a new space that's assigned to `0` or a negative number via PUT.
A space set to `0` will cause Fluent `NotEmpty()` checks to fail, as mentioned in the [docs](https://docs.fluentvalidation.net/en/latest/built-in-validators.html#notempty-validator):
> NotEmpty() Ensures that the specified property is not null, an empty string or whitespace (or the default value for value types, e.g., 0 for int). When used on an IEnumerable (such as arrays, collections, lists, etc.), the validator ensures that the IEnumerable is not empty.

e.g, a space set to `0` exists and the user posts the following batch:
```
{
  "@type": "Collection",
  "member": [
    {
      "id": "b28047345-11744638577974",
      "mediaType": "image/jpg",
      "space": 0,
      "origin": "https://dlcsstage-public-test-objects.s3.eu-west-1.amazonaws.com/wellcome-images/b28047345/b28047345_0001.jp2.jpg",
      "string1": "b28047345",
      "number1": 1
    },
    {
      "id": "b28047345-21744638577974",
      "mediaType": "image/jpg",
      "space": 0,
      "origin": "https://dlcsstage-public-test-objects.s3.eu-west-1.amazonaws.com/wellcome-images/b28047345/b28047345_0002.jp2.jpg",
      "string1": "b28047345",
      "number1": 2
    }]
}
```
This will result in Protagonist returning a `Space cannot be empty` error.

JackLewis-digirati:
## What does this change?

Resolves #997

This PR fixes an issue where spaces could be created at a value of 0 or less.  This will disable the creation of negative spaces and update of the stub asset space

### DIFF
--- a/src/protagonist/API/Features/Space/Requests/PatchSpace.cs
+++ b/src/protagonist/API/Features/Space/Requests/PatchSpace.cs
@@ -1,6 +1,7 @@
 using API.Infrastructure.Requests;
 using DLCS.Core;
 using DLCS.Core.Strings;
+using DLCS.Model.Assets;
 using DLCS.Model.Spaces;
 using MediatR;
 
@@ -30,10 +31,10 @@ public class PatchSpaceHandler : IRequestHandler<PatchSpace, ModifyEntityResult<
     
     public async Task<ModifyEntityResult<DLCS.Model.Spaces.Space>> Handle(PatchSpace request, CancellationToken cancellationToken)
     {
-        if (request.SpaceId <= 0)
+        if (request.SpaceId == AssetDeliveryChannels.StubAssetSpace)
         {
             return ModifyEntityResult<DLCS.Model.Spaces.Space>.Failure(
-                "Space id must be a positive integer.",
+                "The stub asset space cannot be updated",
                 WriteResult.FailedValidation);
         }
 
--- a/src/protagonist/API/Features/Space/Requests/PatchSpace.cs
+++ b/src/protagonist/API/Features/Space/Requests/PatchSpace.cs
@@ -1,7 +1,6 @@
 using API.Infrastructure.Requests;
 using DLCS.Core;
 using DLCS.Core.Strings;
-using DLCS.Model.Assets;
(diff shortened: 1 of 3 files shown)


src/protagonist/API/Features/Space/Requests/PatchSpace.cs
```diff
@@ -30,6 +31,13 @@ public PatchSpaceHandler(ISpaceRepository spaceRepository)
     
     public async Task<ModifyEntityResult<DLCS.Model.Spaces.Space>> Handle(PatchSpace request, CancellationToken cancellationToken)
     {
+        if (request.SpaceId == AssetDeliveryChannels.StubAssetSpace)
```
donaldgray: Can this be same logic as `PutSpace` (ie `if (request.SpaceId <= 0)`). 

That way will save a DB lookup for a negative value and just reject it upfront.

src/protagonist/API/Features/Space/Requests/PutSpace.cs
```diff
@@ -31,6 +31,13 @@ public PutSpaceHandler(ISpaceRepository spaceRepository)
     
     public async Task<ModifyEntityResult<DLCS.Model.Spaces.Space>> Handle(PutSpace request, CancellationToken cancellationToken)
     {
+        if (request.SpaceId <= 0)
+        {
+            return ModifyEntityResult<DLCS.Model.Spaces.Space>.Failure(
+                "Space id must be a positive integer.",
```
donaldgray: Tiny, don't think any other failures messages are full sentences.
```suggestion
                "Space id must be a positive integer",
```

donaldgray (approved):

PR #1209, opened 2026-06-02
Allow modifiying the id of a hosted annotation id

JackLewis-digirati:
## What does this change?

Resolves #1194

This PR makes it so that hosted annotations (whether ambient or not) have their id values rewritten to match the id coming from protagonist.  An example of this would be the following json:

```json
{
  "@context": "http://iiif.io/api/annotation/3/context.json",
  "id": "random id",
  "type": "AnnotationPage",
  "items": []
}
```

being rewritten to:

```json
{
  "@context": "http://iiif.io/api/annotation/3/context.json",
  "id": "<orchestrator base url>/adjuncts/26/1/asset-id/adjunct-id",
  "type": "AnnotationPage",
  "items": []
}
```

based on the asset being `asset-id` and the adjunct being `adjunct-id`.  This will also work for rewritten paths. For example, changing the default of `adjuncts` to `supplementing` would see the id rewritten to `"<orchestrator base url>/supplementing/26/1/asset-id/adjunct-id"`

This works by streaming the object in and then rewriting the top level id to avoid additional processing time of fully deserializing the object

> [!Note]
> The code from this PR closely resembles the work written to do similar in iiif-presentation

> [!Note]
> If the `id` property is not present, it will be added

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | Orchestrator | `MaxAdjunctSizeBytes` | N | The maximum size on adjunct that can be retrieved through orchestrator | 5 * 1024 * 1024 (5MB) |

### DIFF
--- a/src/protagonist/Orchestrator.Tests/Integration/AdjunctHandlingTests.cs
+++ b/src/protagonist/Orchestrator.Tests/Integration/AdjunctHandlingTests.cs
@@ -10,7 +10,9 @@ using DLCS.Model.Assets;
 using DLCS.Model.Policies;
 using DLCS.Repository;
 using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
+using Orchestrator.Settings;
 using Orchestrator.Tests.Integration.Infrastructure;
 using Test.Helpers.Data;
 using Test.Helpers.Integration;
@@ -25,6 +27,7 @@ namespace Orchestrator.Tests.Integration;
 [Collection(OrchestratorCollection.CollectionName)]
 public class AdjunctHandlingTests : IClassFixture<ProtagonistAppFactory<Startup>>
 {
+    private readonly ProtagonistAppFactory<Startup> appFactory;
     private readonly DlcsContext dbContext;
     private readonly HttpClient httpClient;
     private readonly string stubAddress;
@@ -40,6 +43,7 @@ public class AdjunctHandlingTests : IClassFixture<ProtagonistAppFactory<Startup>
 
     public AdjunctHandlingTests(ProtagonistAppFactory<Startup> factory, OrchestratorFixture orchestratorFixture)
     {
+        appFactory = factory;
         var dbFixture1 = orchestratorFixture.DbFixture;
         dbContext = dbFixture1.DbContext;
         stubAddress = orchestratorFixture.ApiStub.Address;
@@ -197,14 +201,12 @@ public class AdjunctHandlingTests : IClassFixture<ProtagonistAppFactory<Startup>
         response.Headers.Should().ContainKey("x-asset-id").WhoseValue.Should().ContainSingle(id.ToString());
     }
     
-    [Theory]
-    [InlineData(IIIFLinkType.Annotations)]
-    [InlineData(IIIFLinkType.InlineAnnotation)]
-    public async Task Get_AnnotationAdjunct_RewritesTopLevelId(IIIFLinkType iiifLinkType)
+    [Fact]
+    public async Task Get_AnnotationAdjunct_RewritesTopLevelId()
     {
         // Arrange
         var id = AssetIdGenerator.GetAssetId();
-        var adjunctId = $"{nameof(Get_AnnotationAdjunct_RewritesTopLevelId)}_{iiifLinkType}";
+        var adjunctId = nameof(Get_AnnotationAdjunct_RewritesTopLevelId);
         var s3Key = $"{id}/adjuncts/{adjunctId}";
         var origin = $"http://{LocalStackFixture.OriginBucketName}.s3.amazonaws.com/{s3Key}";
 
@@ -227,7 +229,7 @@ public class AdjunctHandlingTests : IClassFixture<ProtagonistAppFactory<Startup>
         await dbContext.Images
             .AddTestAsset(id, mediaType: "text/plain", origin: $"{stubAddress}/testfile",
                 imageDeliveryChannels: deliveryChannelsForFile)
-            .WithTestAdjunct(adjunctId, mediaType: "application/json", iiifLinkType: iiifLinkType,
+            .WithTestAdjunct(adjunctId, mediaType: "application/json", iiifLinkType: IIIFLinkType.Annotations,
                 origin: origin);
         await dbContext.SaveChangesAsync();
 
@@ -293,9 +295,9 @@ public class AdjunctHandlingTests : IClassFixture<ProtagonistAppFactory<Startup>
     }
 
     [Theory]
-    [InlineData("my-proxy.com",      "const_value/{0}/{1}/{2}")]
+    [InlineData("my-proxy.com", "const_value/{0}/{1}/{2}")]
     [InlineData("non-versioned.com", "adjuncts/{0}/{1}/{2}")]
-    [InlineData("versioned.com",     "adj/_{1}/{2}")]
+    [InlineData("versioned.com", "adj/_{1}/{2}")]
     public async Task Get_AnnotationAdjunct_RewrittenId_RespectsPathTemplate(string host, string pathFormat)
     {
         // {0}=customer, {1}=asset, {2}=adjunctId. All three templates drop {space} and produce a path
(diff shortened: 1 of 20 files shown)


src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ProxyActionResult.cs
```diff
@@ -97,6 +98,16 @@ public class StatusCodeResult(HttpStatusCode statusCode, string? message = null)
     public static StatusCodeResult NotFound => new(HttpStatusCode.NotFound);
 }
 
+/// <summary>
+/// Result for annotation adjunct requests that require the top-level <c>id</c> to be rewritten on the fly.
```
donaldgray: Tiny - this isn't really tied to annotation adjuncts, it's just what we're using it for now.
```suggestion
/// Result for json objects that require the top-level <c>id</c> to be rewritten on the fly.
```

src/protagonist/Orchestrator/Infrastructure/ReverseProxy/ProxyActionResult.cs
```diff
@@ -97,6 +98,16 @@ public class StatusCodeResult(HttpStatusCode statusCode, string? message = null)
...
+/// </summary>
+public class IdRewriteProxyActionResult(ObjectInBucket source, string newId) : IProxyActionResult
+{
+    public ObjectInBucket Source { get; } = source;
```
donaldgray, line 109: For consistency, can these have XML comments like the other `IProxyActionResult` objects?

src/protagonist/Orchestrator/Infrastructure/IdRewriter/StreamingJsonProcessor.cs
```diff
@@ -0,0 +1,114 @@
+using System;
+using System.IO;
+using Microsoft.Extensions.Logging;
+using System.Text.Json;
+
+namespace Orchestrator.Infrastructure.IdRewriter;
+
+public static class StreamingJsonProcessor
```
donaldgray, line 14: Probably worth adding a `<remarks>` or something to indicate it's based on iiif-presentation implementation, so that it's a signpost should we be making any changes in here.

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRouteHandlers.cs
```diff
@@ -59,11 +66,59 @@ private static async Task ProcessResponse(ILogger logger, HttpContext httpContex
...
+    {
+        try
+        {
+            var objectFromBucket =
+                await bucketReader.GetObjectFromBucket(rewriteResult.Source, httpContext.RequestAborted);
+
+            if (objectFromBucket.Stream.IsNull())
+            {
+                logger.LogWarning("Annotation adjunct not found in S3 for {Path}", httpContext.Request.Path);
+                httpContext.Response.StatusCode = 500;
```
donaldgray, line 103: Should this be a 404, rather than a 500? Logged warning is "not found" so feels like a 404. How do assets handle this?

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRouteHandlers.cs
```diff
@@ -59,11 +66,59 @@ private static async Task ProcessResponse(ILogger logger, HttpContext httpContex
...
+                return;
+            }
+
+            await using var s3Stream = objectFromBucket.Stream;
+            using var bufferedInput = new MemoryStream();
```
donaldgray: We can pre-size `bufferedInput` and `outputStream` as we will get the `ContentLength` from the `GetObjectFromBucket` request. This will avoid reallocations to increase stream size.

```cs
using var bufferedInput = new MemoryStream((int?)objectFromBucket.Headers.ContentLength ?? 0);

using var outputStream = new MemoryStream((int)bufferedInput.Length);
```

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRouteHandlers.cs
```diff
@@ -59,11 +66,59 @@ private static async Task ProcessResponse(ILogger logger, HttpContext httpContex
...
+        {
+            await RewriteAndStreamAdjunct(logger, httpContext, bucketReader, rewriteResult);
+            return;
+        }
+
+        var proxyAction = proxyActionResult as ProxyActionResult;
         await ProxyRequest(logger, httpContext, forwarder, proxyAction);
     }
 
+    private static async Task RewriteAndStreamAdjunct(ILogger logger, HttpContext httpContext,
```
donaldgray, line 79: Is there logging anywhere that will allow us to see how often this is happening? If not we should add that to track, if this ends up being a very frequently hit path we might want to look to optimised (say, via https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream). 

Given we need to load the full adjunct into memory, and we know the ContentLength, we should add that to logs. That will allow us to know the sizes that we're actually streaming. If we log the size we should be able to correlate any memory issues with these requests.

Actually - as I'm writing this I'm recalling that the RFC suggested some max sizes to protect memory. I know it's not on the ticket but it does seem something that's worth adding, and would be relatively straight forward (and worth capturing on ticket):

```cs
if (objectFromBucket.Headers.ContentLength > MaxAdjunctSizeBytes)
{
  logger.LogWarning("too big");
  httpContext.Response.StatusCode = 500;
  return;
}
```

`objectFromBucket.Headers.ContentLength` comes from metadata so we haven't actually read anything into memory yet so allows a quick exit.

If `MaxAdjunctSizeBytes` comes from IOptions we can control without code change. I wonder if it should be part of the `IdRewriteProxyActionResult` - that way the `AdjunctRouteHandler` can set what the size is, could allow us to differentiate based on some rules that the `AdjunctRequestHandler` doesn't need to know about - it can just worry about validating and proxying.

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRequestHandler.cs
```diff
@@ -51,13 +54,27 @@ public async Task<IProxyActionResult> HandleRequest(HttpContext httpContext)
...
+            var newId = assetPathGenerator.GetFullPathForRequest(new BasicPathElements
+            {
+                RoutePrefix = AdjunctRouteHandlers.RoutePrefix,
+                CustomerPathValue = adjunctRequest.CustomerPathValue,
+                Space = adjunctRequest.Space,
+                AssetPath = $"{adjunctRequest.AssetId}/{adjunctRequest.AdjunctId}",
+            }, includeQueryParams: false);
+            var rewriteResult = new IdRewriteProxyActionResult(proxyTarget, newId);
+            rewriteResult.Headers.Add("Content-Type", orchestrationAdjunct.MediaType!.Value);
+            return rewriteResult;
```
donaldgray: Small - is it worth extracting to a helper method to increase readability?
```suggestion
            return GetIdRewriteResult(adjunctRequest, proxyTarget, orchestrationAdjunct);
```

src/protagonist/Orchestrator.Tests/Integration/AdjunctHandlingTests.cs
```diff
@@ -192,11 +197,238 @@ await dbContext.Images.AddTestAsset(id,
...
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        var body = await response.Content.ReadAsStringAsync();
+        var doc = JsonDocument.Parse(body);
+        doc.RootElement.GetProperty("id").GetString().Should().Be(expectedId);
+        doc.RootElement.GetProperty("type").GetString().Should().Be("AnnotationPage");
+    }
+
+    [Theory]
+    [InlineData("my-proxy.com",      "const_value/{0}/{1}/{2}")]
```
donaldgray: Not a fan of the "white space to line up" thing that Claude does as it's not idiomatic c#. The IDE will auto-format this away if you manually touch this method (e.g. add and remove the closing `}`) so I think it's worth removing to avoid noise in future PRs.

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRouteHandlers.cs
JackLewis-digirati: if you mean if the asset doesn't exist, then it would be a 404 - modified to match this now

src/protagonist/Orchestrator/Features/Adjuncts/AdjunctRouteHandlers.cs
JackLewis-digirati: Updated the AC's + added the guard - though I think it's best to pull from a setting than `IdRewriteProxyActionResult ` so it can be modified per environment (though with a default)

JackLewis-digirati: after a chat - this now sets it on `IdRewriteProxyActionResult` to better support future functionality

donaldgray (approved):

PR #1211, opened 2026-06-03
Decrement image storage for adjuncts

JackLewis-digirati:
## What does this change?

Resolves #1208 

This PR fixes an issue where image storage records were not correctly decremented when adjuncts were created

Additionally, there was an issue where customer storage did not decrement adjunct counts when an asset with adjuncts was deleted.  This is also fixed


src/protagonist/DLCS.Repository/Storage/ImageStorageX.cs
```diff
@@ -21,4 +23,16 @@ public static async Task UpsertImageStorageRecord(this DbSet<ImageStorage> image
             }
         }
     }
+
+    /// <summary>
+    /// Decrement adjunct size for specified asset.
+    /// </summary>
+    public static Task DecrementAdjunctSize(this DbSet<ImageStorage> imageStorages, AssetId assetId,
```
donaldgray, line 30: Wouldn't this be better included as part of `IStorageRepository.DecrementAdjunctStorage`? It's the same overall operation. Single method can decrement both the `ImageStorage` and `CustomerStorage`.

Both call sites of `IStorageRepository.DecrementAdjunctStorage` immediately calls this, having 1 single method do both would mean that you can't do one without the other. The consumer shouldn't need to care about exactly which DB tables are updated (e.g. if we added a specific `AdjunctStorage` table in future the consumer would call the same method and internally it can handle the decrement operation wherever is required.

src/protagonist/API/Features/Customer/Requests/DeleteMultipleAdjunctsById.cs
```diff
@@ -57,10 +63,23 @@ private async Task<int> DeleteAdjunctsFromDb(List<Adjunct> adjunctsFromDatabase,
         }
     }
 
+    private async Task DecrementStorageForHostedAdjuncts(List<Adjunct> adjuncts, CancellationToken cancellationToken)
+    {
+        var hostedAdjuncts = adjuncts.Where(a => a.IsHosted()).ToList();
+        if (hostedAdjuncts.Count == 0) return;
+
+        foreach (var adjunct in hostedAdjuncts)
```
donaldgray: Given this is a bulk operation is there any value to attempting to hit the DB as few times as possible? Say I wanted to delete 10 hosted adjuncts of Size=100, each from a single Asset. It'd be more efficient to decrement by `Size:1000, Count:10` once than `Size:100, Count:1` 10 times. Which would overall be 20 db hits rather than 2. 

How much this would actually save in DB work would be determined by use, although I suspect we'll receive more _"delete a lot of adjuncts from few assets"_ than _"delete a lot of adjuncts spread across many assets"_

src/protagonist/DLCS.Model/Storage/IStorageRepository.cs
```diff
@@ -1,11 +1,17 @@
...
     public Task<CustomerStorageSummary> GetCustomerStorageSummary(int customerId, CancellationToken cancellationToken);
     public Task<AssetStorageMetric> GetStorageMetrics(int customerId, CancellationToken cancellationToken);
-    public Task DecrementAdjunctStorage(int customerId, int space, long adjunctSize, CancellationToken cancellationToken);
-}
\ No newline at end of file
+
+    /// <summary>Decrement adjunct counts in CustomerStorage and AdjunctSize in ImageStorage.</summary>
+    public Task DecrementAdjunctStorage(AssetId assetId, long adjunctSize, CancellationToken cancellationToken);
+
+    /// <summary>Decrement adjunct counts in CustomerStorage and AdjunctSize in ImageStorage when there are multip[le adjuncts being removed.</summary>
```
donaldgray:
```suggestion
    /// <summary>Decrement adjunct counts in CustomerStorage and AdjunctSize in ImageStorage when there are multiple adjuncts being removed.</summary>
```

donaldgray (approved):

